PR #9, opened 2022-08-24
Refactor CMake files

Zopolis4:
Just a work in progress, putting it here so nobody doubles up by accident.


Zopolis4:
There is no way this should be building-- look at what I have done to `gui/CMakeLists.txt`. The CI is broken.

Exzap:
> There is no way this should be building-- look at what I have done to `gui/CMakeLists.txt`. The CI is broken.

Looked a bit into it and it's *complicated*, so no quick fix. Seems to only affect commits done after initial pull request. As long as you can confirm it compiles locally I'd merge the pull request. Just let me know when you are done

Zopolis4:
Unfortunately I cannot build locally due to CMake not finding the vcpkg installed dependencies. I'll try a CI PR in a bit.

Zopolis4:
Nvm-- the issue was because my build folder wasn't capitalized.

JakobDev:
#12 and #10 have now been merged. These 2 files need to be installed on Unix Systems in the following locations:
`info.cemu.Cemu.desktop` into `PREFIX/share/applications`
`info.cemu.Cemu.metainfo.xml` into `PREFIX/share/metainfo`
Could you add that into the install step as discussed #20?

Tachi107:
Re-posting here, as it got lost between all the other comments.

From https://github.com/cemu-project/Cemu/issues/1#issuecomment-1227704188

> @Zopolis4 I just noticed your CMake PR. I've been working on improving the current state of CMake too, as I'm looking into packaging Cemu for Debian and Ubuntu (as in creating a distro-provided package, not what's currently available here).
>
> Are you working on packaging too?

src/CMakeLists.txt
```diff
@@ -46,56 +39,50 @@ add_subdirectory(imgui)
...
-
-target_link_libraries(CemuBin PRIVATE CemuCommon CemuComponents CemuCafe CemuConfig CemuGui CemuAudio CemuInput CemuUtil)
-target_link_libraries(CemuBin PRIVATE CemuAsm)
-target_link_libraries(CemuBin PRIVATE OpenSSL::SSL)
-target_link_libraries(CemuBin PRIVATE ZLIB::ZLIB)
-target_link_libraries(CemuBin PRIVATE ${wxWidgets_LIBRARIES})
-target_link_libraries(CemuBin PRIVATE CURL::libcurl)
-target_link_libraries(CemuBin PRIVATE imgui::imgui)
-target_link_libraries(CemuBin PRIVATE pugixml pugixml::static pugixml::pugixml)
+set_target_properties(Cemu PROPERTIES OUTPUT_NAME "Cemu")
```
Tachi107, line 57: As the target is now named Cemu this should be no longer needed

src/CMakeLists.txt
```diff
@@ -46,56 +39,50 @@ add_subdirectory(imgui)
...
+  Util
+  Assembly
+  OpenSSL::SSL
+  ZLIB::ZLIB
+  ${wxWidgets_LIBRARIES}
+  CURL::libcurl
+  imgui::imgui
+  pugixml
+  pugixml::static
+  pugixml::pugixml
```
Tachi107, line 76: Why are you linking with pugixml three times?

src/Cafe/CMakeLists.txt
```diff
@@ -1,29 +1,489 @@
...
+  TitleList/SaveInfo.cpp
+  TitleList/SaveInfo.h
+  TitleList/SaveList.cpp
+  TitleList/SaveList.h
+  TitleList/TitleId.h
+  TitleList/TitleInfo.cpp
+  TitleList/TitleInfo.h
+  TitleList/TitleList.cpp
+  TitleList/TitleList.h
+  )
```
Tachi107, line 466: If they were using globs it probably means that they don't want to list files explicitly. I know it is considered a bad practise, but I would ask the dev's opinion first, as keeping track of this many files is error prone.

src/Cemu/DiscordPresence/DiscordPresence.cpp
```diff
@@ -1,7 +1,5 @@
 #include "DiscordPresence.h"
 
-#ifdef ENABLE_DISCORD_RPC
```
Tachi107, line 3: Why was this removed?

src/CMakeLists.txt
```diff
@@ -46,56 +39,50 @@ add_subdirectory(imgui)
...
+  Cafe
+  Config
+  Gui
+  Audio
+  Input
+  Util
+  Assembly
+  OpenSSL::SSL
+  ZLIB::ZLIB
+  ${wxWidgets_LIBRARIES}
```
Tachi107, line 71: Linking with the libraries is not enough. You should also use the `wxWidgets_INCLUDE_DIRS`, `wxWidgets_DEFINITIONS`, `wxWidgets_CXX_FLAGS`, `wxWidgets_LIBRARY_DIRS`, etc.

Tachi107 (changes requested):

src/CMakeLists.txt
Zopolis4: I don't know, that's not my code. I just refactored it.

src/Cemu/DiscordPresence/DiscordPresence.cpp
Zopolis4: Because the files are now added or not added in CMake

src/Cafe/CMakeLists.txt
Exzap: Nah there is no deeper meaning. The CMakeLists.txt files were cobbled together when port work started and I never got to clean or optimize them. We should list all the files explicitly.

Zopolis4:
The fmt stuff is a bit ugly, and will not work when not using vcpkg unless you decide to install version 7.0.2 manually, but that's about as easy as manually installing 7.0.x, which you needed to do anyways. It only needs to exist until we update to 9.x, anyways.

Tachi107:
> The fmt stuff is a bit ugly, and will not work when not using vcpkg unless you decide to install version 7.0.2 manually, but that's about as easy as manually installing 7.0.x, which you needed to do anyways. It only needs to exist until we update to 9.x, anyways.

Why 7.0.2? Every release on the 7.x branch should work just fine (it works with 7.1.3, for example)

Tachi107:
Anyway, the diff is _huge_, could you split the changes in different commits, where each one contains a single logical change, so that this is easier to review?

Zopolis4:
Most of the diff is just from listing the files, the amount of intellectual changes is pretty low.

src/Common/CMakeLists.txt
```diff
@@ -1,48 +1,50 @@
...
-	CemuComponents
-	Boost::nowide
-	Boost::filesystem
-	glm::glm
+target_link_libraries(Common PRIVATE
+  Boost::nowide
+  Boost::filesystem
+  Boost::headers
+  glm::glm
+  fmt::fmt
```
Tachi107, line 43: Why were `Boost::headers` and `fmt::fmt` moved from `PUBLIC` to `PRIVATE?

Tachi107: I don't understand what you mean. What were you trying to use `target_include_directories()` for?

Zopolis4: To add the headers that were previously added by making `Boost::headers` and `fmt::fmt:` `PUBLIC`.

Tachi107: You shouldn't use `target_include_directories()` for that. `Boost::headers` and `fmt::fmt` are "imported targets". They are not actual libraries, but are a like logical container of libraries, headers, flags, etc. They represent a dependency.

The way CMake decided that you should make use of these imported targets is by using the `target_link_libraries()` option. When you do, for example, `target_link_libraries(MyTarget PRIVATE Boost::headers)` what actually happens it that CMake will look at the contents of `Boost::headers` and add `-I/usr/include/boost-1.74` to the list of flags of MyTarget.

If you don't know what these functions really do, please take a look at CMake's documentation; it'll make your life easier:

- https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#pseudo-targets
- https://cmake.org/cmake/help/latest/command/target_link_libraries.html

Zopolis4: Ah, that makes sense now. Thanks

Zopolis4:
Closing in favour of a series of smaller PRs.

PR #12, opened 2022-08-24
Add .desktop file

JakobDev:
This file is needed to show Cemu in the Menu


TheLovinator1:
You forgot to change `Exec=jdMinecraftLauncher` :-)

JakobDev:
> You forgot to change `Exec=jdMinecraftLauncher` :-)

It is now fixed. Happens sometimes with copy/paste.

pascalpuffke:
Minor detail, wouldn't it be better if you squash those two commits into one?

info.cemu.cemu.desktop
```diff
@@ -0,0 +1,9 @@
+[Desktop Entry]
+Name=Cemu
+Type=Application
+Terminal=false
+Icon=info.cemu.cemu
```
Doomsdayrs, line 5: Capitalize the program name?

Tatsh:
Should also add file associations for WUD, WUX, WUA, RPX.

JakobDev:
> Capitalize the program name?

I ping @Exzap to ask if he prefers `info.cemu.cemu` or `info.cemu.Cemu`

> Should also add file associations for WUD, WUX, WUA, RPX.

This will first need a custom MimeType.

ReillyBrogan:
> > Capitalize the program name?
> 
> I ping @Exzap to ask if he prefers `info.cemu.cemu` or `info.cemu.Cemu`

There's actually a specification for this that should be followed. 

> \[[1](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s02.html)\] For applications, the part of the name of the desktop file before the .desktop extension should be a valid [D-Bus well-known name](https://dbus.freedesktop.org/doc/dbus-specification.html#message-protocol-names). This means that it is a sequence of non-empty elements separated by dots (U+002E FULL STOP), none of which starts with a digit, and each of which contains only characters from the set [A-Za-z0-9-_]: ASCII letters, digits, dash (U+002D HYPHEN-MINUS) and underscore (U+005F LOW LINE).
>
> The name of the desktop entry should follow the "reverse DNS" convention: it should start with a reversed DNS domain name controlled by the author of the application, in lower case. The domain name should be followed by the name of the application, which is conventionally written with words run together and initial capital letters (CamelCase). For example, if the owner of example.org writes "Foo Viewer", they might choose the name org.example.FooViewer, resulting in a file named org.example.FooViewer.desktop.

So the name of the desktop file should be `info.cemu.Cemu.desktop`, NOT `info.cemu.cemu.desktop` (which is non-conformant with the standard). The icon should be `info.cemu.Cemu` in order to match the name of the desktop file.

Tatsh:
> > Capitalize the program name?
> 
> I ping @Exzap to ask if he prefers `info.cemu.cemu` or `info.cemu.Cemu`
> 
> > Should also add file associations for WUD, WUX, WUA, RPX.
> 
> This will first need a custom MimeType.

Mimetype already exists coming from ROM Properties: `application/x-wii-u-rom`. This covers ISO, WUD and WUX. You can add the line `MimeType=application/x-wii-u-rom;`.

JakobDev:
> So the name of the desktop file should be info.cemu.Cemu.desktop, NOT info.cemu.cemu.desktop

Both are valid and in use, but since so many request Cemu instead of cemu, I changed it.

> Mimetype already exists coming from ROM Properties: application/x-wii-u-rom

Thanks. Didn't know that.

Exzap:
`info.cemu.Cemu.desktop` is fine with me.

Also, can we move this into a `dist/linux/` subdirectory to keep the root directory tidy?

JakobDev:
> Also, can we move this into a dist/linux/ subdirectory to keep the root directory tidy?

Done

Exzap:
Thanks, will merge this now. I'll also add an icon soon

PR #14, opened 2022-08-24
chore: init PKGBUILD

ElijahPepe:


infirit:
Why have it in the source repo, AUR is the place for community pkgbuilds.

ElijahPepe:
I suppose it's fine to put it in the source repo, though I agree that the AUR is a much better place. I'll open a GH issue for this to decide how we should go about putting Cemu (the native Linux port, not the Wine) on the AUR.

ElijahPepe:
I'm going to close this until the details surrounding #1 are cleared up.

PR #28, opened 2022-08-24
.github: Add issue template and funding information

FearlessTobi:
This adds a basic issue template and a Patreon link to the repository.
Suggestions for better wording/content are very welcome.

The files are loosely based on those of yuzu, but since I wrote a large part of the yuzu ones myself, there should be no licensing issues.


Exzap:
This is good, I removed FUNDING.yml because we don't promote the Patreon anymore, as its pretty uncertain what will happen to it.

PR #38, opened 2022-08-25
Rewrite Readme.md

Zopolis4:
Added building instructions into Readme.md, renamed it, removed repeated links, made more concise. I recognize that readme design is very much a matter of opinion, so this is just my personal idea.


Exzap:
There is some good stuff in there, but I'm not a fan of merging build instructions into the readme. Also I'm not sure why you removed the Windows build instructions and helper .bat. Anyway, I think our readme is mostly ok the way it is but I'll take some of the changes you made into account for a future readme.md/build.md overhaul. Thanks

Zopolis4:
Fair enough. I removed the .bat because I'm not a huge fan of the way it works, and think the CMake integration works well enough. I removed the build instructions from there because some of them do not apply and I felt some of them should have been assumed knowledge.

PR #41, opened 2022-08-25
Fix ImGui::Text() format strings

Rubo3:
This patch fixes the compiler warnings/errors with `format-security` enabled.


Exzap:
thanks!

PR #56, opened 2022-08-25
Add clang-format config file and lint workflow

### LINKED ISSUE #32: Consistent coding style
I would highly recommend that we add some sort of documentation on how to format code.

The best solution for this IMO would be a `.clang-format` file. That way the `clang-format` tool can format everything in a consistent style.

These are pretty easy to set up, and IDEs like CLion support them out of the box for automatic, on file save / hotkey formatting. A minimal `.clang-format` file only needs two lines if Exzap is okay with using a predefined style. (Just don't pick GNU.)

[Here's an example of good documentation](https://github.com/SerenityOS/serenity/blob/master/Documentation/CodingStyle.md) regarding this topic, and [here's a sample .clang-format](https://github.com/SerenityOS/serenity/blob/master/.clang-format).

Zopolis4:
Resolves #32.
Code style is very much a matter of personal opinion, and the file I linked is by no means final. 
To that end, I have not actually yet linted the tree, as it would be a pain to redo every time a file is modified without linting or the format file is changed.


src/.clang-format
```diff
@@ -0,0 +1,34 @@
...
+SpacesBeforeTrailingComments: 2
+SpacesInAngles: false
+SpacesInCStyleCastParentheses: false
+SpacesInConditionalStatement: false
+SpacesInContainerLiterals: false
+SpacesInParentheses: false
+SpacesInSquareBrackets: false
+Standard: Latest
+ReferenceAlignment: Pointer
+UseTab: Never
```
Tachi107, line 34: The current codebase uses tabs

Zopolis4: Ah. This was taken from my personal style, so i'm willing to change it if it is required. Does the codebase consistently use tabs?

Tachi107: Yes, just take a look at the source files. The CMake files are the only ones who don't do that, for some reason.

Tachi107: Anyway, the file should be adapted to Cemu's style, not yours ;)

FearlessTobi:
I made some changes to the clang format file you provided, to make it more similar to Cemus current coding style:

```
AlignAfterOpenBracket: Align
AlignConsecutiveAssignments: false
AlignConsecutiveDeclarations: false
AlignEscapedNewlinesLeft: false
AlignOperands: true
AlignTrailingComments: true
AllowShortBlocksOnASingleLine: Empty
AllowShortCaseLabelsOnASingleLine: false
AllowShortEnumsOnASingleLine: true
AllowShortFunctionsOnASingleLine: Empty
AllowShortIfStatementsOnASingleLine: false
AlwaysBreakTemplateDeclarations: true
BinPackArguments: true
BinPackParameters: true
BreakBeforeBraces: Allman
BreakBeforeTernaryOperators: true
ColumnLimit: 100
ConstructorInitializerAllOnOneLineOrOnePerLine: false
ConstructorInitializerIndentWidth: 4
ContinuationIndentWidth: 4
Cpp11BracedListStyle: true
IndentWidth: 4
KeepEmptyLinesAtTheStartOfBlocks: false
Language: Cpp
MaxEmptyLinesToKeep: 1
ObjCSpaceAfterProperty: false
PointerAlignment: Left
ReflowComments: true
SortIncludes: true
SortUsingDeclarations: true
SpaceAfterCStyleCast: false
SpaceAfterLogicalNot: false
SpaceAfterTemplateKeyword: false
SpaceBeforeAssignmentOperators: true
SpaceBeforeCaseColon: false
SpaceBeforeParens: ControlStatements
SpaceBeforeRangeBasedForLoopColon: true
SpaceBeforeSquareBrackets: false
SpaceInEmptyBlock: false
SpaceInEmptyParentheses: false
SpacesBeforeTrailingComments: 1
SpacesInAngles: false
SpacesInCStyleCastParentheses: false
SpacesInConditionalStatement: false
SpacesInContainerLiterals: true
SpacesInParentheses: false
SpacesInSquareBrackets: false
Standard: Latest
TabWidth: 4
UseTab: Always
```

Also clang-format 14, is not supported yet by many Linux distros, so maybe we should stick to version 12 for now?

Tatsh:
Unfortunately the `...` operator is not consistently used. Sometimes it's on the right and other times on the left.

```c++
	template <typename ...TArgs>
	[[nodiscard]] static fs::path GetMlcPath(std::wstring_view format, TArgs&&... args)
```

Zopolis4:
Re: clang-format 14, thats just the version in the CI. On the ground, it shouldn't make a ton of difference as we don't use any clang-format 14 specific settings, but it does have some bugfixes and advantages over 12. Also, when we have the option to use a more modern version, I think we should use the modern version.

Exzap:
Superseded by #181

Issue #48, opened 2022-08-25
[Bug] Random number generator causing static assert


ArtemisX64:
Cemu 
Version 2.0(Latest Upstream)
Test Platform: Linux

## Issue
Discover why random number generator is causing static assert particularly in cemu
## Other Notes
It works outside of cemu repository using clang14 and libstdc++
### Example of function having issue
```cc
std::string GenerateRandomString(size_t length, std::string_view characters)
{
    assert(!characters.empty());
    std::stringstream result;

    std::random_device rd;
    std::mt19937 gen(rd());
    std::uniform_int_distribution<decltype(characters.size())> index_dist(0, characters.size() - 1);
    for (uint32_t i = 0; i < length; ++i)
    {
        result << characters[index_dist(gen)]; //Issue occurs with this function[index_dist(gen)] during compilation
    }

    return result.str();
}
```

Exzap:
Reopening this because the actual cause has yet to be found

PR #64, opened 2022-08-26
Proper fix for static asserts

ArtemisX64:
Uses uniform_int_distribution function provided by boost::random instead of relying on standard library in systems other than Windows
It fixes https://github.com/cemu-project/Cemu/issues/48

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,7 +49,7 @@ jobs:
     - name: "Install system dependencies"
       run: |
         sudo apt update -qq
-        sudo apt install -y ninja-build cmake libgtk-3-dev libsecret-1-dev libgcrypt20-dev libsystemd-dev freeglut3-dev clang-12 nasm libpulse-dev
+        sudo apt install -y ninja-build cmake libgtk-3-dev libsecret-1-dev libgcrypt20-dev libsystemd-dev freeglut3-dev clang-12 nasm
 
     - name: "Bootstrap vcpkg"
       run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,7 +49,7 @@ jobs:
     - name: "Install system dependencies"
       run: |
         sudo apt update -qq
-        sudo apt install -y ninja-build cmake libgtk-3-dev libsecret-1-dev libgcrypt20-dev libsystemd-dev freeglut3-dev clang-12 nasm
+        sudo apt install -y ninja-build cmake libgtk-3-dev libsecret-1-dev libgcrypt20-dev libsystemd-dev freeglut3-dev clang-12 nasm libpulse-dev
 
     - name: "Bootstrap vcpkg"
       run: |
--- a/dependencies/ih264d/decoder/x86/ih264d_function_selector.c
+++ b/dependencies/ih264d/decoder/x86/ih264d_function_selector.c
@@ -77,7 +77,7 @@ void ih264d_init_function_ptr(dec_struct_t *ps_codec)
     }
 }
 
-#ifndef __WIN32
+#ifdef __clang__
 #include <cpuid.h>
 
 void __cpuid2(signed int* cpuInfo, unsigned int level)
--- a/dependencies/ih264d/decoder/x86/ih264d_function_selector.c
+++ b/dependencies/ih264d/decoder/x86/ih264d_function_selector.c
@@ -77,7 +77,7 @@ void ih264d_init_function_ptr(dec_struct_t *ps_codec)
     }
 }
 
-#ifdef __clang__
+#ifndef __WIN32
 #include <cpuid.h>
 
 void __cpuid2(signed int* cpuInfo, unsigned int level)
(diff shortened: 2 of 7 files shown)


Exzap:
I think it would be better if we put something like this into precompiled.h:
```
// workaround for a compile error with certain combos of Linux compilers and STL
#if BOOST_OS_LINUX
#include <boost/random/uniform_int.hpp>
namespace std_rd_workaround = boost::random;
#else
namespace std_rd_workaround = std;
#endif
```
And then in the .cpp files just replace `std::uniform_int_distribution` with `std_rd_workaround::uniform_int_distribution` instead of having all those #if-#else. It also keeps the workaround a bit more centralized

But since you added boost-random to vcpkg we may as well just use boost::random on all platforms and call it a day.

ArtemisX64:
> I think it would be better if we put something like this into precompiled.h:
> 
> ```
> // workaround for a compile error with certain combos of Linux compilers and STL
> #if BOOST_OS_LINUX
> #include <boost/random/uniform_int.hpp>
> namespace std_rd_workaround = boost::random;
> #else
> namespace std_rd_workaround = std;
> #endif
> ```
> 
> And then in the .cpp files just replace `std::uniform_int_distribution` with `std_rd_workaround::uniform_int_distribution` instead of having all those #if-#else. It also keeps the workaround a bit more centralized
> 
> But since you added boost-random to vcpkg we may as well just use boost::random on all platforms and call it a day.

Ok

Tachi107:
Why is the standard random number generator causing issues?

ArtemisX64:
> Why is the standard random number generator causing issues?

Currently, I can't pinpoint the real issue. (Also, because my laptop is slow, large compilation times). Most likely related with the build environment because outside of it, it compiles fine. (Verifed using clang++ and g++ using libstdc++12 and libc++14). But I think using the boost random generator is better than disabling static asserts in general

ArtemisX64:
@Exzap Can you review the change?

Tachi107:
> Currently, I can't pinpoint the real issue. (Also, because my laptop is slow, large compilation times). Most likely related with the build environment because outside of it, it compiles fine. (Verifed using clang++ and g++ using libstdc++12 and libc++14). But I think using the boost random generator is better than disabling static asserts in general

I instead think that the static assert could instead be telling us that there is a real issue in the code... Are you able to reproduce this with a minimal example? Possibly on compiler-explorer.com so that I can see the error too?

ArtemisX64:
@Tachi107 https://compiler-explorer.com/z/rbPP7Mdz1

Tachi107:
This is odd. Are you able to show the output of the static_assert?

Crementif:
I'd prefer to extract the ubuntu audio change out of this PR. Could you make it a separate PR cuz it's a great change otherwise?

ArtemisX64:
> I'd prefer to extract the pulseaudio-dev CI change out of this PR. Could you make it a separate PR cuz it's a great change otherwise?

OOps my bad. Made to run in my localized repo. I will make the change

ArtemisX64:
@Tachi107 
```
static_assert(!_Up_traits::__is_signed, "U must be unsigned");
static_assert(!_Wp_traits::__is_signed, "W must be unsigned");
```
This was the static_asserts we were getting

Tachi107:
@ArtemisX64 

> ```
> static_assert(!_Up_traits::__is_signed, "U must be unsigned");
> static_assert(!_Wp_traits::__is_signed, "W must be unsigned");
> ```
> 
> This was the static_asserts we were getting with cemu(When calling generators)

Stupid question: does explicitly using `U` fixes the issue?

```cpp
std::uniform_int_distribution<decltype(characters.size())> index_dist(0U, characters.size() - 1U);
```

ArtemisX64:
@Tachi107 Nope. Still the same static_assert. (Tried now). Yesterday, I explicitly created two variables to check it and I had the same issue.(Only with Cemu)

ArtemisX64:
> This is really odd, and really interesting too. This only happens on Clang 14 and above, right? 
My findings is it happens on latest libstdc++ version. (Which makes sense since it compiles on Ubuntu 20.04 and not on other systems with latest libstdc++12 and clang12 and iirc, new version added static_assert patch which is harmless in general)
>And only with libstdc++ or also with libc++?
It failes to compile with libc++ because of WxWidgets iirc. (Tried once)

ArtemisX64:
I am planning on making it compile on g++. But at first, I have to deal with dllexports. (Or -fms-extensions flag or windows only functions in general). Do you know any alternatives for -fms-extensions flag for gcc?

ArtemisX64:
@Exzap I think it is ready to be merged.

Tachi107:
Could you point me to the "dllexport" code? I think I know how to deal with it, but maybe I'm thinking at something else

ArtemisX64:
> Could you point me to the "dllexport" code? I think I know how to deal with it, but maybe I'm thinking at something else

The fix is simple. What I need is not to do it for now. (Since I don't know how to replace _umul64() )

ArtemisX64:
@Tachi107 You can find them by disabling -fms-extensions flag

Exzap:
One last nitpick: Please start single-line comments with a lowercase letter and a leading space to match the rest of the codebase. 
E.g. `// use boost ...` instead of `//Use boost ...`
Everything else is ok and I'll merge it after you change the comments. Thanks for this improved workaround!

PR #66, opened 2022-08-26
perf: optimize GenerateRandomString()

Tachi107:
The previous implementation used an `std::stringstream` to concatenate the generated random string.

The new one uses a simple preallocated `std::string`, as the size of the output is already known - it is the `length` parameter.

It also uses `std::generate_n()` instead of an explicit loop, making the code more concise and potentially faster, as no calls to `std::string::operator+=` are needed.

Calling `GenerateRandomString(1'000'000)` with the `std::stringstream`-based implementation allocated 16 times, for a total of 3'173'516 bytes. The new one cuts this down to 4 allocs, for a total of 1'076'864 bytes.

Inspired by #64

### DIFF
--- a/src/util/helpers/helpers.cpp
+++ b/src/util/helpers/helpers.cpp
@@ -417,18 +417,20 @@ std::string GenerateRandomString(size_t length)
 	return GenerateRandomString(length, kCharacters);
 }
 
-std::string GenerateRandomString(size_t length, std::string_view characters)
+std::string GenerateRandomString(const size_t length, const std::string_view characters)
 {
 	assert(!characters.empty());
-	std::stringstream result;
+	std::string result;
+	result.resize(length);
 
 	std::random_device rd;
 	std::mt19937 gen(rd());
 	std::uniform_int_distribution<decltype(characters.size())> index_dist(0, characters.size() - 1);
-	for (uint32_t i = 0; i < length; ++i)
-	{
-		result << characters[index_dist(gen)];
-	}
+	std::generate_n(
+		result.begin(),
+		length,
+		[&] { return characters[index_dist(gen)]; }
+	);
 
-	return result.str();
+	return result;
 }


src/util/helpers/helpers.cpp
```diff
@@ -417,18 +417,19 @@ std::string GenerateRandomString(size_t length)
 	return GenerateRandomString(length, kCharacters);
 }
 
-std::string GenerateRandomString(size_t length, std::string_view characters)
+std::string GenerateRandomString(const size_t length, const std::string_view characters)
 {
 	assert(!characters.empty());
-	std::stringstream result;
+	std::string result;
+	result.reserve(length);
```
Exzap, line 424: This can be made even more efficient with two simple changes. The first is to use .resize here

src/util/helpers/helpers.cpp
```diff
@@ -417,18 +417,19 @@ std::string GenerateRandomString(size_t length)
...
 
 	std::random_device rd;
 	std::mt19937 gen(rd());
 	std::uniform_int_distribution<decltype(characters.size())> index_dist(0, characters.size() - 1);
 	for (uint32_t i = 0; i < length; ++i)
 	{
-		result << characters[index_dist(gen)];
+		result += characters[index_dist(gen)];
```
Exzap, line 431: And replace this with `result[i] = characters[index_dist(gen)];`
That lets you avoid the overhead that comes with appending (size tracking and checks)

Tachi107: I've changed this to std::generate_n(), so that no loop is needed at all.

src/util/helpers/helpers.cpp
Tachi107: Done

Exzap:
thanks!

src/util/helpers/helpers.cpp
Tachi107: Just curious, how does `.resize()` improve things compared to `.reserve()`?

PR #68, opened 2022-08-26
Add Rudimentray Proxy Support

tastymeatball:
This will add rudimentary proxy support for napi.
This can be used to manipulate/sniff the packets. (Needs CACERT_NINTENDO_CA_G3.der to be swapped out)
Or simply connect to the official servers when a proxy server is required.

Right now this will add a new config to the settings.xml called <proxy_server>.
This has to be edited to something like <proxy_server>http://127.0.0.1:8080</proxy_server> manually to use a proxy because no GUI exists right now.

### DIFF
--- a/src/Cemu/napi/napi_helper.cpp
+++ b/src/Cemu/napi/napi_helper.cpp
@@ -73,6 +73,10 @@ CurlRequestHelper::CurlRequestHelper()
 
 	curl_easy_setopt(m_curl, CURLOPT_FOLLOWLOCATION, 1);
 	curl_easy_setopt(m_curl, CURLOPT_MAXREDIRS, 2);
+
+	if(GetConfig().proxy_server.GetValue() != "") {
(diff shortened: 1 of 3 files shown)


src/Cemu/napi/napi_helper.cpp
```diff
@@ -73,6 +73,10 @@ CurlRequestHelper::CurlRequestHelper()
 
 	curl_easy_setopt(m_curl, CURLOPT_FOLLOWLOCATION, 1);
 	curl_easy_setopt(m_curl, CURLOPT_MAXREDIRS, 2);
+
+	if(GetConfig().proxy_server.GetValue() != "") {
```
Exzap, line 77: Can you move the brace to the next line to match our formatting style

Exzap:
The proxy should also be set in `CurlSOAPHelper::CurlSOAPHelper()` for the SOAP requests

Exzap:
Looking good. Thanks

Issue #65, opened 2022-08-26
[Bug] Game updates installed on wrong path on Unix


varlesh:
When you install updates, Cemu used Windows paths... but it's wrong on Linux (and MacOS too).
For example,
wrong `mlc01/usr\title\0005000e\10101d00`
right  `mlc01/usr/title/0005000e/10101d00`

![image](https://user-images.githubusercontent.com/8083855/186870151-d97c348a-124a-47fc-a19c-dae9ad248791.png)

PR #70, opened 2022-08-26
Fixes Path Issues

ArtemisX64:
Fixes path issues in *nix
Need to confirm on Windows
Fixes https://github.com/cemu-project/Cemu/issues/65


Wunkolo:
Looks like a use-case for [std::filesystem::path](https://en.cppreference.com/w/cpp/filesystem/path).

ArtemisX64:
@Wunkolo Do I really need to use std::filesystem::path ? (Since windows already supports / for paths)

Wunkolo:
It would allow for stuff like `blah / "usr" / "title" / fmt::format(...)` and the platform would handle the `/` and `\` automatically and might handle stuff like unicode characters or long paths better.

If windows already handles `/` safely and predictably already then yeah there is no need to.

ArtemisX64:
@Wunkolo Makes sense using std::filesystem::path. I will correct my PR accordingly

ArtemisX64:
@Wunkolo and @Exzap Can you review the code?

ArtemisX64:
@Exzap ready for review

Exzap:
Using std::filesystem::path seems unnecessarily verbose when forward slashes work just fine on any platform we might target. But more importantly this function should also work for Wii U internal paths where forward slashes are required.

ArtemisX64:
> Using std::filesystem::path seems unnecessarily verbose when forward slashes work just fine on any platform we might target. But more importantly this function should also work for Wii U internal paths where forward slashes are required.

fixed

PR #76, opened 2022-08-26
fix some printf format specifiers

notspiff:
use the definition in cinttypes for portability


Exzap:
We may want to use either std::format or fmtlib everywhere in the future but for now this is fine. Thanks

PR #77, opened 2022-08-26
gui: keep shown graphic pack when clicking on preset dropdown

gurrgur:
On linux, the right side panel of the graphic pack window (2) is being cleared when clicking on any of the preset dropdown menus.
This PR prevents this from happening.


Exzap:
Thanks!
Not that I'd expect any side effects on Windows but I did a quick check and it's good.

PR #75, opened 2022-08-26
build: improve the Linux aspect of things

Tachi107:
This is still a work in progress, and is likely going to conflict with #9 (Cc: @Zopolis4)

With this PR I aim at making Cemu's build script more compatible with how Linux operates, so this mainly means that system dependencies will be used when possible.

I'll split the changes in different commits on purpose, so please don't squash this on merge.

Could potentially fix #1.

PR #80, opened 2022-08-26
GCC

tomlally:
Now compiles on GCC and Clang.


ArtemisX64:
Can confirm using gcc also fixes the final linking error

ArtemisX64:
Weird but it no longer compiles on clang

ArtemisX64:
It will work fine in clang14(libstdc++12)

src/Cafe/Account/Account.cpp
```diff
@@ -416,7 +416,7 @@ fs::path Account::GetFileName(uint32 persistent_id)
 	if (persistent_id < kMinPersistendId)
 		throw std::invalid_argument(fmt::format("persistent id {:#x} is invalid", persistent_id));
 	
-	return CemuApp::GetMLCPath(fmt::format(L"usr\\save\\system\\act\\{:08x}\\account.dat", persistent_id)).ToStdWstring();
+	return CemuApp::GetMLCPath(fmt::format(L"usr/save/system/act/{:08x}/account.dat", persistent_id)).ToStdWstring();
```
Tachi107, line 419: Maybe this have to be `ifdef`d, it might create issues on Windows, no?

src/Cafe/CMakeLists.txt
```diff
@@ -1,10 +1,16 @@
 project(CemuCafe)
 
 include_directories(".")
+
+if((CMAKE_C_COMPILER_ID MATCHES "GNU") OR (CMAKE_C_COMPILER_ID MATCHES "Clang"))
```
Tachi107, line 5: Use `STREQUAL` instead of `MATCHES`. The latter is for regular expressions.

```suggestion
if((CMAKE_C_COMPILER_ID STREQUAL "GNU") OR (CMAKE_C_COMPILER_ID STREQUAL "Clang"))
```

src/Cafe/HW/Espresso/PPCTimer.cpp
```diff
@@ -7,9 +7,11 @@
 
 #if BOOST_OS_LINUX || BOOST_OS_MACOS
 static __inline__
-unsigned __int64 _umul128(unsigned __int64,
-                          unsigned __int64,
-                          unsigned __int64*);
+uint64 _umul128(uint64 multiplier, uint64 multiplicand, uint64 *highProduct) {
```
Tachi107, line 10: Is `uint64` standard? Why not `uint64_t`?

src/Common/precompiled.h
```diff
@@ -176,6 +176,16 @@ inline sint16 _swapEndianS16(sint16 v)
...
 
+inline uint64 _rotl64 (uint64 x, sint8 r)
+{
+    return (x << r) | (x >> (64 - r));
+}
+
+inline uint64 _rotr64 (uint64 x, sint8 r)
+{
+    return (x >> r) | (x << (64 - r));
+}
```
Tachi107, line 187: C++20 has [`std::rotl`](https://en.cppreference.com/w/cpp/numeric/rotl) and [`std::rotr`](https://en.cppreference.com/w/cpp/numeric/rotr)

src/Common/precompiled.h
```diff
@@ -210,11 +220,26 @@ typedef union _LARGE_INTEGER {
...
 // macros
 #if BOOST_OS_WINDOWS
-#define DLLEXPORT __declspec(dllexport)
+    #define DLLEXPORT __declspec(dllexport)
+    #define DLLIMPORT  __declspec(dllimport)
+    #define DEBUG_BREAK __debugbreak()
+    #define ALIGN(N) __declspec(align(N))
+    #define NOINLINE __declspec(noinline)
+    #define ASSUME(X) __assume((X)
+    #define THREAD_LOCAL __declspec(thread)
```
Tachi107, line 229: [`thread_local`](https://en.cppreference.com/w/cpp/keyword/thread_local) is standard since C++11, you can drop this macro

src/Common/precompiled.h
```diff
@@ -210,11 +220,26 @@ typedef union _LARGE_INTEGER {
...
+    #define POPCNT(X) __popcnt((X))
 #else
-#define DLLEXPORT __declspec(dllexport)
+    #define DLLEXPORT
+    #define DLLIMPORT
+    #define DEBUG_BREAK
+    #define ALIGN(N) __attribute__((aligned (N)))
+    #define NOINLINE __attribute__((noinline))
+    #define ASSUME(X)
```
Tachi107, line 237: Stackoverflow suggests a [solution](https://stackoverflow.com/questions/25667901/assume-clause-in-gcc/26195434#26195434), I don't know how affective it is.

src/Common/precompiled.h
```diff
@@ -210,11 +220,26 @@ typedef union _LARGE_INTEGER {
...
+    #define ALIGN(N) __declspec(align(N))
+    #define NOINLINE __declspec(noinline)
+    #define ASSUME(X) __assume((X)
+    #define THREAD_LOCAL __declspec(thread)
+    #define POPCNT(X) __popcnt((X))
 #else
-#define DLLEXPORT __declspec(dllexport)
+    #define DLLEXPORT
+    #define DLLIMPORT
+    #define DEBUG_BREAK
```
Tachi107, line 234: I've never used break points in code (my code editor allows me to insert them at a specific line without altering the source), so the best I can do is link this [SO page](https://stackoverflow.com/questions/173618/is-there-a-portable-equivalent-to-debugbreak-debugbreak)

src/Common/precompiled.h
```diff
@@ -393,7 +418,7 @@ class RunAtCemuBoot // -> replaces this with direct function calls. Linkers othe
 template<typename T>
 bool future_is_ready(std::future<T>& f)
 {
-#ifdef __clang__
+#if defined(__clang__) || defined(__GNUC__)
```
Tachi107, line 421: Clang defines `__GNUC__` already

src/config/XMLConfig.h
```diff
@@ -300,7 +291,6 @@ class XMLConfigParser
 		return set_attribute(name, value.GetValue());
 	}
 	
-	template <>
```
Tachi107, line 303: ?

Tachi107 (changes requested):
Thanks for working on this! I've left a few comments below.

src/Common/precompiled.h
ArtemisX64: It causes UB when used with 64 bit values. (Need to investigate further)

src/Common/precompiled.h
ArtemisX64: I added it in a separate PR

src/Cafe/HW/Espresso/PPCTimer.cpp
ArtemisX64: I think it's for consistency. (Since uint64 is used throughout the code)

src/Cafe/Account/Account.cpp
ArtemisX64: It won't. See https://github.com/cemu-project/Cemu/pull/70

src/Common/precompiled.h
ArtemisX64: https://stackoverflow.com/questions/13106049/what-is-the-performance-penalty-of-c11-thread-local-variables-in-gcc-4-8

src/Common/precompiled.h
ArtemisX64: I have added it in a separate PR

src/config/XMLConfig.h
ArtemisX64: Fixes the need for -fdelayed-template-parsing flag in clang and make it work on gcc

src/Common/precompiled.h
Tachi107: Oh, wow, didn't know that. But it also says:

> This wrapper is not needed for in every use case of thread_local though. This can be revealed from decl2.c. The wrapper is generated only when:
> 
>     It is not function-local, and,
>         It is extern (the example shown above), or
>         The type has a non-trivial destructor (which is not allowed for __thread variables), or
>         The type variable is initialized by a non-constant-expression (which is also not allowed for __thread variables).
> 
> In all other use cases, it behaves the same as __thread. That means, unless you have some extern __thread variables, you could replace all __thread by thread_local without any loss of performance.

src/Cafe/HW/Espresso/PPCTimer.cpp
Tachi107: I would personally replace all instances of `uint64` with `uint64_t`, but it's not a big deal obviously :)

src/Common/precompiled.h
Tachi107: What UB does it cause? The C++ standard says that the implementation should behave basically as if it was implemented like these two functions, see https://timsong-cpp.github.io/cppwp/n4868/bit.rotate

> Let `r` be `s % N`.
>
> Returns: If `r` is `0`, `x`; if `r` is positive, `(x << r) | (x >> (N - r))`; if `r` is negative, `rotr(x, -r)`.

ArtemisX64: I don't know why but yesterday, when I used it, I got an UB. Might be related to the code I was testing

Tachi107:
What kind of UB did you encounter?

-- 

ArtemisX64:
> What kind of UB did you encounter?
> […](#)
> -- OpenPGP key: 66DE F152 8299 0C21 99EF A801 A8A1 28A8 AB1C EE49

Can't recollect. But I think it will be fine. (As long as it receives uint64_t as a param)

Tachi107:
> > What kind of UB did you encounter?
>
> Can't recollect. But I think it will be fine. (As long as it receives uint64_t as a param)

Maybe the issue was that the parameter type was uint64 instead of uint64_t, and `numeric_­limits<uint64>​::​digits` doesn't work on some platforms?

RyzenDew:
after hours of testing and compiling on Fedora 36/37 and Arch and testing games ive not had any issues or crashing that isn't normal for the main linux build great work LGTM

RyzenDew (approved):

src/Cafe/CMakeLists.txt
tomlally: Yep, thanks. Fixed.

src/Common/precompiled.h
tomlally: I only wrote these wrappers because _rotl64 and _rotr64 were used in the code and I wanted to get it compiling asap. Since then I've added <bit> to precompiled.h so maybe it's a good idea to just go through and replace all occurrences with the std functions. Thoughts?

src/Common/precompiled.h
tomlally: See https://github.com/tomlally/Cemu/commit/73a8e6faaca15a80ae3e90cf6cbf603d9e6f03b1

src/Cafe/HW/Espresso/PPCTimer.cpp
tomlally: > I think it's for consistency. (Since uint64 is used throughout the code)

Yep, that was my reasoning.

src/Common/precompiled.h
tomlally: I was more concerned about breaking windows versions because I have no easy way of testing. However, MSDN advises using thread_local so I expect it would be fine.

src/Common/precompiled.h
tomlally: I've realised `__debugbreak` was only used in `precompiled.h` so now I'm thinking it may be tidier to conditionally replace the cemu_assert_... functions.

src/Common/precompiled.h
```diff
@@ -203,18 +220,52 @@ typedef union _LARGE_INTEGER {
...
+    #error No definition for DLLEXPORT and DLLIMPORT
+#endif
+
+#if defined(_MSC_VER)
+    #define NOINLINE __declspec(noinline)
+#elif defined(__GNUC__)
+    #define NOINLINE __attribute__((noinline))
+#else
+    #error No definition for NOINLINE
+#endif
```
Crementif, line 262: NOINLINE can be removed, see other comment.

src/main.cpp
```diff
@@ -27,6 +27,8 @@
 #include "Cafe/IOSU/legacy/iosu_crypto.h"
 #include "Cafe/OS/libs/vpad/vpad.h"
 
+#include "Common/platform.h"
+
```
Crementif, line 31: Is this include necessary for compilation on GCC?

Crementif (changes requested):
Seems like the PR kinda got stacked with changes meant for Linux. Nice work!

Also, you've got empty line changes in many (unrelated) files that can probably be cleaned up as well. And if you could follow @Tachi107 suggestions.

src/Cafe/HW/Espresso/Recompiler/PPCRecompilerX64.cpp
```diff
@@ -2247,7 +2250,10 @@ void PPCRecompilerX64Gen_imlInstruction_name_r(PPCRecFunction_t* PPCRecFunction,
 		else if (sprIndex == SPR_XER)
 			x64Emit_mov_mem32_reg64(x64GenContext, REG_RSP, offsetof(PPCInterpreter_t, spr.XER), tempToRealRegister(imlInstruction->op_r_name.registerIndex));
 		else if (sprIndex >= SPR_UGQR0 && sprIndex <= SPR_UGQR7)
-			x64Emit_mov_mem32_reg64(x64GenContext, REG_RSP, offsetof(PPCInterpreter_t, spr.UGQR[sprIndex-SPR_UGQR0]), tempToRealRegister(imlInstruction->op_r_name.registerIndex));
+		{
+			sint32 memOffset = offsetof(PPCInterpreter_t, spr.UGQR) + sizeof(PPCInterpreter_t::spr.UGQR[0]) * (sprIndex - SPR_UGQR0);
+			x64Emit_mov_mem32_reg64(x64GenContext, REG_RSP, memOffset, tempToRealRegister(imlInstruction->op_r_name.registerIndex));
```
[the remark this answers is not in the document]
Exzap: I believe doing offsetof where the member is a dynamic expression like in `spr.UGQR[sprIndex-SPR_UGQR0]` is not very portable. So this change is fine.

src/main.cpp
ArtemisX64: > Is this include necessary for compilation on GCC?

I believe main.cpp uses __declspec or other windows specific functions which are fixed using macros in precompiled.h

src/Common/precompiled.h
ArtemisX64: > I only wrote these wrappers because _rotl64 and _rotr64 were used in the code and I wanted to get it compiling asap. Since then I've added `<bit>` to `precompiled.h` so maybe it's a good idea to just go through and replace all occurrences with the std functions. Thoughts?

Since we are on C++20, it is safe to use rotr or rotl rather than a custom implementation. And also, I found rotr and rotl to be more optimized in godbolt

src/Cafe/HW/Latte/Renderer/Vulkan/RendererShaderVk.cpp
```diff
@@ -131,7 +131,7 @@ const TBuiltInResource DefaultTBuiltInResource = {
 	/* .maxDualSourceDrawBuffersEXT = */ 1,
 #endif
 
-	/* .limits = */ {
+	/* .limits = */
```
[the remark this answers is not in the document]
Tatsh: I ran into this same issue with GCC 11.3.0.

src/main.cpp
tomlally: I think it requires unix/platform which is included in Common/platform. Will double check.

src/Cafe/HW/Espresso/Interpreter/PPCInterpreterMain.cpp
```diff
@@ -34,7 +34,7 @@ PPCInterpreter_t* PPCInterpreter_getCurrentInstance()
 	return ppcInterpreterCurrentInstance;
 }
 
-__declspec(noinline) uint64 PPCInterpreter_getMainCoreCycleCounter()
+NOINLINE uint64 PPCInterpreter_getMainCoreCycleCounter()
```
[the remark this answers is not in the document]
tomlally: Noted. Will remove NOINLINE.

src/Common/precompiled.h
tomlally: Done.

src/Cafe/OS/libs/coreinit/coreinit_Time.cpp
```diff
@@ -28,7 +28,7 @@ namespace coreinit
 		osLib_returnFromFunction64(hCPU, osTime);
 	}
 
-	__declspec(noinline) uint64 coreinit_getTimeBase_dummy()
+	NOINLINE uint64 coreinit_getTimeBase_dummy()
```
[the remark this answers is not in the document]
tomlally: Thanks, done.

src/Cafe/HW/Latte/Renderer/Vulkan/RendererShaderVk.cpp
```diff
@@ -7,7 +7,7 @@
 #endif
 
 #include <glslang/Public/ShaderLang.h>
-#if GLSLANG_VERSION_LESS_OR_EQUAL_TO(11, 0, 0)
+#if 1 //GLSLANG_VERSION_LESS_OR_EQUAL_TO(11, 0, 0)
```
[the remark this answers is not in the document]
tomlally: Sorry, didn't see this one. That was blatant hacking, I'll undo this. The actual solution is to have the spirv-headers package installed.

src/Cafe/HW/Latte/Renderer/Vulkan/RendererShaderVk.cpp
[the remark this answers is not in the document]
tomlally: I think this was causing problems in clang-14. Will investigate further.

src/Common/precompiled.h
tomlally: I suppose `__clang__` or `__GNUC__` is somewhat redundant. Changed to just `__GNUC__` in https://github.com/cemu-project/Cemu/commit/71af6e54f44339ab9151c878335a0dfdd1bd1b05

src/Cafe/HW/Latte/Renderer/Vulkan/RendererShaderVk.cpp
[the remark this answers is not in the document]
tomlally: It was causing problems for some people. I was following advice in this comment. https://github.com/cemu-project/Cemu/issues/1#issuecomment-1225289477. `RendererShaderVk.cpp:134:18: error: excess elements in scalar initializer` was the error.

src/main.cpp
tomlally: Looks okay w/out. Removed here https://github.com/tomlally/Cemu/commit/3d28b5e0142ad641ac68887ceaff54adaf6c0ae3

src/Common/precompiled.h
tomlally: > 

I've gone ahead replaced _rotl64, _rotr64. Slightly different semantics since std::rot on type T also returns T but there's only one place where that matters: https://github.com/tomlally/Cemu/blob/6408cfdd4b391ca47f8296fe002cd1bf8adec37a/src/Cafe/HW/Latte/Core/LatteShader.cpp#L228

tomlally: > 



> > I only wrote these wrappers because _rotl64 and _rotr64 were used in the code and I wanted to get it compiling asap. Since then I've added `<bit>` to `precompiled.h` so maybe it's a good idea to just go through and replace all occurrences with the std functions. Thoughts?
> 
> Since we are on C++20, it is safe to use rotr or rotl rather than a custom implementation. And also, I found rotr and rotl to be more optimized in godbolt

I've gone ahead replaced _rotl64, _rotr64. Slightly different semantics since std::rot on type T also returns T but there's only one place where that matters: https://github.com/tomlally/Cemu/blob/6408cfdd4b391ca47f8296fe002cd1bf8adec37a/src/Cafe/HW/Latte/Core/LatteShader.cpp#L228

src/Common/precompiled.h
```diff
@@ -203,18 +220,52 @@ typedef union _LARGE_INTEGER {
...
+#elif defined(__GNUC__)
+    #define THREAD_LOCAL __thread
 #else
-#define DLLEXPORT __declspec(dllexport)
+    #define THREAD_LOCAL thread_local
 #endif
 
+#if defined(_MSC_VER)
+    #define DLLEXPORT __declspec(dllexport)
+    #define DLLIMPORT __declspec(dllimport)
```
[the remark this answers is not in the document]
tomlally: Thanks. Removed DLLIMPORT here https://github.com/cemu-project/Cemu/commit/3b5eb4946993064949f2f7bce6c74bc3e918458a

src/Cafe/HW/Latte/Renderer/Vulkan/RendererShaderVk.cpp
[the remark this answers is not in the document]
tomlally: Works w/out on gcc-12.2.0 and clang-14.0.6 (debug) so I've undone this change (see https://github.com/tomlally/Cemu/commit/056ae1f882726ed46cd3389f00538a47bfcb1b86). Is it possible the trailing comma was causing the error?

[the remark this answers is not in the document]
Crementif: Maybe this is caused by using outdated system packages if they were building with GCC, and thus not having the maxDualSourceDrawBuffersEXT installed?

Maybe all compilers would be happier if this was just a separate struct that we use for limits.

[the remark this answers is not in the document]
tomlally: That sounds quite likely. C++20 designated initialisers may be useful here?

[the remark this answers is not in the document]
Crementif: Maybe, but can be done in a later PR unless you wanna do another pass.

Crementif (approved):
LGTM

src/Cafe/HW/Latte/Renderer/Vulkan/RendererShaderVk.cpp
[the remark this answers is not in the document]
Tachi107: I'm facing the same error in https://github.com/cemu-project/Cemu/pull/75#issuecomment-1229451547 - do you have any suggestions?

src/Common/precompiled.h
Tachi107: As mentioned in the previous comment using `thread_local` the way it is currently used has no performance impact on GCC, while MSVC's documentation explicitly states that:

> On Windows, `thread_local` is functionally equivalent to the Microsoft-specific `__declspec( thread )` attribute.

See this for details: https://docs.microsoft.com/en-us/cpp/cpp/storage-classes-cpp?view=msvc-170#thread_local

bitscher:
The addition of the `-mavx2` flag is causing issues on pre-Haswell CPUs #144

src/Common/precompiled.h
tomlally: Thanks. https://github.com/cemu-project/Cemu/pull/101

tomlally:
`-mavx2` is required for all the `_mm256_...` functions.
`-mssse3` is required for `_mm_shuffle_epi8`

Other instructions used in Cemu:
  - sse `_mm_prefetch`, `_mm_pause`, `_mm_setcsr`
  - sse2 `_mm_mfence`, `_mm_set_epi8`, `_mm_set_epi16`, `_mm_loadu_si128`, `_mm_store_si128`
  - sse4.1 `_mm_min_epu16`, `_mm_max_epu16`

So it may actually be a good idea to add additional flags.

Tachi107:
What about adding `-march=x86-64-v2` and trying to stick to the [x86_64-v2 psABI](https://gitlab.com/x86-psABIs/x86-64-ABI/-/wikis/uploads/01de35b2c8adc7545de52604cc45d942/x86-64-psABI-2021-05-20.pdf#page=16)?

Exzap:
Both clang and gcc have a way to selectively enable CPU extensions for individual functions. [Here](https://godbolt.org/z/c9cP49Psc) is an example. It's better to do it this way then to set `-mavx2` or similar globally as Cemu doesn't have a hard requirement on anything beyond what every x86-64 CPU offers.

tomlally:
> Both clang and gcc have a way to selectively enable CPU extensions for individual functions. [Here](https://godbolt.org/z/c9cP49Psc) is an example. It's better to do it this way than to set `-mavx2` or similar globally as Cemu doesn't have a hard requirement on any of the optional CPU extensions.

As far as I can tell that would require runtime checks.
From https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#Common-Function-Attributes
> For instance, on an x86, you could declare one function with the target("sse4.1,arch=core2") attribute and another with target("sse4a,arch=amdfam10"). This is equivalent to compiling the first function with -msse4.1 and -march=core2 options, and the second function with -msse4a and -march=amdfam10 options. It is up to you to make sure that a function is only invoked on a machine that supports the particular ISA it is compiled for

Exzap:
> I agree, I can make a pr changing this if you wish

It would be appreciated.

> I don't think it would help insofar as compiling Cemu for pre-haswell CPUs on GCC.

Why not? As it says in your quote from the spec, you have to manually make sure the CPU extensions are supported at runtime before running the function with the intrinsics. This is something we already do. Optionally running the hand-optimized functions if the CPU extensions are supported and otherwise falling back to a generic implementation that can run on any CPU. The code for handling GPU drawcall indices is a good [example](https://github.com/cemu-project/Cemu/blob/68fa5b32a16d9c74dbd4c99854f00c84b62e2921/src/Cafe/HW/Latte/Core/LatteIndices.cpp#L685). On Windows this is all pretty well tested and people run Cemu with the most ancient of CPUs.

tomlally:
Here you go https://github.com/cemu-project/Cemu/pull/152

tomlally:
I must have missed these clang specific pragmas when I made #80. Sorry about that. https://github.com/cemu-project/Cemu/blob/68fa5b32a16d9c74dbd4c99854f00c84b62e2921/src/Cafe/HW/Latte/Core/LatteIndices.cpp#L288

PR #81, opened 2022-08-27
Added french translation to the shortcut comment.

Loatchi:


FakeMichau:
In my opinion it's reasonable to also translate the metainfo file at the same time

Loatchi:
Just did that, thanks a lot !

Exzap:
thanks!

PR #91, opened 2022-08-27
Nvidia Multithread Pipelines

niko1point0:
Allowing multithreaded pipelines on Nvidia, if driver version 515 or higher is found

### DIFF
--- a/src/Cafe/HW/Latte/Renderer/Vulkan/VulkanPipelineStableCache.cpp
+++ b/src/Cafe/HW/Latte/Renderer/Vulkan/VulkanPipelineStableCache.cpp
@@ -46,7 +46,7 @@ uint32 VulkanPipelineStableCache::BeginLoading(uint64 cacheTitleId)
 	m_numCompilationThreads = std::clamp(cpuCoreCount, 1u, 8u);
 	if (g_renderer->GetVendor() == GfxVendor::Nvidia)
 	{
-		if (VulkanRenderer::GetInstance()->GetDriverVersion() < 515.0f)
+		if(VulkanRenderer::GetInstance()->GetDriverVersion() < 515)
 		{
 			forceLog_printf("Disable multi-threaded pipeline loading due to an issue with Nvidia drivers");
 			m_numCompilationThreads = 1;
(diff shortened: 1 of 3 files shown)


src/Cafe/HW/Latte/Renderer/Vulkan/VulkanPipelineStableCache.cpp
```diff
@@ -46,8 +46,11 @@ uint32 VulkanPipelineStableCache::BeginLoading(uint64 cacheTitleId)
 	m_numCompilationThreads = std::clamp(cpuCoreCount, 1u, 8u);
 	if (g_renderer->GetVendor() == GfxVendor::Nvidia)
```
bitscher, line 47: If instead of saving the driver version you were simply saving a "m_disableMultithreadedCompilation" flag you could move out the vendor/driver-specific logic out and this file and simply this block further to
```c
if(VulkanRenderer::GetInstance()->GetDisableMultithreadedCompilation())
    m_numCompilationThreads = 1;
```

src/Cafe/HW/Latte/Renderer/Vulkan/VulkanRenderer.cpp
```diff
@@ -194,10 +195,22 @@ void VulkanRenderer::DetermineVendor()
...
 
 	forceLog_printf("Using GPU: %s", properties.properties.deviceName);
+
 	if (m_featureControl.deviceExtensions.driver_properties)
-		forceLog_printf("Driver version: %s", driverProperties.driverInfo)
+	{
+		forceLog_printf("Driver version: %s", driverProperties.driverInfo);
+
+		// needed for multithreaded pipelines on nvidia (requires 515 or higher)
+		driverVersion = StringHelpers::ToInt(std::string(driverProperties.driverInfo));
```
bitscher, line 204: m_disableMultithreadedCompilation = (StringHelpers::ToInt(std::string(driverProperties.driverInfo)) < 515);

src/Cafe/HW/Latte/Renderer/Vulkan/VulkanPipelineStableCache.cpp
Exzap: I agree with this suggestion. The driver version format differs by vendor so there isn't actually a nice way to have a singular `GetDriverVersion` function that makes sense.
The `disableMultithreadedCompilation` flag can be put into the `m_featureControl` member struct and it's overkill to wrap it in a getter. We don't do it with other options like `debugMarkersSupported`

niko1point0: Made this use GetDisableMultithreadedCompilation() as a getter, because m_featureControl was private

src/Cafe/HW/Latte/Renderer/Vulkan/VulkanRenderer.cpp
niko1point0: Done!

PR #95, opened 2022-08-28
nn_act: Don't swap the Mii name again

IntriguingTiles:
The name obtained from `GetMiiEx` is already swapped, so it shouldn't be swapped again. I've written a [minimal program](https://github.com/cemu-project/Cemu/files/9438841/miiname.rpx.zip) ([source](https://gist.github.com/IntriguingTiles/2d3d579d07a00eab2033596fec818942)) to test this that outputs the Mii name and the raw data of the Mii name in hex.

<details><summary>Screenshots</summary>
<p>
Previous behavior:

New behavior:

Behavior on console:

</p>
</details>

There are still some mystery 0x100s in the Mii name that shouldn't be there.

### DIFF
--- a/src/Cafe/Account/Account.cpp
+++ b/src/Cafe/Account/Account.cpp
@@ -101,6 +101,7 @@ Account::Account(uint32 persistent_id, std::wstring_view mii_name)
 	// set default name
 	FFLData_t* fflData = (FFLData_t*)m_mii_data.data();
 	const auto tmp_name = GetMiiName();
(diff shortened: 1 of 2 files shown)


Exzap:
Nice catch. The extra 0x0100 values are weird, I think there is some missing or incorrect zero initialization somewhere in Account.cpp
Do you want to look into this as well?

IntriguingTiles:
Works as expected now.
image

Exzap:
thanks!

PR #96, opened 2022-08-28
Compile Unix Public release with -O2

bitscher:


bitscher:
Actually the change seems to break LTO.

bitscher:
My bad `-O3` is already present when compiling with `-DPUBLIC_RELEASE` hence why it was breaking at link time due to LTO.

Exzap:
Optimization flags are already covered by using cmake CMAKE_BUILD_TYPE `Release` or `RelWithDebInfo`. It's not something you should manually override except maybe for non-standard flags like we do for MSVC.

bitscher:
I noticed that yes.
I am really not familiar with CMAKE :)

PR #101, opened 2022-08-28
Thread local cleanup

tomlally:
As discussed in https://github.com/cemu-project/Cemu/pull/80.
Removed the macro THREAD_LOCAL which expanded to `__declspec(thread)`, `__thread`, or `thread_local` depending on the compiler. Replaced all instances with `thread_local`.

From [MSDN](https://docs.microsoft.com/en-us/cpp/cpp/storage-classes-cpp?view=msvc-170#thread_local):
> On Windows, thread_local is functionally equivalent to the Microsoft-specific __declspec( thread ) attribute.


Exzap:
LGTM. Thanks

PR #102, opened 2022-08-28
Update .gitignore

n0toose:
I would normally sort the files alphabetically, but I intuitively chose to go with the file type, somewhat, instead.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,13 +17,16 @@
 build/
 out/
 .cache/
-bin/Cemu
 
 # Cemu bin files
+bin/Cemu
+bin/dump
+bin/graphicPacks
+bin/mlc01/*
 bin/otp.bin
 bin/seeprom.bin
+bin/keys.txt
 bin/Cemu.pdb
-bin/mlc01/*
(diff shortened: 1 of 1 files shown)


Tachi107:
While you're updating stuff, you might want to put a leading slash `/` to most paths (basically anything other than the ones starting with `*.`), otherwise git will ignore files matching the patterns in any directory.

n0toose:
Yeah, you're right about that.

n0toose:
I can't really see in which case e.g. the `/bin/Cemu` directory would be in any other place other than the root of the project if documentation is to be followed, though. Are you sure about tha?

Tachi107:
Yeah it generally shouldn't happen but I usually prefer to be explicit and avoid surprises (e.g. in the future Cemu might want to create a subdir named `bin/` in `src/`, and `src/bin/Cemu` will get ignored).

It's not important anyway, so you can simply leave things as they are. If the devs want to add `/` in the future they'll do it themself, I guess :)

Exzap:
We already have these covered on a branch. Will be merged soon

PR #110, opened 2022-08-29
Add pt_BR translation to desktop file and metainfo

baduhai:


dist/linux/info.cemu.Cemu.metainfo.xml
```diff
@@ -8,6 +8,7 @@
   <summary xml:lang="fr">Application pour émuler des jeux et applications Wii U sur PC</summary>
   <summary xml:lang="nl">Applicatie om Wii U spellen en applicaties te emuleren op PC</summary>
   <summary xml:lang="es">Software para emular juegos y aplicaciones de Wii U en PC</summary>
+  <summary xml:lang="es">Software para emular jogos e aplicativos de Wii U no PC</summary>
```
JakobDev, line 11:
```suggestion
  <summary xml:lang="pt_BR">Software para emular jogos e aplicativos de Wii U no PC</summary>
```

JakobDev:
Would it not better to have a general Portuguese translation instead of Brazilian Portuguese like the other translations? Or is the difference to big?

baduhai:
The difference is big enough to make some of the words I use sound awkward in European Portuguese, but not incomprehensible. I think it's better to have two different Portuguese translations, I'm just not familiar enough with European Portuguese to also contribute that translation.

baduhai:
Squashed commits

Exzap:
Thanks!

PR #111, opened 2022-08-29
Check Linux dist files in CI

JakobDev:
This PR adds a CI step to check if Linux dist files are valid.

In case you wondering why I use Arch: The AppStream spec has gotten a few new features since the last Ubuntu release. This is not a problem, since unknown elements will just be ignored by older AppStream versions, but the validation might fail. The validation has also been improved since them. Which Arch we always have the latest version.

The validation will currently fail, because the Linux dist files currently have a CRLF line separator while a LF separator is required.

### DIFF
--- a/.github/workflows/linux-dist-check.yml
+++ b/.github/workflows/linux-dist-check.yml
@@ -1,0 +1,37 @@
+name: Linux dist check
+
+on:
+  pull_request:
+    paths:
+      - "dist/linux/info.cemu.Cemu.desktop"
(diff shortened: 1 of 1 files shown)


Exzap:
Apologies for the late response on this. It's a sound suggestion but the dist files rarely ever get changed. They also get verified as part of the flatpak (and maybe appimage?) build. So in the end I think this would just add some bloat to the CI with very little practical benefit.

PR #115, opened 2022-08-29
Center README.md h1 and badges + fix GitHub casing

ghost:


README.md
```diff
@@ -1,7 +1,15 @@
...
+</div>
+
+<div align="center">
+<a href="https://github.com/cemu-project/Cemu/actions/workflows/build.yml">
+ <img src="https://github.com/cemu-project/Cemu/actions/workflows/build.yml/badge.svg" alt="Build Process"/>
+ </a>
+ <a href="https://discord.gg/5psYsup">
+  <img src="https://img.shields.io/discord/286429969104764928?label=Cemu&logo=discord&logoColor=FFFFFF" alt="Discord"/>
+ </a>
+</div>
```
Tachi107, line 12: Using HTML tags in Markdown files is not supported by a lot of parsers. Also, why did you replace Markdown-style images with HTML `` tags?

Exzap:
I think left-justified is fine, there is no reason to change it. Especially if it means we would have to use HTML.

PR #118, opened 2022-08-30
Bug #620 Fixed crash when no audio devices are present

krystainBooker:
[Bug #620](http://bugs.cemu.info/issues/620) Fixed crash when no audio devices are present

Steps to reproduce: 
1. In Window Sound manager, under Playback devices, disable all output devices
2. Start Cemu
3. Start any game
4. Cemu immediately closes

Updated functionality
1. When no audio devices are present on the PC, Cemu will no longer crash when starting a game.
2. Options-> General Settings ->  Audio, for both the APIs DirectSound and XAudio2, the TV/Device will now show as "Disabled" when no audio devices are present, previously it would display as "Primary Sound Driver"


Exzap:
LGTM
thanks!

PR #119, opened 2022-08-30
Fix various compiler warnings and update OpenSSL hash functions

bitscher:
`SHA1_Init`, `SHA1_Update`, `SHA1_Final`, `SHA256_Init`, `SHA256_Update`, `SHA256_Final` have been deprecated in OpenSSL 3.0 and will be removed in future versions.
- Replaced several simple calls to the `SHA1()` or `SHA256()` macro provided by OpenSSL
- Replaced other calls to deprecated functions with `EVP_DigestXXX` functions.
- Fixed an invalid sscanf format in DownloadManager
- Fixed unset return value  for `operator=` in VKRBase.h
- Fixed always-true checks for otpMem in `iosu_crypto`

### DIFF
--- a/src/Cafe/Account/Account.cpp
+++ b/src/Cafe/Account/Account.cpp
@@ -562,9 +562,9 @@ void makePWHash(uint8* input, sint32 length, uint32 magic, uint8* output)
 	buffer[2] = (magic >> 16) & 0xFF;
 	buffer[3] = (magic >> 24) & 0xFF;
 	memcpy(buffer + 8, input, length);
-	uint8 md[32];
+	uint8 md[SHA256_DIGEST_LENGTH];
 	SHA256(buffer, 8 + length, md);
-	memcpy(output, md, 32);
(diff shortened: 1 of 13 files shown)


Exzap:
Looks good to me. Thanks!

Tachi107:
@bitscher does this mean that the minimum required OpenSSL version is now 3.0.0?

bitscher:
@Tachi107  That shouldn't be the case, the `EVP_MD_ctx` and `EVP_Digest` interface also exists in the 1.1.1 branch https://www.openssl.org/docs/man1.1.1/man3/EVP_DigestInit.html

Tachi107:
> @Tachi107  That shouldn't be the case, the `EVP_MD_CTX` and `EVP_Digest` interface also exists in the 1.1.1 branch https://www.openssl.org/docs/man1.1.1/man3/EVP_DigestInit.html

OK, thanks for the info :)

PR #121, opened 2022-08-30
Misc. Linux improvements and bug fixes.

tomlally:
Various improvements for compiling on Linux, including:
  - **Fixing the argument order in `__cpuid`**
  - Removing unused variables (`g_isGameRunning`) 
  - Disabling some compiler warnings (`multichar`, `invalid-offsetof`)
  - Removing `-mssse` and `-mavx` on clang
  - Removing some flags (`-fms-extensions`, `-fpermissive`) and some small changes to the source to compile without.
      - Replacing MSVC intrinsics with c++20 std functions. (`_rotl`, `_rotr`)
      - Using `static_cast` for narrowing conversions.
      - https://github.com/cemu-project/Cemu/commit/ad735f019c473e1f14e22a545f982984e77fb01c#diff-8082b4ce62591b8559436956fdfa1255bff3b4567f75d910afef80599d32aa4c


Crementif (approved):
Seems good to me, @Exzap can decide to

tomlally:
> Oh, I did note something while reviewing your other PR. But could you use uint32 instead of unsigned int and the like?
> 
> Also, while you're making changes could you also std::rotr with a templated type too just for safety, so e.g. `std::rotl<uint32>(value, 3)` instead of `std::rotl(value, 3)`.
> 
> If you want I'm also fine if you adjust it in #137 and then we just merge that since that supersedes this PR.

Sure. The rationale was to match [these](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/rotl-rotl64-rotr-rotr64?view=msvc-170) exactly. Is `uint32` or `uint32_t` preferred? Before merging #137 - I'm not yet sure if the CPU use is reading correctly on windows.

Exzap:
> Is uint32 or uint32_t preferred?

Always without _t in our codebase. See the `// base types` comment in precompiled.h for the full list of the basic types we use

PR #52, opened 2022-08-25
Initial macOS port bringup

m-chojnacki:
Hello!

This is initial bringup of macOS port. It's _almost_ everything needed in order to build the Cemu under macOS, and run the wxWidgets GUI.

Support for rendering is still missing, but as long as you are fine with mainlining this port, I will commit to brinbugs ging the renderers (Vulkan over MoltenVK? native Metal renderer?) to macOS allowing users to run actual games.

Some possibly questionable design decisions made here:
- renamed "linux" resources to "embedded" as they are actually platform independent,
- renamed "linux" platform to "unix" as it'll work fine for other unixes too,
- extracted X11 headers from new "unix" platform and moved it to platform ifdef.

Also, this PR a bugs on other platforms as well, just affecting macOS more:
- Fixed use of string_view that treats the second parameter as "end pointer" not as "length", causing out-of-bounds read on macOS, and apparently not mattering much on other platforms.

As a bonus, Cemu UI running on macOS
Zrzut ekranu 2022-08-25 o 22 22 09

I'm excited to be able to contribute to the project, and I'm looking forward to any feedback 😊

PR #124, opened 2022-08-30
Initial implementation of macOS Vulkan renderer over MoltenVK

m-chojnacki:
Hi there!

As I promised here #52, this PR brings initial support for Vulkan over MoltenVK renderer on macOS.

This is enough to actually run some games to some extent. For now, I got a game to the stage where splash screen and loading screen appears, and actually works.

Some additional tinkering would be definitely necessary for the renderer to work on more non-trivial scenarios, but I guess it's a good step forward!

Geometry shaders and logic ops are explicitly turned off, so if any shader would emit code containing those features, it'll probably break. I believe we can work around this in future.

Some screenshots for you:
Zrzut ekranu 2022-08-30 o 16 46 09
Zrzut ekranu 2022-08-30 o 16 46 17
Zrzut ekranu 2022-08-30 o 16 46 31

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -89,7 +89,7 @@ if (NOT TARGET glslang::SPIRV AND TARGET SPIRV)
 	add_library(glslang::SPIRV ALIAS SPIRV)
 endif()
 
-if (UNIX)
+if (UNIX AND NOT APPLE)
 	find_package(X11 REQUIRED)
 endif()
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -20,7 +20,11 @@ if(MSVC)
     # _SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING
     # _SILENCE_ALL_CXX17_DEPRECATION_WARNINGS
 elseif(UNIX)
-    if(NOT APPLE)
+    if(APPLE)
+        add_definitions(-D_XOPEN_SOURCE)
+        add_definitions(-DVK_USE_PLATFORM_MACOS_MVK)
+        add_definitions(-DVK_USE_PLATFORM_METAL_EXT)
+    else()
         add_definitions(-DVK_USE_PLATFORM_XLIB_KHR) # legacy. Do we need to support XLIB surfaces?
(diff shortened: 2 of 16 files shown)


m-chojnacki:
Thanks a lot for the tip @Crementif, I'll try to acquire a physical copy 😉

Exzap:
The implementation is very clean. Good job!

msoham123:
https://github.com/KhronosGroup/MoltenVK/issues/1670

It seems like logic ops functionality does indeed exist in Metal and by extent Moltenvk, but it is hidden behind a private API (most likely to support the OpenGL driver).

Maybe this feature will have to be emulated unless Apple makes these API's public or the Cemu team is okay with using an unsafe api (Apple could remove at any point).

I think geometry shader support is planned with the release of Metal 3: https://github.com/KhronosGroup/MoltenVK/discussions/1616.

tellowkrinkle:
> It seems like logic ops functionality does indeed exist in Metal and by extent Moltenvk, but it is hidden behind a private API (most likely to support the OpenGL driver). https://github.com/KhronosGroup/MoltenVK/issues/1670

As I said on one of the other MVK issues, there's not much point to using that private API.  It offers no benefit over framebuffer fetch, which is a 100% public and supported API.  If you want logic ops, implement them with fbfetch.  It'll work on the same set of GPUs that implement logic op support in Metal (Intel Skylake+ and Apple), with the benefit of the Apple GPU support being official and guaranteed to not go away or break in the future (Intel support is not official).

msoham123:
> > It seems like logic ops functionality does indeed exist in Metal and by extent Moltenvk, but it is hidden behind a private API (most likely to support the OpenGL driver). [KhronosGroup/MoltenVK#1670](https://github.com/KhronosGroup/MoltenVK/issues/1670)
> 
> As I said on one of the other MVK issues, there's not much point to using that private API. It offers no benefit over framebuffer fetch, which is a 100% public and supported API. If you want logic ops, implement them with fbfetch. It'll work on the same set of GPUs that implement logic op support in Metal (Intel Skylake+ and Apple), with the benefit of the Apple GPU support being official and guaranteed to not go away or break in the future (Intel support is not official).

Yep, this is a far better approach than a private api for sure.

Issue #117, opened 2022-08-29
Arch Linux: PKGBUILD


Anuskuss:
After #1 has been closed I'm trying to not pollute the continuation thread #107 with Arch specific issues so I decided to create my own thread. I am aware that there's already [cemu-git](https://aur.archlinux.org/packages/cemu-git) on the AUR, however I have some of my own ideas and the author seems to be inactive on GitHub.

<details>
<summary>Old PKGBUILD (full vcpkg)</summary>

[History](https://github.com/cemu-project/Cemu/issues/1#issuecomment-1226610562)
```shell
pkgname=cemu-git
pkgver=2.0.49
pkgrel=1
pkgdes='Cemu is a Wii U emulator'
arch=(x86_64)
url=https://github.com/cemu-project/Cemu
licence=(MPL2)
depends=(
	# vcpkg/wxwidgets
	gtk3 libgcrypt systemd-libs
)
makedepends=(
	# temporary
	patch
	# cemu
	nasm
	# pkgbuild
	git cmake make
	# vcpkg
	curl zip unzip tar python pkgconf # linux-headers
	# vcpkg/wxwidgets
	glu # libsecret
)
optdepends=(
	'vulkan-driver: Vulkan graphics'
)
provides=(cemu)
conflicts=(cemu)
source=(
	git+https://github.com/cemu-project/Cemu#commit=0f24b0663efb677a3e7d7ffd14a569fb0e366c35
	git+https://github.com/Exzap/ZArchive.git
	git+https://github.com/mozilla/cubeb.git
	git+https://github.com/microsoft/vcpkg.git
	# cubeb dependencies
	git+https://github.com/arsenm/sanitizers-cmake.git
	git+https://github.com/google/googletest.git
	git+https://github.com/KhronosGroup/Vulkan-Headers.git
)
sha256sums=(
	SKIP
	SKIP
	SKIP
	SKIP
	SKIP
	SKIP
	SKIP
)

pkgver() {
	cd Cemu
	MAJ=$(awk -F '\t' '/LEAD/ {print $NF}' src/Common/version.h)
	MIN=$(awk -F '\t' '/MAJOR/ {print $NF}' src/Common/version.h)
	CNT=$(git rev-list --count HEAD)
	# HASH=$(git rev-parse --short HEAD)
	echo "${MAJ}.${MIN}.${CNT}"
}

prepare() {
	# vcpkg cache
	mkdir -p "$startdir/.vcpkg"
[...]

PR #112, opened 2022-08-29
build: minor refactoring and fixes

Tachi107:
Some bits originated from #75

- Fix `target_precompile_headers()` usage; the CemuCommon target exposes the `src/Common/precompiled.h` precompiled header as part of its public interface with `target_precompile_headers(CemuCommon PUBLIC precompiled.h)`, so all the other targets wanting to use the precompiled header have to link to the CemuCommon target with `target_precompile_headers(TargetName PRIVATE CemuCommon)`.
- Set the project version to 2.0
- Set `RUNTIME_OUTPUT_DIRECTORY` instead of only their `_DEBUG` and `_RELEASE` variants, fixing the compilation when neither build types are defined (see https://github.com/cemu-project/Cemu/issues/1#issuecomment-1229359719)
- Use a consistent indentation style (tabs, like in the .cpp files)
- Use "modern" variants of some functions, e.g. `add_definitions` -> `add_compile_definitions`
- Use `std::enable_if` & co instead of `boost::`

Issue #107, opened 2022-08-29
State of Linux (Cont)


Exzap:
This is a continuation of #1 

The current top priorities for our Linux build are:

- [x] Providing AppImage releases
- [x] Wayland support (at least for Vulkan)
- [x] Finding a way to correctly emulate Wii U's case-insensitive filesystem on case-sensitive host filesystems
- [ ] Fixing various niche or distro specific bugs and crashes (see individual reported issues for these)

There are also some features that haven't been ported over yet:
- [ ] HID support (Wii U nsyshid.rpl driver). Required for USB portals (Skylanders Portal of Power, Lego Dimension etc.) (Currently work in progress via libusb)
- [x] nlibcurl. Has been ported but not tested a lot
- [x] Some debugger features like memory breakpoints

Nice to have:
- [x] Providing flatpak releases

(List last updated on 13-01-2023)

When building from source we use [vcpkg](https://github.com/microsoft/vcpkg) for managing dependencies on all platforms and our officially supported and recommended compiler is Clang 12 or higher (and MSVC on Windows). With the `-DENABLE_VCPKG=OFF` option you can use system libraries instead of vcpkg but we generally discourage it as we prefer having builds with the same exact library configuration across all platforms for simplicity of debugging. We also won't officially support distributing Cemu as a distro package. Of course package maintainers are free to package it, but our focus is to streamline the troubleshooting and bug reporting process and it's way easier if there aren't dozens of different builds.

Lastly, when making a new Linux-specific issue try to name it `[Linux] <Bug description>`.

Tachi107:
> Finding a way to correctly emulate Wii U's case-insensitive filesystem on case-sensitive host filesystems

ext4 has case-folding support. Would it be possible to make it a requirement for Cemu? Users would have to store stuff accessed by the Wii U in a separate volume with this feature enabled though, which could be hard to do.

JakobDev:
> ext4 has [case-folding](https://www.collabora.com/news-and-blog/blog/2020/08/27/using-the-linux-kernel-case-insensitive-feature-in-ext4/) support. Would it be possible to make it a requirement for Cemu? Users would have to store stuff accessed by the Wii U in a separate volume with this feature enabled though, which could be hard to do.

That's a very bad idea. This requires, as you said, a seprate volume. Users might want to use a case-sensitive  filesystem, so they need a own partition/drive just for Cemu.

[...]

PR #126, opened 2022-08-30
Add ability to save/read files based on XDG specification

Tatsh:
This adds the option `ENABLE_XDG_DIRS` to enable following the XDG specification as closely as possible.

This also allows relocating the binary so that it does not have be adjacent to the resources, etc. It can be at `/usr/bin/cemu` as long as `-DSYSTEM_DATA_PATH=...` is used when configuring and the resources are stored in that path (items in `bin/`).

Regarding logs, the XDG specification says to use the state directory but I have never seen a single app do this, so logs go in the `XDG_DATA_HOME`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,6 +50,13 @@ option(ENABLE_OPENGL "Enables the OpenGL backend" ON)
 option(ENABLE_VULKAN "Enables the Vulkan backend" ON)
 option(ENABLE_DISCORD_RPC "Enables the Discord Rich Presence feature" ON)
 
+if (UNIX)
+  option(ENABLE_XDG_DIRS "Enable storage according to XDG specification" ON)
+endif()
+if (SYSTEM_DATA_PATH)
+  add_compile_definitions(SYSTEM_DATA_PATH=\"${SYSTEM_DATA_PATH}\")
+endif()
+
 # input backends
 if (WIN32)
   option(ENABLE_XINPUT "Enables the usage of XInput" ON)
@@ -123,6 +130,10 @@ if (ENABLE_CUBEB)
 		add_library(cubeb::cubeb ALIAS cubeb)
 	endif()
 	add_compile_definitions("HAS_CUBEB=1")
+endif()
+
+if (ENABLE_XDG_DIRS)
+  add_compile_definitions(XDG)
 endif()
 
 add_subdirectory("dependencies/ih264d")
--- a/src/config/ActiveSettings.cpp
+++ b/src/config/ActiveSettings.cpp
@@ -18,7 +18,7 @@ void ActiveSettings::LoadOnce()
 	s_path = s_full_path.parent_path();
 	s_filename = s_full_path.filename();
 
-	g_config.SetFilename(GetPath("settings.xml").generic_wstring());
+	g_config.SetFilename(GetConfigPath("settings.xml").generic_wstring());
 	g_config.Load();
 
 	std::wstring additionalErrorInfo;
@@ -92,7 +92,7 @@ GraphicAPI ActiveSettings::GetGraphicsAPI()
 	// check if vulkan even available
 	if (api == kVulkan && !g_vulkan_available)
 		api = kOpenGL;
-	
+
 	return api;
 }
 
@@ -220,8 +220,59 @@ fs::path ActiveSettings::GetMlcPath()
 	return GetDefaultMLCPath();
 }
 
-fs::path ActiveSettings::GetDefaultMLCPath()
+#ifdef XDG
+std::string ActiveSettings::GetXDGPath(const char* envVar, const std::string& defaultValue)
 {
-	return GetPath("mlc01");
+	auto raw_value = std::getenv(envVar);
+	return !raw_value || std::strlen(raw_value) == 0 ? defaultValue : std::string(raw_value);
+}
+#endif
+
+fs::path ActiveSettings::GetConfigPath()
+{
+#ifdef XDG
+	auto config_home = GetXDGPath("XDG_CONFIG_HOME", fmt::format("{}/.config", std::getenv("HOME")));
+	auto dir = fs::path(fmt::format("{}/cemu", config_home));
+	if (!fs::exists(dir))
+		fs::create_directories(dir);
+	return dir;
+#endif
+	return s_full_path;
 }
 
+fs::path ActiveSettings::GetCachePath()
+{
+#ifdef XDG
+	auto config_home = GetXDGPath("XDG_CACHE_HOME", fmt::format("{}/.cache", std::getenv("HOME")));
+	auto dir = fs::path(fmt::format("{}/cemu", config_home));
+	if (!fs::exists(dir))
+		fs::create_directories(dir);
+	return dir;
+#endif
+	return s_full_path;
+}
+
+fs::path ActiveSettings::GetDataPath()
+{
+#ifdef XDG
+	auto config_home = GetXDGPath("XDG_DATA_HOME", fmt::format("{}/.local/share", std::getenv("HOME")));
+	auto dir = fs::path(fmt::format("{}/cemu", config_home));
+	if (!fs::exists(dir))
+		fs::create_directories(dir);
+	return dir;
+#endif
+	return s_full_path;
+}
+
+fs::path ActiveSettings::GetSystemDataPath()
+{
+#ifdef SYSTEM_DATA_PATH
+	return fs::path(SYSTEM_DATA_PATH);
+#endif
+	return s_full_path;
+}
+
+fs::path ActiveSettings::GetDefaultMLCPath()
+{
+	return GetDataPath("mlc01");
+}
--- a/src/Cafe/Account/Account.cpp
+++ b/src/Cafe/Account/Account.cpp
@@ -424,7 +424,7 @@ OnlineValidator Account::ValidateOnlineFiles() const
 {
 	OnlineValidator result{};
 	
-	const auto otp = ActiveSettings::GetPath("otp.bin");
+	const auto otp = ActiveSettings::GetDataPath("otp.bin");
 	if (!fs::exists(otp))
 		result.otp = OnlineValidator::FileState::Missing;
 	else if (fs::file_size(otp) != 1024)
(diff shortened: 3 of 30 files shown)


CMakeLists.txt
```diff
@@ -50,6 +50,13 @@ option(ENABLE_OPENGL "Enables the OpenGL backend" ON)
 option(ENABLE_VULKAN "Enables the Vulkan backend" ON)
 option(ENABLE_DISCORD_RPC "Enables the Discord Rich Presence feature" ON)
 
+if (UNIX)
+  option(ENABLE_XDG_DIRS "Enable storage according to XDG specification" ON)
+endif()
+if (SYSTEM_DATA_PATH)
+  add_definitions(-DSYSTEM_DATA_PATH=\"${SYSTEM_DATA_PATH}\")
```
Tachi107, line 57:
```suggestion
  add_compile_definitions(SYSTEM_DATA_PATH=\"${SYSTEM_DATA_PATH}\")
```

CMakeLists.txt
```diff
@@ -125,6 +132,10 @@ if (ENABLE_CUBEB)
 	add_compile_definitions("HAS_CUBEB=1")
 endif()
 
+if (ENABLE_XDG_DIRS)
+  add_definitions(-DXDG)
```
Tachi107, line 136:
```suggestion
  add_compile_definitions(XDG)
```

src/config/ActiveSettings.cpp
```diff
@@ -220,8 +220,63 @@ fs::path ActiveSettings::GetMlcPath()
...
+    auto config_home = std::getenv("XDG_CONFIG_HOME");
+    auto config_home_s = std::string();
+    if (!config_home) {
+        config_home_s = fmt::format("{}/.config", std::getenv("HOME"));
+    }
+    auto dir = fs::path(fmt::format("{}/cemu", config_home ? config_home : config_home_s));
+    if (!fs::exists(dir)) {
+      fs::create_directories(dir);
+    }
+    return dir;
```
Tachi107, line 234: `std::getenv("XDG_CONFIG_HOME")` will evaluate to false if `XDG_CONFIG_HOME` is unset (as `getenv` returns `nullptr` in this case), but what if it is set to an empty string?

Also, you are checking if `config_home` is set twice. What about:

```suggestion
	auto config_home = std::getenv("XDG_CONFIG_HOME");
	auto dir = fs::path(fmt::format(
		"{}/cemu",
		config_home ? config_home : fmt::format("{}/.config", std::getenv("HOME"))
	));
	if (!fs::exists(dir)) {
		fs::create_directories(dir);
	}
	return dir;
```

src/config/ActiveSettings.cpp
```diff
@@ -220,8 +220,63 @@ fs::path ActiveSettings::GetMlcPath()
...
+    auto cache_home_env = std::getenv("XDG_CACHE_HOME");
+    auto cache_home_s = std::string();
+    if (!cache_home_env) {
+        cache_home_s = fmt::format("{}/.cache", std::getenv("HOME"));
+    }
+    auto dir = fs::path(fmt::format("{}/cemu", cache_home_env ? cache_home_env : cache_home_s));
+    if (!fs::exists(dir)) {
+      fs::create_directories(dir);
+    }
+    return dir;
```
Tachi107, line 250: Same as above

src/config/ActiveSettings.cpp
```diff
@@ -220,8 +220,63 @@ fs::path ActiveSettings::GetMlcPath()
...
+    auto data_home = std::getenv("XDG_DATA_HOME");
+    auto data_home_s = std::string();
+    if (!data_home) {
+        data_home_s = fmt::format("{}/.local/share", std::getenv("HOME"));
+    }
+    auto dir = fs::path(fmt::format("{}/cemu", data_home ? data_home : data_home_s));
+    if (!fs::exists(dir)) {
+      fs::create_directories(dir);
+    }
+    return dir;
```
Tachi107, line 266: Same as above

Tachi107 (changes requested):
Thanks for your work on this! I have a few minor concerns though.

src/config/ActiveSettings.cpp
Tatsh: Should probably make a wrapper function that handles these cases and always returns `std::string`.

Tachi107: I guess that a simple function like this should be enough

```cpp
std::string get_xdg_dir(const std::string_view name) {
	std::string config_home = std::getenv(name);
	if (config_home.empty()) {
		config_home = fmt::format("{}/.config", std::getenv("HOME"));
	}
	return config_home;
}
```

Crementif (changes requested):
I've got some notes and corrections (plus it seems like the indentation settings you used is set wrong).

I do feel conflicted about merging these changes, so I'll let @Exzap decide if he wants to merge this since these are definitely good changes (thanks!) and are also improve the flatpak support we eventually wanna provide (afaik). But my personal worries are that this is a pretty big departure from Cemu's portable nature on both Windows and Linux and it's mostly aimed for the unofficial packaging efforts that exist atm.

I've already spoken about it in https://github.com/cemu-project/Cemu/issues/117#issuecomment-1231900267, but to explain why I'd rather wait with this PR for now is because if we don't make our intentions very clear and put at least a damper on this whole thing, it seems like the packaging efforts will continue and that end-users will just end up having a bad experience with an super-early unofficial Cemu package which'll cause them to write Cemu off as a buggy, incompatible experience.

I'd much rather have a focus being put on core issues that the Linux build has at the moment (see our list at the top of #107) and the regressions that Cemu 2.0 has from the previous versions due to all the reworks.

Tatsh:
I am in no hurry. Users of my overlay understand all the software, especially emulators, is experimental.

Tatsh:
> I've got some notes and corrections (plus it seems like the indentation settings you used is wrong).
> 
> I do feel conflicted about merging these changes, so I'll let @Exzap decide if he wants to merge this since these are definitely good changes (thanks!) and are also improve the flatpak support we eventually wanna provide (afaik). But my personal worries are that this is a pretty big departure from Cemu's portable nature on both Windows and Linux and it's mostly aimed for the unofficial packaging efforts that exist atm.
> 
> I've already spoken about it in [#117 (comment)](https://github.com/cemu-project/Cemu/issues/117#issuecomment-1231900267), but to explain why I'd rather wait with this PR for now is because if we don't make our intentions very clear and put at least a damper on this whole thing, it seems like the packaging efforts will continue and that end-users will just end up having a bad experience with an super-early unofficial Cemu package which'll cause them to write Cemu off as a buggy, incompatible experience.
> 
> I'd much rather have a focus being put on core issues that the Linux build has at the moment (see our list at the top of #107) and the regressions that Cemu 2.0 has from the previous versions due to all the reworks.

This change should not affect current builds as the original portable functionality remains as long as `-DENABLE_XDG_DIRS` is not passed (and `-DENABLE_XDG_DIRS` is only available `if (UNIX)`). The goal is to have XDG be entirely optional. If this PR does not have the original behaviour when `-DENABLE_XDG_DIRS` is not passed, then I need to work on it more.

In the long term I do think you should offer both an installer and a portable version for Windows. Windows has equivalent directories for XDG. People only need to be concerned with where their games are generally. I am not sure why they need to be concerned about where the mlc goes? I am considering removing that option altogether in my Gentoo version.

Anuskuss:
@Tatsh Putting the `bin` stuff into `SYSTEM_DATA_PATH` will also put `gameProfiles` in there, which will prevent a user launching cemu from creating game profiles (as `SYSTEM_DATA_PATH/gameProfiles` belongs to `root`). Can multiple `gameProfiles` be supported or should I just remove it from `SYSTEM_DATA_PATH` and put it in `$XDG_DATA_HOME/cemu`?

> I'd rather wait with this PR for now is because if we don't make our intentions very clear and put at least a damper on this whole thing, it seems like the packaging efforts will continue and that end-users will just end up having a bad experience with an super-early unofficial Cemu package

@Crementif I know you mean well but this is such a short sighted view on the thing. Just don't provide binaries until you think it's ready. Compiling alone is a good enough deterrent for the Linux newcomers. Also, to be honest, regardless what will happen to this PR, the Linux community efforts will continue anyway. The same thing happend when Cemu was Windows-only - the community found a way (through Wine). Even "against the intent" of the developers if you will.

Tatsh:
> @Tatsh Putting the `bin` stuff into `SYSTEM_DATA_PATH` will also put `gameProfiles` in there, which will prevent a user launching cemu from creating game profiles (as `SYSTEM_DATA_PATH/gameProfiles` belongs to `root`). Can multiple `gameProfiles` be supported or should I just remove it from `SYSTEM_DATA_PATH` and put it in `$XDG_DATA_HOME/cemu`?

I thought dist game profiles were in `bin/gameProfiles/default` and the ones created by user were in `bin/gameProfiles` (normally `gameProfiles` in portable install). The code is patched here to read default profiles from `SYSTEM_DATA_PATH` and read user-made/installed profiles from `$XDG_DATA_HOME/cemu`.

Anuskuss:
Whoops, I meant `$XDG_CONFIG_HOME/cemu/gameProfiles` of course. They are properly created there, just not read. Maybe I'm doing something wrong though, will investigate more later.

Crementif:
> Users of my overlay understand all the software, especially emulators, is experimental.

Cemu 2.0 has an unfinished Linux port that'll take probably at least a month to be usable at the current pace. It's not just "experimental". The current Linux "build" is **only** useful for devs who want to contribute fixes/features or by people who've got a working wine installation and just want to give it a shot.

> In the long term I do think you should offer both an installer and a portable version for Windows. 

Yes, like I said, we eventually wanna implement this optional but still major breaking change. Also, this would require more work then just a built-time option, and I don't see this Linux PR making any Windows changes. But for now this seems to me like it is just upstreaming fixes for a packaging script which we don't want to support atm.

Tachi107:
I agree with @Crementif here. While I'd love to package Cemu for Debian (and derivatives like Ubuntu), it is just not ready yet. I'm doing my best to submit PRs that improve the CMake side of things (#75, #112, and more to come...), as I've worked with it a lot, but packaging it now feels, to me, like a mistake; the Linux side of things needs to improve first. Not by patching in distro packages, but by collaborating with upstream and figuring out the best way to handle the existing problems.

PR #129, opened 2022-08-31
Fix build using clang-cl on windows and improve driver detection

jambonmcyeah:


CMakeLists.txt
```diff
@@ -39,7 +39,13 @@ endif()
 if (MSVC)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /EHsc")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /fp:precise") # floating point model: precise
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /GT") # fiber safe optimizations
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /arch:AVX2")
```
bitscher, line 42: The dyanmic recompiler seems to use instruction set extensions based on their availability.
Unless see a measurable performance improvement by compiling the C++ code with AVX2, this probably should not be made a default flag.

jambonmcyeah: fixed

src/Cafe/HW/Latte/Renderer/Vulkan/VulkanRenderer.cpp
```diff
@@ -197,32 +197,62 @@ void VulkanRenderer::DetermineVendor()
...
-		if(m_vendor == GfxVendor::Nvidia)
-		{
-			// if the driver does not support the extension,
-			// it is assumed the driver is under version 515
-			m_featureControl.disableMultithreadedCompilation = true;
-		}
+	}
+}
+
+const std::tuple<uint16, uint16, uint16, uint16> VulkanRenderer::GetDriverVersion()
```
Exzap, line 229: Do we really need GetDriverVersion()? The only necessary use-case is for the Nvidia pipeline check which is already a solved problem. We really don't want to encourage more conditional logic to be built on top of the driver version, especially since vendors are free to change their versioning and driverID encoding at any moment.

src/asm/CMakeLists.txt
```diff
@@ -3,10 +3,12 @@ project(CemuAsm C)
 IF (WIN32)
 
 enable_language(C ASM_MASM)
+set(CMAKE_ASM_MASM_CREATE_STATIC_LIBRARY "<CMAKE_AR> /OUT:<TARGET> <LINK_FLAGS> <OBJECTS>")
```
Exzap, line 6: Since this is a workaround please add a comment explaining what's it for and make it conditional to the affected environment

Exzap (changes requested):

src/Cafe/HW/Latte/Renderer/Vulkan/VulkanRenderer.cpp
jambonmcyeah: removed it

src/asm/CMakeLists.txt
jambonmcyeah: added

jambonmcyeah:
fixed everything

src/CMakeLists.txt
```diff
@@ -95,7 +95,7 @@ target_link_libraries(CemuBin PRIVATE
 )
 
 target_link_libraries(CemuBin PRIVATE CemuAsm)
-target_link_libraries(CemuBin PRIVATE SDL2::SDL2 SDL2::SDL2main) # is SDL2main needed?
+target_link_libraries(CemuBin PRIVATE SDL2::SDL2) # is SDL2main needed?
```
Tachi107, line 98: Why was SDL2main removed? If you're getting build failures because CMake can't find the target, it is a known issue that affects the latest SDL2 release, see https://github.com/cemu-project/Cemu/pull/75#issuecomment-1232851866

jambonmcyeah: I was getting link errors with lld-link because multiple main has been defined

According to the docs, since `SDL_MAIN_HANDLED` is set in `main.cpp` and `SDL_SetMainReady()` is called in our main function we shouldn't need to link SDL2main

Tachi107: Ok, thanks for the clarification! I was also unsure if SDL2main was actually needed, hence the comment in CMakeLists.txt. Since you removed the library, could you also remove the comment? Thanks :)

jambonmcyeah: sure

jambonmcyeah:
everything should be fixed

CMakeLists.txt
```diff
@@ -39,7 +39,13 @@ set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 if (MSVC)
 	set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY VS_STARTUP_PROJECT CemuBin)
 	# floating point model: precise, fiber safe optimizations
-	add_compile_options(/EHsc /fp:precise /GT)
+	add_compile_options(/EHsc /fp:precise)
+	if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+		add_compile_options(/clang:-maes)
```
bitscher, line 44: What's the purpose of adding the -maes flag?

bitscher: I see it exists in src/CMakeLists.txt as well, so it makes sense to have it reflected here.
What requires it though?

jambonmcyeah: compiling used to fail because of missing aes intrinsics due to missing that flag
I just tested again and it no longer fails so I'm going to remove it.

CMakeLists.txt
```diff
@@ -39,7 +39,12 @@ set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 if (MSVC)
 	set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY VS_STARTUP_PROJECT CemuBin)
 	# floating point model: precise, fiber safe optimizations
-	add_compile_options(/EHsc /fp:precise /GT)
+	add_compile_options(/EHsc /fp:precise)
+	if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+		set(CMAKE_STATIC_LINKER_FLAGS "${CMAKE_STATIC_LINKER_FLAGS} /llvmlibthin")
```
Tachi107, line 44: Could you please add a comment explaining why this is needed? Also, prefer using target options like [`STATIC_LIBRARY_OPTIONS`](https://cmake.org/cmake/help/latest/prop_tgt/STATIC_LIBRARY_OPTIONS.html) to modifying global `CMAKE_*` variables

bitscher: [`add_link_options` ](https://cmake.org/cmake/help/latest/command/add_link_options.html#command:add_link_options) seems to exist.

Tachi107: Yes, but it doesn't change options for static libraries since they are not "linked" (or that's what CMake's documentation seems to suggest)

jambonmcyeah: done

jambonmcyeah: > Yes, but it doesn't change options for static libraries since they are not "linked" (or that's what CMake's documentation seems to suggest)

yeah they aren't linked, CMake uses this term for the archiving step meaning this sets the flags to `lib.exe` on windows and `ar` on unix
guess the term is a little confusing

CMakeLists.txt
```diff
@@ -39,7 +39,13 @@ set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 if (MSVC)
 	set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY VS_STARTUP_PROJECT CemuBin)
 	# floating point model: precise, fiber safe optimizations
-	add_compile_options(/EHsc /fp:precise /GT)
+	add_compile_options(/EHsc /fp:precise)
+	if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+		# Speeds up static linking (especially helpful in incremental compilation)
+		set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY STATIC_LIBRARY_OPTIONS /llvmlibthin)
```
Tachi107, line 45: I did a quick search and it seems that `/llvmlibthin` is only supported when using the LLD linker with the link.exe on Windows. Does using clang-cl necessarily mean that one is also using LLD? Is it possible to use clang-cl with MSVC's link.exe?

jambonmcyeah: good catch!
it's fixed now.

CMakeLists.txt
```diff
@@ -132,4 +140,4 @@ if (NOT ZArchive_FOUND)
 	add_subdirectory("dependencies/ZArchive" EXCLUDE_FROM_ALL)
 endif()
 
-add_subdirectory(src)
+add_subdirectory(src)
```
Tachi107, line 143: It seems that the ending newline was deleted, and Git is unhappy

src/Cafe/HW/Latte/Renderer/Vulkan/VulkanRenderer.cpp
```diff
@@ -1106,7 +1108,7 @@ bool VulkanRenderer::CheckDeviceExtensionSupport(const VkPhysicalDevice device,
 	{
 		requiredExtensions.erase(extension.extensionName);
 	}
-
+
```
Tachi107, line 1111: It isn't clear by the diff, but it seems that here you added a bit of unneeded whitespace

Tachi107 (commented):
A few whitespace-related comments. Also, was src/resource/cemu.rc deleted on purpose?

Exzap:
> Also, was src/resource/cemu.rc deleted on purpose?

I got confused by that too, but it's not deleted. They removed some unused `#ifdef` and changed the encoding from utf16 to utf8 making the file much smaller.

jambonmcyeah:
> > Also, was src/resource/cemu.rc deleted on purpose?
> 
> I got confused by that too, but it's not deleted. They removed some unused `#ifdef` and changed the encoding from utf16 to utf8 making the file much smaller.

I changed this because clang doesn't support utf16

src/Cafe/HW/Latte/Renderer/Vulkan/VulkanRenderer.cpp
jambonmcyeah: done

CMakeLists.txt
jambonmcyeah: done

Exzap:
Looks good to me now. Thanks!

PR #130, opened 2022-08-31
Add ability to save/read files based on XDG specification

Tatsh:
- Fix some \\ in paths to use /

I accidentally deleted the branch in #126. Sorry :(

This is the latest version and all suggestions in the previous PR are handled here. Also there should be no unrelated whitespace changes in this PR. Formatting should be close to what Cemu is currently using.


Anuskuss:
> @Tatsh Putting the `bin` stuff into `SYSTEM_DATA_PATH` will also put `gameProfiles` in there, which will prevent a user launching cemu from creating game profiles (as `SYSTEM_DATA_PATH/gameProfiles` belongs to `root`). Can multiple `gameProfiles` be supported or should I just remove it from `SYSTEM_DATA_PATH` and put it in `$XDG_CONFIG_HOME/cemu`?

Okay I think I tested it throughly now and it definitely doesn't work. My observations:
- It won't write to `$XDG_CONFIG_HOME/cemu/gameProfiles` at all if the folder doesn't exists (if you can find the code and fix that that'd be great, otherwise I'll open an issue).
- It will write `$XDG_CONFIG_HOME/cemu/gameProfiles/$id.ini` with the correct values but will not read them on a restart and overwrite the ini with default values.
- It doesn't matter if `SYSTEM_DATA_PATH/{gameProfiles}` exists at all. It will however only read from `SYSTEM_DATA_PATH/gameProfiles/default` (even if `$XDG_CONFIG_HOME/cemu/gameProfiles` doesn't exists).

This is a weird corner case and it should be taken care of before merging. Thanks for your work by the way. Not a big fan of the whole XDG Base Directory spec and how I have to search my whole home directory to find things, but feels great to have atleast some kind of rules on Linux desktops.

Tatsh:
Ever since I accidentally closed my previous PR, I am very confused about the commit showing up here: https://github.com/cemu-project/Cemu/pull/130/commits/cd02fa64c0f34c7f084f9e46d0d1c80f490972e8

Tatsh:
Can you check if the gameProfile folder is getting created at some point?

The way it is designed (not by me but by Cemu originally with `GetMlcPath`) is that `GetMlcPath(std::string, std::string)` takes a format string and args for that, and then that calls the no-argument `GetMlcPath()` (in this PR one of the `Get*Path()` methods). In the no-argument `Get*Path()` methods, if the base directory does not exist, it gets created (including all parent directories).

```c++
	template <typename ...TArgs>
	[[nodiscard]] static fs::path GetCachePath(std::string_view format, TArgs&&... args)
	{
		// ... call no-arg GetCachePath()
		return GetCachePath() / fmt::format(format, std::forward<TArgs>(args)...);
	}

fs::path ActiveSettings::GetCachePath()
{
#ifdef XDG
	auto config_home = GetXDGPath("XDG_CACHE_HOME", fmt::format("{}/.cache", std::getenv("HOME")));
	auto dir = fs::path(fmt::format("{}/cemu", config_home));
        // Create the directory and all parents if it does not exist
	if (!fs::exists(dir))
		fs::create_directories(dir);
	return dir;
#endif
	return s_full_path;
}
```

Is this not working for `GetDataPath()`? It has been working for me but I have not messed with game profiles.

Tatsh:
Now I see the issue. I think there is a missing call to `fs::create_directories()` in the game profile code.

Kussanus:
@Tatsh Game profiles work now. There was some missing code and one mistake. See last patch in https://github.com/cemu-project/Cemu/issues/117#issuecomment-1233412235.

Tatsh:
I found the issue @Kussanus and just sent it in.

Anuskussss:
@Tatsh You missed the "create gameProfiles if it doesn't exists" patch. Look at my [PKGBUILD](https://github.com/cemu-project/Cemu/issues/117#issuecomment-1234808968) if you wanna know where it needs to go. You should also consider renaming 497e3df3e720a04e53852dbd0a24379ba22e754f :laughing:

Tatsh:
@Anuskuss I have no idea why https://github.com/cemu-project/Cemu/commit/497e3df3e720a04e53852dbd0a24379ba22e754f is showing up here.

PR #136, opened 2022-09-01
cmake: use more system libraries

abouvier:
Allow using `discord-rpc` and `vulkan-headers` from the system packages.


Crementif:
For the methods of distribution planned by us there is no reason to bypass git submodule dependencies. In general we rather have everyone be on the same exact config of dependencies to simplify troubleshooting and development.

abouvier:
That's why you can specify the exact version of the lib you need in `find_package`. So everybody will be obligated to end up with the same config.
I really don't understand why most (emulator?) developers tend to turn their project into some kind of package manager :p

And please don't revert them, but you already allowed two submodules (zarchive and cubeb) to be bypassed.

Exzap:
> I really don't understand why most (emulator?) developers tend to turn their project into some kind of package manager :p

I don't understand what you mean by "turning it into a package manager". Isn't that literally what your PR does by adding more ways of sourcing dependencies?
Anyway, we just prefer a solution that works the same on all platforms, including Windows.

abouvier:
`CMakeLists.txt` files should only contains a bunch of `find_package` calls. The task of fetching/compiling actual libraries should happens on the OS that build the project. In the script setting up the VM for your CI you should use vcpkg in command line to install the libraries listed in `vcpkg.json` for Windows, and use apt-get, pacman or whatever to install them for Linux. 98% of users that will download this repo are package maintainers who know how to install dependencies, they don't need CMake to do that. Other users will download the compiled binary directly.

PR #137, opened 2022-09-01
Linux Overlay Fixes

tomlally:
Added some missing stats to the overlay on Linux:
  - Process CPU use
  - Global CPU use by thread
  - RAM use

Bug Fixes:
  - Segfault when 'CPU per core usage' was enabled.


FearlessTobi:
It seems like something went wrong during your rebase.

PR #139, opened 2022-09-01
Fixes for building on macOS

OatmealDome:
Various fixes, including:

* `_umul128` is a built-in on clang, so we shouldn't attempt to redefine it.
* Remove usages of `u8string` in `wxTitleManagerList`.
* Always convert to `std::string` first when using `boost:icompare` in `wxLogCtrl` and `MemorySearcherTool` (not sure if this is 100% sane, but it does fix errors similar to those in the comments from [the prior macOS PR](https://github.com/cemu-project/Cemu/pull/52#issuecomment-1228516082)).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,6 +51,11 @@ if (MSVC)
   endif()
 endif()
 
+if (CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 14.0)
+  # Clang versions below 14 have issues with consteval in fmt: https://github.com/fmtlib/fmt/issues/2455
+  add_compile_definitions(FMT_CONSTEVAL=)
+endif()
+
 option(ENABLE_OPENGL "Enables the OpenGL backend" ON)
 option(ENABLE_VULKAN "Enables the Vulkan backend" ON)
 option(ENABLE_DISCORD_RPC "Enables the Discord Rich Presence feature" ON)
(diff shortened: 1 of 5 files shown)


src/gui/components/wxTitleManagerList.cpp
```diff
@@ -200,21 +200,21 @@ boost::optional<wxTitleManagerList::TitleEntry&> wxTitleManagerList::GetTitleEnt
 
 boost::optional<const wxTitleManagerList::TitleEntry&> wxTitleManagerList::GetTitleEntry(const fs::path& path) const
 {
-	const auto tmp = path.generic_u8string();
+	const auto tmp = path.generic_string();
```
JosJuice, line 203: This changes what encoding you get on Windows from UTF-8 to the system codepage, which I'm assuming is not intended. Would this work as an alternative?

```suggestion
	const auto tmp = _utf8Wrapper(path);
```

CMakeLists.txt
```diff
@@ -51,6 +51,15 @@ if (MSVC)
   endif()
 endif()
 
+if (CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 14.0)
+  # Clang versions below 14 have issues with consteval in fmt: https://github.com/fmtlib/fmt/issues/2455
+  add_definitions(-DFMT_CONSTEVAL=)
```
Tachi107, line 56: Please use `STREQUAL` instead of `MATCHES`, as the latter is for regular expressions. Also, prefer `add_compile_definitions()` to `add_definitions()`, as the latter is semi-deprecated.

PR #142, opened 2022-09-01
Linux Overlay (2)

tomlally:
Added some missing stats to the overlay on Linux:
  - Process CPU use
  - Global CPU use by thread
  - RAM use

Bug Fixes:
  - Segfault when 'CPU per core usage' is enabled.


src/util/SystemInfo/SystemInfoUnix.cpp
```diff
@@ -0,0 +1,56 @@
+#if BOOST_OS_LINUX
```
bitscher, line 1: Since this file is Linux-specific with its use of `/proc/self/statm`, I suggest renaming this file with a Linux suffix instead of Unix (which itself really should be renamed to Posix everywhere IMO).
You should also add a stub (empty) implementation of `GetProcessorCount()`, `QueryRamUsage()`, `QueryProcTime()` , `QueryCoreTimes()`  for the rest of the systems (think of MacOS and the BSDs)

CMakeList should also be modified to include the platform-dependent source files, by doing so you can remove the preprocessor guards in this file and `SystemInfoWin.cpp`

@Exzap I think we need a clear naming convention for platform specific files.
How does the following sound?
`ClassName_posix.cpp` : Implementation common across posix systems.
`ClassName_win32.cpp` : Implementation specific to Windows.
`ClassName_linux.cpp` : Implementation specific to Linux.
`ClassName_macos.cpp` : Implementation specific to MacOS X.

src/util/SystemInfo/SystemInfoUnix.cpp
```diff
@@ -0,0 +1,56 @@
...
+uint32 GetProcessorCount()
+{
+	return std::thread::hardware_concurrency();
+}
+
+uint64 QueryRamUsage()
+{
+	long page_size = sysconf(_SC_PAGESIZE);
+
+	std::ifstream file("/proc/self/statm");
```
bitscher, line 19: You should look into using `getrusage`, it's supported on Linux, OSX, FreeBSD and OpenBSD
https://man7.org/linux/man-pages/man2/getrusage.2.html
https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/getrusage.2.html
https://www.freebsd.org/cgi/man.cgi?query=getrusage&sektion=2
https://man.openbsd.org/getrusage.2

tomlally: I know about `getrusage` but I don't think you can get the current memory usage from that. Please let me know if it is possible.

src/util/SystemInfo/SystemInfoUnix.cpp
tomlally: > Since this file is Linux-specific with its use of `/proc/self/statm`, I suggest renaming this file with a Linux suffix instead of Unix

Agreed.

> You should also add a stub (empty) implementation of `GetProcessorCount()`, `QueryRamUsage()`, `QueryProcTime()` , `QueryCoreTimes()` for the rest of the systems (think of MacOS and the BSDs)

Agreed. In `FileName_stubs.cpp`?

> CMakeList should also be modified to include the platform-dependent source files, by doing so you can remove the preprocessor guards in this file and `SystemInfoWin.cpp`

The CMakeLists contains this: `file(GLOB_RECURSE CPP_FILES *.cpp)` which makes it difficult to select files on a per-platform basis. `util/MemMapper` uses the same style of pre-processor guards, however I agree it is less than ideal. I think somebody has a PR open with some major refactors to the CMakeLists so I don't want to cause problems for them.

bitscher: Have a look at https://github.com/cemu-project/Cemu/blob/main/src/Common/CMakeLists.txt

src/util/SystemInfo/SystemInfoUnix.cpp
bitscher: I think you just need to read `ru_maxrss`

```
ru_maxrss (since Linux 2.6.32)
              This is the maximum resident set size used (in kilobytes).  For RUSAGE_CHILDREN, this is the resident set size of the largest child, not the maximum resident set size of the process tree.
```

bitscher: Oh I see the issue, it's the maximum value on the lifetime of the process isn't it?

src/util/SystemInfo/SystemInfoUnix.cpp
tomlally: That doesn't use `GLOB_RECURSE`.

bitscher: Right. This kind of recursive rule is usually a bad idea lets hope this goes away soon enough.

bitscher: > > You should also add a stub (empty) implementation of `GetProcessorCount()`, `QueryRamUsage()`, `QueryProcTime()` , `QueryCoreTimes()` for the rest of the systems (think of MacOS and the BSDs)
> 
> Agreed. In `FileName_stubs.cpp`?

I think that would be ok.

Also the `GetProcessorCount()` implementation with `std::thread::hardware_concurrency()` seems portable, so you could make it common and remove the windows version.

dmanlfc:
patched this in after adding the files to the cmake file - works well, thanks!

tomlally:
> patched this in after adding the files to the cmake file - works well, thanks!

Thanks. Does windows still give the correct reading for per-core usage? I'm not sure if it's done correctly in this; that's why I didn't open this PR earlier.

dmanlfc:
> > patched this in after adding the files to the cmake file - works well, thanks!
> 
> Thanks. Does windows still give the correct reading for per-core usage? I'm not sure if it's done correctly in this; that's why I didn't open this PR earlier.

i don't windows...

PR #145, opened 2022-09-02
Add SIGINT handler on posix systems 

bitscher:
There's currently no overlap between the windows and posix
implementation of the exception handler so I split them into
different files for better clarity.

Added a signal handler for `SIGINT` to allow exiting on `CTRL+C`.
Note that like the rest of the program it does not exit cleanly and
triggers a `SIGABRT.`

Also removes function `CemuApp& wxGetAppWrapper()`
that was seemingly unused but declared with C linkage despite
returning a C++ reference, which was causing a warning.


Exzap:
LGTM. Thanks

Issue #140, opened 2022-09-01
Online Broken for NA Network IDs.


Merik2013:
The Wii U update from a few days ago appears to be blocking people using NA nnids from connecting to online game servers using Cemu. Redumping the online files from an updated system doesnt appear to fix this.

sneekee:
I am experiencing the same issue, here are the relevant entries from my log.txt file

```
[10:28:25] HTTP request received response 400 but expected 200
[10:28:25] Account response with error code 11
[10:28:25] Message from server: System update is required
[10:28:25] ACT_GetNexToken(): Failed to retrieve OAuth token
```

Masamune3210:
You likely need to update your WiiU and then redump the certs, there might have been a change in them since the update was focused around disallowing accounts to add funds

Merik2013:
Exzap released a hotfix in the discord. Check the pins in #cemu_discussion.

Exzap:
I pushed a fix for this, it will be in the next experimental release of 2.0 which should be ready soon. Until then you can use my patched version of 1.27.1 that I posted on discord and [reddit](https://www.reddit.com/r/cemu/comments/x3j4dh/workaround_for_online_issues_with_north_america/).

I'll leave this open for visibility.

PR #147, opened 2022-09-02
Fix online for NA console logins

Exzap:


bitscher:
Is this linked to the new firmware update they recently released in the US?

Crementif:
Yeah, they bumped the system version that's required to connect to online services for US consoles. EU and Japan seems to be fine though.

bitscher:
Resolves #140

PR #148, opened 2022-09-02
gfxpack: Fix graphic pack content not being loaded

Crementif:
This prevented custom shaders and file mods from loading for graphic packs.

It's seemingly indicating a more complicated problem with the graphic packs being "activated" after the game has started using custom shaders.


Crementif:
Superseded by #168

Issue #123, opened 2022-08-30
Mouse cursor doesn't hide in fullscreen


kturques:
Cursor doesn't go away in full-screen mode.

Operating System: openSUSE Tumbleweed 20220827
KDE Plasma Version: 5.25.4
KDE Frameworks Version: 5.97.0
Qt Version: 5.15.5
Graphics Platform: Wayland

![Screenshot_20220830_102202](https://user-images.githubusercontent.com/72952967/187449530-043a8eb0-2739-41c6-af94-2541f48105d5.png)

PR #149, opened 2022-09-02
Linux cursor hide.

tomlally:
https://github.com/cemu-project/Cemu/issues/123


src/gui/MainWindow.cpp
```diff
@@ -1176,7 +1176,14 @@ void MainWindow::ShowCursor(bool state)
 		} while (counter >= 0);
 	}
 	#else
-	cemuLog_log(LogType::Force, "MainWindow::ShowCursor - todo");
+	if (state)
```
bitscher, line 1179: This actually looks like a portable solution.
Maybe we don't need this windows version that relies on the windows API.

PR #150, opened 2022-09-02
cmake: Exclude unnecessary dep build targets

Ziemas:
Stops building dependency build targets that are not needed. This stops cubeb running doxygen on every build.


Exzap:
thanks!

PR #151, opened 2022-09-02
Remove boost::filesystem

Zopolis4:
`boost::filesystem::windows_name` hasn't been defined since very old versions of boost.


Tachi107:
`boost::filesystem` is used in different places in the codebase, often behind the `fs::` namespace

bitscher:
I still see a definition in the git code base though.

https://github.com/boostorg/filesystem/blob/develop/src/portability.cpp#L56

That said, I think we should be using something like
[`portable_file_name()`](https://github.com/boostorg/filesystem/blob/develop/src/portability.cpp#L74) instead on `windows_name` but it's something that can be implemented using the standard library I think.

bitscher:
That's what's causing the compilation error:
```
void ActiveSettings::LoadOnce()
{
	s_full_path = boost::dll::program_location().generic_wstring();
```
`boost::dll` seems to rely on `boost::fileystem`, if you find a way to get rid of it too you should be good :)

Exzap:
Closing this since it seems to be stale. If you ever make any progress on this feel free to open a new PR. Removing the dependency on boost filesystem would actually be pretty nice since it comes with a lot of baggage that we don't need.

Issue #144, opened 2022-09-02
Linux - Illegal instruction error


leifliddy:
I've compiled ```Cemu``` on Fedora 36 (and 37) according to the (gcc) instructions in the ```BUILD.md``` guide. 
When I execute the binary I receive the following error:
```
[leif.liddy@black cemu]$ ./Cemu 
Illegal instruction (core dumped)
```

I'm still rockin a ```i7-3770K``` which is pretty dated at this point. Going to upgrade when the Zen4 chips are released at the end of the month --but I figured this could help someone out who might be running into the same issue. 

I believe this error is related age of this chip and the instruction sets it supports as the  same ```Cemu``` binary runs fine on my laptop with a more modern processor. I  figured there must be a flag I can pass to cmake that enables all instruction subsets supported by my CPU. But I'm not sure how to do that....

Here are pertinent details:

**GDB**
```
(gdb) run
Starting program: /home/leif.liddy/programs/cemu/Cemu 
....
[Thread debugging using libthread_db enabled]                                                                                                                                                                                                
Using host libthread_db library "/lib64/libthread_db.so.1".

Program received signal SIGILL, Illegal instruction.
0x00000000004e3451 in _GLOBAL__sub_I_rangeStore_ppcRanges ()

(gdb) bt
#0  0x00000000004e3451 in _GLOBAL__sub_I_rangeStore_ppcRanges ()
#1  0x00007ffff5df667e in call_init (env=<optimized out>, argv=0x7fffffffde68, argc=1) at ../csu/libc-start.c:145
#2  __libc_start_main_impl (main=0x4de400 <main>, argc=1, argv=0x7fffffffde68, init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fffffffde58) at ../csu/libc-start.c:368
#3  0x0000000000546a25 in _start ()

(gdb) display/i $pc
1: x/i $pc
=> 0x4e3451 <_GLOBAL__sub_I_rangeStore_ppcRanges+689>:	vinserti128 $0x1,%xmm1,%ymm0,%ymm0
```

**coredump**
```
       Storage: /var/lib/systemd/coredump/core.Cemu.1000.9d2fb048c3fe43919b2d12c9f3665568.510047.1662087909000000.zst (present)
     Disk Size: 1012.6K
       Message: Process 510047 (Cemu) of user 1000 dumped core.
                
[...]

PR #152, opened 2022-09-02
Per-function target attribute on clang and GCC.

tomlally:
Resolves https://github.com/cemu-project/Cemu/issues/144


Exzap:
For `LatteIndices_fastConvertU16_SSE41` can you make a separate ATTRIBUTE_SSE42 macro? With the target being `"sse4.2"`. A sse4.1 target would make more sense but it seems like that doesn't exist.

tomlally:
"sse4.1" works okay for me.

bitscher:
Speedy resolution! :+1:

PR #153, opened 2022-09-02
style: use std::enable_if & co instead of boost::

Tachi107:
Most of the codebase already does this

(Was part of #112)


bitscher:
I think `#include <type_traits>` be removed from the headers too.

PR #156, opened 2022-09-03
Portable way of opening URLs and directories

bitscher:
Replaced every instance of ShellExecute with wxLaunchDefaultBrowser

### DIFF
--- a/src/gui/components/wxGameList.cpp
+++ b/src/gui/components/wxGameList.cpp
@@ -4,9 +4,6 @@
 #include "util/helpers/helpers.h"
 #include "gui/GameProfileWindow.h"
 
-#ifdef _WIN32
-#include <shellapi.h>
-#endif
 #include <numeric>
 
 #include <wx/wupdlock.h>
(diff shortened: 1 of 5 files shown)


src/gui/components/wxGameList.cpp
```diff
@@ -481,7 +478,7 @@ void wxGameList::OnContextMenu(wxContextMenuEvent& event)
 			menu.AppendSeparator();
 			menu.Append(kWikiPage, _("&Wiki page"));
 			menu.Append(kContextMenuGameFolder, _("&Game directory"));
-			menu.Append(kContextMenuSaveFolder, _("&Save directory"))->Enable(true);
+			menu.Append(kContextMenuSaveFolder, _("&Save directory"))->Enable(wxDirExists(gameInfo.GetSaveFolder().string()));
```
Exzap, line 481: Never call .string() on `std::filesystem::path`, it comes with all sorts of implementation-defined quirks. We have a small helper function for this, `_utf8Wrapper` which turns a path into a `std::string` with utf8 encoding. Which then can be turned into a wxString via helper `to_wxString`.

But in this case you can also just do `fs::exists(gameInfo.GetSaveFolder())` (fs is our alias for std::filesystem). But even this comes with a caveat. Under special (platform-specific) circumstances it can throw an exception which has to be handled. This can be avoided by providing the optional error code parameter and then instead of throwing it will return false. The robust solution would be:
```
std::error_code ec;
menu.Append(kContextMenuSaveFolder, _("&Save directory"))->Enable(fs::exists(gameInfo.GetSaveFolder(), ec));
```

bitscher: Right, I totally forgot that windows relies heavily on wchar instead of UTF-8.

PR #157, opened 2022-09-03
ci: colors in CMake

Tachi107:
This way logs will be a bit more readable, but only when using CMake 3.24 or newer.

I've also removed trailing whitespace and changed the `mkdir build && cd build && cmake ..` ritual in the more concise (and less error prone) `cmake -B build`


Exzap:
Closing this since it became stale, which is partly my fault for sitting on this so long. Sorry :(
I'll manually integrate these changes next time I do some work on the CI, if I don't forget. Thanks for the suggestion to use color output

Tachi107:
Don't worry, I didn't even notice that this got out of sync with master. I can rebase and reopen, if you wish

Exzap:
Appreciated but you don't need to bother. The CI has changed quite a bit so you'd have to redo it and not just rebase.

PR #161, opened 2022-09-03
Add curl as linux build dependency

uranuspucksaxophone:
This fixes an issue I had about bootstraping vcpkg because curl wasn't found.


Crementif:
Seems good, thanks!

PR #162, opened 2022-09-03
Include build-essential as ubuntu/debian dependency

ghost:
This fixed another issue with the build because it couldn't find a compiler.


Tachi107:
`build-essential` is not what you should install to get a C++ compiler on Debian; as its package description says, _"If you do not plan to build Debian packages, you don't need this package."_

The correct package to install would be either `clang` or `g++`

PR #163, opened 2022-09-03
Fix game list ordering & Favorite checkbox

bitscher:
Game ordering was broken when clicking on the name column or adding/removing a favorite.
    
Unlike the standard lib's that requires a strict weak ordering comparator, wxWidgets needs a compare function that returns a negative number, 0 or a positive number when entries are respectively lower than, equal to and greater than.
    
Fixed using C++20 comparison API.

### DIFF
--- a/src/gui/components/wxGameList.cpp
+++ b/src/gui/components/wxGameList.cpp
@@ -278,23 +278,27 @@ void wxGameList::UpdateItemColors(sint32 startIndex)
 	}
 }
 
+static inline int strongorder_to_int(const std::strong_ordering &wo)
+{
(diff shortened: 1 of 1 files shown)


src/gui/components/wxGameList.cpp
```diff
@@ -474,6 +474,7 @@ void wxGameList::OnContextMenu(wxContextMenuEvent& event)
 			menu.Append(kContextMenuStart, _("&Start"));
 
 			bool isFavorite = GetConfig().IsGameListFavorite(title_id);
+      std::error_code ec;
```
bitscher, line 477: Uh oh, it looks like github merge resolution tool did something weird.

PR #170, opened 2022-09-04
Fix various compiler warnings

bitscher:


src/Cafe/HW/Latte/Core/LatteCommandProcessor.cpp
```diff
@@ -616,7 +616,7 @@ LatteCMDPtr LatteCP_itDrawIndex2(LatteCMDPtr cmd, uint32 nWords, DrawPassContext
 	uint32 count = LatteReadCMD();
 	uint32 ukn3 = LatteReadCMD();
 
-	performanceMonitor.cycle[performanceMonitor.cycleIndex].drawCallCounter++;
+	performanceMonitor.cycle[performanceMonitor.cycleIndex].drawCallCounter = performanceMonitor.cycle[performanceMonitor.cycleIndex].drawCallCounter + 1;
```
Tachi107, line 619: This doesn't look like "fixing" the warning, it is simply hiding it :/

Can't `volatile` be replaced with something else?

bitscher: I can't say I disagree with you.
It probably should be using of one C++11 atomic classes instead but I am not familiar enough with the Latte code to know if this will have a side effect or not.

bitscher: I'm going until @Exzap chimes in on this one about removing the volatile before doing anything else, otherwise I will simply drop this commit.

Exzap: You can just drop the `volatile` here. Seems like modern MSVC ignores it anyway. Strict atomic guarantees are not necessary and I'm unsure of the performance impact that `std::atomic` would have since all the increments are on the hot path.

On a related note, performanceMonitor as a whole is pretty dated code and could be implemented a lot cleaner. It should get a complete overhaul at some point but for now it's not super important since it's only used for the overlay statistics.

bitscher: Alright, I'm going to remove the volatile in the performance monitor.

What about `LatteGPUState.osScreen.screen[screenIndex].flipRequestCount` ?

Exzap: That one should be std::atomic

bitscher: That's out of the scope of this PR, I will take the time to analyze the implications of changing `flipRequestCount` (and `flipExecuteCount` to `std::atomic` in another change.

PR #174, opened 2022-09-04
fix H264_IsBotW() return false while BotW's TitleRegion is JP

lijunyu-cn:
This cause fail to playback the 1080p videos from the BotW Switch version if BotW's TitleRegion is JP.


Exzap:
thanks!

PR #175, opened 2022-09-04
Use latest OS versions for CI/CD

ghost:


qurious-pixel:
Ubuntu-latest is 20.04, but why change the image? This raises the glibc requirement when it moves to Ubuntu 22.04, making the binary unusable on any system that does not meet this new minimum.

PR #177, opened 2022-09-04
Update fmt version to 9.1.0

SSimco:


Crementif:
Nice! Would be interesting to see if clang 14 works after this change since that had a string formatting template issue on the old fmt version.

Does it compile on fmt 9.0.0 too? Not sure if it's preferable, but it could allow us to remove the need to overlay the fmt package.

SSimco:
I've tried compiling it on Arch with clang 14 and it works.
It doesn't compile with fmt 9.0.0, maybe the fmt package that's on vcpkg will be updated in the near future and then we can use the one on vcpkg.

bitscher:
Hopefully soon enough `std::fmt` will be supported by all major compilers/libc++ and we can get rid of the external lib altogether.

Exzap:
Looks good to me. Thanks!

PR #181, opened 2022-09-05
clang-format: Add linter and reformat code

FearlessTobi:
Since the author of https://github.com/cemu-project/Cemu/pull/56 seems to be unresponsive, I've made a new PR, using the sugestions from the comments.

It is now using an updated .clang-format file, and clang version 12 instead of 14 for better compatiblity.

I've also let clang-format run on all of the files in the project using `powershell.exe -Command "Get-ChildItem '${SRCS}/*' -Include *.cpp,*.h -Recurse | Foreach {&'${CLANG_FORMAT}' --style=file -i $_.fullname}`

Fixes https://github.com/cemu-project/Cemu/issues/32.


Zopolis4:
Why do we need to use clang-format 12 in the actions? 14 has more bugfixes and accurate behaviour. I see no reason to intentionally use an out of date version when we can use a better one.

FearlessTobi:
Many Linux distros (e.g. Ubuntu LTS) only ship with clang version 12 right now and updating is kind of a hassle.
Right now, for building on Ubuntu 20.04, the build guide recommends version 12 as well. 
If we used clang-format 14. someone on this OS would have to figure out how to install a newer compiler version, only to be able to format their code. 
Also right now, clang 12 is the version used for building in CI as well (https://github.com/cemu-project/Cemu/blob/main/.github/workflows/build.yml#L60).

So all in all, IMO the small improvements of clang-format 14 are not worth the hassle it would cause for some users.

Exzap:
As mentioned on Discord a while ago, we'll handle this internally.
Current plan of action is to write coding style guidelines and then we'll look into a way to introduce `.clang-format` without causing too many conflicts for contributors.

I'll leave this open for visibility.

Exzap:
Superseded by #968

PR #184, opened 2022-09-05
Use ANSI encoding for cemu.rc (remove utf8 BOM)

Exzap:
Clang has issues with utf16 and MSVC does not like utf8. So ANSI it is


Tachi107:
> MSVC does not like utf8

Can't the [`/utf-8`](https://docs.microsoft.com/en-us/cpp/build/reference/utf-8-set-source-and-executable-character-sets-to-utf-8) flag help?

Exzap:
I rather not clutter the CMakeLists with extra flags if it can easily be avoided. We don't need more than ansi/ascii for the resource file anyway

PR #185, opened 2022-09-05
Use utf8 aware string conversion for wxLaunchDefaultBrowser()

Exzap:


bitscher:
Oh, sorry it should have been part of my previous PR.

Exzap:
No problem, I didn't catch it either when merging. Only noticed while testing

PR #188, opened 2022-09-06
Improved clarity.

tomlally:
`_MSC_VER < 1923` evaluates to `true` on every compiler that does not define `_MSC_VER`, so this expression `#if _MSC_VER < 1923 || defined(__clang__)` almost always evaluates to true and is potentially misleading. Replaced with a (hopefully) clearer condition.

### DIFF
--- a/src/Cafe/HW/Espresso/PPCTimer.cpp
+++ b/src/Cafe/HW/Espresso/PPCTimer.cpp
@@ -111,10 +111,10 @@ uint64 PPCTimer_tscToMicroseconds(uint64 us)
 
 	uint64 remainder;
 
-#if _MSC_VER < 1923 || defined(__clang__)
-	const uint64 microseconds = udiv128(r.low, r.high, _rdtscFrequency, &remainder);
-#else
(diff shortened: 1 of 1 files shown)


bitscher:
Good catch, this check made sense when MSVC was the only target compiler but not anymore.
I wouldn't be surprised if the codebase had more instance of this kind of things.
If you have a bit of time to spare I think it'd be a good idea to go through every conditional processor expression that may use comparisons like this that may not be properly defined on other compilers.

Exzap:
I don't think Cemu even compiles with MSVC 1923 or lower anymore due to us now relying on pretty recent C++ features like concepts. I believe we could clean this up further and get rid of the assembly implementation by always using the native `_udiv128` intrinsic on MSVC and on clang/gcc we reimplement it by using `unsigned __int128`.

tomlally:
> I believe we could clean this up further and get rid of the assembly implementation by always using the native `_udiv128` intrinsic on MSVC and on clang/gcc we reimplement it by using `unsigned __int128`.

I agree, I'll make a PR in a moment.

PR #190, opened 2022-09-06
Separate filestream.h into a implementation .cpp file.

goeiecool9999:
Can someone check for build errors on windows?
Should the comments be only in the header? I did so for the windows version but not for the linux version.
Best place for the comments is in the header files I suppose. Would be bad to have duplicates in the .cpp file.


goeiecool9999:
Oops. IDE was set to spaces and original uses tabs.

Exzap:
It's probably worth going one step further and splitting the Windows and Linux/MacOS implementations into two different files. I suggest naming them `FileStream_win32.cpp` and `FileStream_unix.cpp`. I also think we should capitalize the file name to match the class name, same for the header.

Regarding comments, having them in the header makes the most sense. For implementation details those can be in the .cpp obviously. If you are unsure you can remove comments since I believe most of the interface is pretty self-explanatory anyway.

goeiecool9999:
All done. When I refactored the windows functions I forgot to remove the default arguments from the implementation. Should be building now unless I messed up the CMake file

goeiecool9999:
That should do it

PR #192, opened 2022-09-07
Fix various issues related to title names

lijunyu-cn:
Fixed various issues related to title names：
1. If console language is set to non-English, the title names in Title Manager may be empty.
2. Game list do not show custom names if list style set to icon or small icon.
3. Custom title names for some languages (like JP or CN) appear confusing if re-edited.
4. Unexpected overrides of variable values in ParsedMetaXml.h


Exzap:
Looks good to me. Thanks!

PR #193, opened 2022-09-07
Replace MSVC_RUNTIME_LIBRARY generator expression with a compile option

Zopolis4:
See https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html


Squall-Leonhart:
wxbase32ud.lib(init.cpp.obj) : error LNK2019: unresolved external symbol _CrtSetDbgFlag referenced in function "public: __cdecl EnableMemLeakChecking::EnableMemLeakChecking(void)" (??0EnableMemLeakChecking@@QEAA@XZ) [D:\a\Cemu\Cemu\build\src\CemuBin.vcxproj]

Zopolis4:
The CI fix should work, but I can't test it apparently.

Squall-Leonhart:
turn on the workflow and trigger either release or experimental build.

Zopolis4:
...what? I don't have the permissions to trigger a build on-command, and apparently this repo is set up in such a way where my CI modifications dont affect the CI on the PR they're in.

Squall-Leonhart:
you can trigger actions on your own repo in the forked workflow.

for some reason yours is missing Release though.

Tachi107:
Why? Replacing a target property with a manual build flag seems like a downgrade to me :/

Zopolis4:
It's the same? The target property is a generator expression, which is only called when the configuration is set to debug. This is the same, but in a cleaner and more readable way.

Tachi107:
> It's the same? The target property is a generator expression, which is only called when the configuration is set to debug. This is the same, but in a cleaner and more readable way.

No, it's not the same.

- If you can use a target property, use it. Global state is not great, and manually fiddling with compiler flags isn't great either
- Using generator expressions is the only way to properly set this property, as generator expressions are evaluated at build time as opposed to configure (`cmake -B build`) time. Writing `if (CMAKE_BUILD_TYPE)` has purpose only in single-config build tools like make and Ninja; Visual Studio doesn't define the build type at configure time, but can dynamically set it during build time (by selecting either a Debug or Release build in the VS GUI). See [`cmake-buildsystem(7)`](https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#build-configurations) for details

> For single configuration generators like [Makefile Generators](https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html#makefile-generators) and [Ninja](https://cmake.org/cmake/help/latest/generator/Ninja.html#generator:Ninja), the configuration is specified at configure time by the [CMAKE_BUILD_TYPE](https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html#variable:CMAKE_BUILD_TYPE) variable. For multi-configuration generators like [Visual Studio](https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html#visual-studio-generators), [Xcode](https://cmake.org/cmake/help/latest/generator/Xcode.html#generator:Xcode), and [Ninja Multi-Config](https://cmake.org/cmake/help/latest/generator/Ninja%20Multi-Config.html#generator:Ninja%20Multi-Config), the configuration is chosen by the user at build time and **[CMAKE_BUILD_TYPE](https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html#variable:CMAKE_BUILD_TYPE) is ignored**.

Exzap:
The current solution is fine. It's closer to what modern cmake encourages.

A point can be made that the runtime has to match for all targets, hence why it should be set globally instead of per target. But either way it works just fine so why change it.

Tachi107:
> A point can be made that the runtime has to match for all targets, hence why it should be set globally instead of per target. But either way it works just fine so why change it.

In theory what you say makes perfect sense, in practice CMake's crazy scoping rules make these "globals" unreliable, as any subproject could override the `CMAKE_` variable at any time, breaking everything in amazing, or even worse, subtle ways. As far as I know, target properties are the only somewhat reliable thing in CMake.

PR #194, opened 2022-09-07
Set timer milliseconds for "Controller settings"

drizzt:
This is needed to avoid having the Controller settings window black on Linux


Exzap:
thanks!

PR #196, opened 2022-09-07
Find case-insensitive file/folder path if not found

SSimco:


src/Common/unix/FileStream_unix.cpp
```diff
@@ -1,5 +1,31 @@
...
+			parrentPath = std::move(tempParrentPath.value());
+		else
+			return {};
+	}
+	std::string fName = path.filename().string();
+	if (fs::exists(parrentPath))
+		for (auto&& dirEntry : fs::directory_iterator(parrentPath))
+		{
+			std::string dirFName = dirEntry.path().filename().string();
+			if (boost::iequals(dirFName, fName))
```
Tachi107, line 21: Does this work with non-ascii strings?

SSimco: If the filename contains non-ASCII characters, e.g. "fileỰ.txt", and the case matches for those characters, e.g. "FiLeỰ.TxT", then it works. But it will not work if it tries to look for "fileự.txt".
I could make it work for non-ASCII file using Boost.Locale if it's necessary.

goeiecool9999:
I am working on a patch too. Will put up a PR soon.

Exzap:
The Wii U part of the path will always be ANSI. So `boost::iequals` should suffice. But ideally we want something that is utf8 aware but still only compares characters within the ASCII character set case-insensitive (the Wii U OS only case-insensitively compares a-z).

goeiecool9999:
Heads up! I have another PR for this PR that fixes the bug I created with the first PR.

goeiecool9999:
Another PR is up https://github.com/SSimco/Cemu/pull/3

PR #199, opened 2022-09-07
Ignore case for game files when on a unix system.

goeiecool9999:
Implemented as described in https://github.com/cemu-project/Cemu/issues/107#issuecomment-1231171415

### DIFF
--- a/src/Cafe/Account/Account.cpp
+++ b/src/Cafe/Account/Account.cpp
@@ -175,7 +175,7 @@ std::error_code Account::Load()
 
 std::error_code Account::Save()
 {
-	FSPath path = CemuApp::GetMLCPath(fmt::format(L"usr/save/system/act/{:08x}", m_persistent_id)).ToStdWstring();
+	fs::path path = CemuApp::GetMLCPath(fmt::format(L"usr/save/system/act/{:08x}", m_persistent_id)).ToStdWstring();
 	if (!fs::exists(path))
 	{
 		std::error_code ec;
--- a/src/Cafe/Account/Account.cpp
+++ b/src/Cafe/Account/Account.cpp
@@ -175,7 +175,7 @@ std::error_code Account::Load()
 
 std::error_code Account::Save()
 {
-	fs::path path = CemuApp::GetMLCPath(fmt::format(L"usr/save/system/act/{:08x}", m_persistent_id)).ToStdWstring();
+	FSPath path = CemuApp::GetMLCPath(fmt::format(L"usr/save/system/act/{:08x}", m_persistent_id)).ToStdWstring();
 	if (!fs::exists(path))
 	{
 		std::error_code ec;
--- a/src/Cafe/Filesystem/FSPath.cpp
+++ b/src/Cafe/Filesystem/FSPath.cpp
@@ -1,5 +1,7 @@
 #include "FSPath.h"
 
+#include <boost/locale/conversion.hpp>
+
 #ifdef BOOST_OS_UNIX
 FSPath operator/ (const FSPath& lhs, const FSPath& rhs)
 {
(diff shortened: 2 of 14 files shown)


goeiecool9999:
Is the use of the lambda function okay? I find it a bit messy and it could be useful in other places. Also not 100% if wstring was the right choice for iterating over unicode characters originating from C++ path objects.

goeiecool9999:
I foresee a bug though. This solution goes from the root of the path. That means the process needs directory list permissions on every step up to the target directory. If it doesn't have that I don't know what happens.

goeiecool9999:
Yeah. If the user doesn't have read permissions on every directory leading up to the game files cemu crashes using this method.

goeiecool9999:
I swear I thought I added these changes to the previous commit...

goeiecool9999:
I'm thinking about two possible solutions to fix the permission problems.
1. Before trying to find the right case, navigate up the tree and check how far up you can go before you arrive at the folder you don't have permission to list.
2. Move the logic to ignore case to fsc_lookupPath. However you'd need the ability to see if the device is the host filesystem. Could be done by checking the inheritance of the device pointer. However fscDeviceHostFSC lives entirely inside fscDeviceHostFS.cpp so it would require a refactor to put it in the header so that fsc.cpp can be aware of the type.

goeiecool9999:
> is it just a duplicate of #196?

No, the other PR implements it in a different file that would make all file access by Cemu case-insensitive and not just the code looking for Wii U files and it also doesn't search for case differences recursively.

Exzap:
It's a tricky problem. While this PR covers all the file access done by games themselves, it doesn't cover the many instances where Cemu accesses MLC files in it's HLE logic (usually where GetMlcPath() is called, but sometimes it manually constructs the path).

I'm ok with a quick&dirty solution in the style of PR #196 where it just tries to case-correct each path. But ideally we should introduce a more fine grained solution. Something like adding `FileStream::OpenFile(fs::path basePath, fs::path relativeCIPath)` (same for Create) which then has explicit knowledge about which part of the path needs to be case-corrected. Of course this requires lots of code to be rewritten to use this new interface, but it shouldn't be too big of an effort.

goeiecool9999:
A little clumsy on the commit there once again. I only looked for manual path appending when it came it files that cemu doesn't attempt to create. Though I guess that's wrong when the file already exists and had it's case changed for some reason.

goeiecool9999:
If anyone has a better name for the FSPath class I'd love to hear it 😅

goeiecool9999:
I tried to find every place where Wii U paths are used. Please let me know if I missed one.

goeiecool9999:
> I've got a question, Windows supports case insensitive network drives, so shouldn't this be more filesystem or protocol specific (Windows supports 9P for example). Sorry for the misunderstanding.

This PR has nothing to do with windows. On unix systems paths are case sensitive and on windows they are not. If I understand correctly the Wii U filesystem handles case the same as windows and Cemu relied on Windows' case insensitivity. That's why bugs were introduced when Cemu was ported to linux. This PR modifies code where Wii U paths are being handled to explicitly find all files that exist on the filesystem with the same ignoring case.

Exzap:
I have some reservations in regards to your helper class:
- Operator overloading may be convenient, but it hides intent. Especially if the behavior is non-intuitive (nobody expects their paths to be case-altered).
- Since the case is altered permanently, you are losing information, but only on affected platforms. It's hard to quantify the side-effects of this.
- Inheriting from `fs::path` instead of making it a member means you leak lots of API and overloaded operators from the underlying implementation. I think that's your intent here but in large projects it's better to have "fool proof" classes where you can't accidentally bypass a feature by e.g. calling a different member function that is not wrapped. Solution: Inherit privately or better, make `fs::path` a class member. The latter avoids having surprise issues when you inherit a massive amount of implementation details.

But there is also a bigger design issue. The problem of path case mismatch is localized to file open and create operations in FileStream. Thus the solution should ideally be constrained to the same. Your PR moves a lot of the burden to the origin of where path building first happens (even altering GetMlcPath). I understand why you are doing this and it makes sense (and it's efficient). But it introduces a modification and constraint to the whole code base when there exists a viable alternative where we could transparently translate paths behind the scenes like PR #196 does. Basically a more modularization approach would be preferred.

goeiecool9999:
> But there is also a bigger design issue. The problem of path case mismatch is localized to file open and create operations in FileStream. Thus the solution should ideally be constrained to the same.

There are a couple instances where paths are built where cemu uses an (i/0)fstream to access the file instead of the standard FileStream class. Would those need to be changes as well then?

Does #196 run into any issues if multiple levels are case-mismatched?

Exzap:
> There are a couple instances where paths are built where cemu uses an (i/o)fstream to access the file instead of the standard FileStream class. Would those need to be changed as well then?

Yeah, it's old code that needs to be refactored.

goeiecool9999:
I'll close it for now. Seems like like #196 is the better candidate.

PR #202, opened 2022-09-07
fix clang-cl build

jambonmcyeah:
#188 broke it so here's the fix


tomlally:
Clang-cl does not support `_udiv128`?

jambonmcyeah:
> Clang-cl does not support `_udiv128`?

apparently not

Issue #178, opened 2022-09-04
[Linux] SIGABRT triggers when closing Cemu


superbonaci:
Compiled with latest commit [1373c76](https://github.com/cemu-project/Cemu/commit/1373c76eb9eb8251f6f12f69b4b26c8bd3f06518) on Manjaro Linux, then run it and close it:
```
./bin/Cemu

(Cemu:69679): Gtk-CRITICAL **: 10:47:45.694: gtk_box_gadget_distribute: assertion 'size >= 0' failed in GtkScrollbar
terminate called without an active exception
Aborted (core dumped)
```
The backtrace:
```
$ gdb ./bin/Cemu
GNU gdb (GDB) 12.1
Copyright (C) 2022 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-pc-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from ./bin/Cemu...
(gdb) run
Starting program: /home/user/Cemu/bin/Cemu 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".
[New Thread 0x7ffff2d3e6c0 (LWP 69924)]
[New Thread 0x7ffff21ff6c0 (LWP 69925)]
[New Thread 0x7ffff17b36c0 (LWP 69926)]
[New Thread 0x7ffff0fb26c0 (LWP 69927)]
[New Thread 0x7ffedbfff6c0 (LWP 69928)]
[New Thread 0x7ffedb7fe6c0 (LWP 69929)]
[New Thread 0x7ffedaffd6c0 (LWP 69930)]
[New Thread 0x7ffeda7fc6c0 (LWP 69931)]
[New Thread 0x7ffed9ffb6c0 (LWP 69932)]
[New Thread 0x7ffed97fa6c0 (LWP 69933)]
[New Thread 0x7ffed8ff96c0 (LWP 69934)]
[...]

PR #204, opened 2022-09-07
Use _Exit instead of exit under Linux

drizzt:
The best ExitProcess alternative for Linux is _exit since it doesn't call exit handlers.

This fixes also #178 since, with exit, it aborts due to some std::threads not correctly joined.

In the future, if VS2022 supports it, we may use std::jthread instead of std::thread since they automatically joins; when jthread will be used, `CemuApp::OnExit` may be removed since ExitProcess/_exit will not be necessary anymore

### DIFF
--- a/src/Common/ExceptionHandler/ExceptionHandler_posix.cpp
+++ b/src/Common/ExceptionHandler/ExceptionHandler_posix.cpp
@@ -26,7 +26,7 @@ void handler_SIGINT(int sig)
      * by any mean ends up with a SIGABRT from the standard library destroying
      * threads.
      */
-    exit(0);
+    _Exit(0);
 }
 
 void ExceptionHandler_init()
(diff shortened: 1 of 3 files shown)


Tachi107:
Wouldn't it be better to call `_Exit()`? It has the same functionality, but is part of the C (and C++) standard.

drizzt:
> Wouldn't it be better to call `_Exit()`? It has the same functionality, but is part of the C (and C++) standard.

I don’t have an opinion about that. _exit is POSIX and since on Windows ifdef a Windows function was used I just used the POSIX function for non-Windows case, but _Exit works the same for me

Tachi107:
_Exit is POSIX too; from `man 3p _Exit`:

 > The _Exit() and _exit() functions shall be functionally equivalent.

-- 

bitscher:
Could you also replace this call please? https://github.com/cemu-project/Cemu/blob/6cdb6eed1730cde23ede99099a12092b9abe8aa3/src/Common/ExceptionHandler/ExceptionHandler_posix.cpp#L29

drizzt:
I should have fixed all the problems reported in comments

Exzap:
In the long term we should work towards properly shutting down all the systems, it also goes hand in hand with the ability to restart emulation. But for now this suffices as a workaround.

PR #205, opened 2022-09-07
Posix/Linux: Add setting to disable coredumps

goeiecool9999:


tomlally:
This feels like it should be an opt-in option when compiling.

tomlally:
This feels like it should be an opt-in option when compiling.

goeiecool9999:
> This feels like it should be an opt-in option.

I agree. But since the linux version crashes every time you close it, in this case it may make more sense to be opt-out?

goeiecool9999:
#204 Would make this PR a lot less useful. I still think dumping on every crash is something the user should explicitly enable. All crashes other than segfaults would still dump core.

Tatsh:
I have not had this issue really despite running the app a lot. systemd rotates the coredumps. Probably can be disabled there too.

Exzap:
We have a crash dump option under general settings -> debug which is currently used only on Windows. Maybe it could be tied to that?

goeiecool9999:
It respects the user setting now but Lite has no meaning on Linux. I tried to hide the Lite setting, but then Full looks out of place so I changed it to "Enabled" but surprisingly there's no existing translation for it. I found editing the translation files not straightforward. Not sure how to neatly change a setting from 3 options to 2. If we leave the Config enum untouched Lite would become Enabled. It also raises questions in regard to settings compatibility. Like when moving an older config to a newer version where dump is was to Full. Code check would probably still work if we change it to `> Disabled`, but in the settings menu it would still bug out.

Exzap:
I don't think it makes sense to try and match the Windows options. I recommend just making it all OS-specific. Maybe rename the `CrashDump` enum to `CrashDumpWindows` and then add `CrashDumpLinux`. It should be stored separately for settings.xml anyway so when people transfer their settings between Windows/Linux it doesn't contaminate the dump option.
For the UI you can make it OS-specific too. A bit of `#if BOOST_OS_LINUX` and similar is ok.

goeiecool9999:
I think the only thing that's left is translation of the new strings.

Exzap:
Looks good. Thanks!

PR #206, opened 2022-09-07
Cleanup udiv128

tomlally:
As discussed in https://github.com/cemu-project/Cemu/pull/188. Implemented _udiv128 (on compilers that don't have it built in) using 128 bit types. Some dubious performance testing [here](https://github.com/tomlally/udiv128_test).


Exzap:
LGTM. Thanks!

PR #214, opened 2022-09-09
UI: Auto resize last column

bitscher:
In the list view, auto resize the last column to fit the available window space.

- Fix region column hiding/checkbox and config save.
- Gets rid of magic value -3
- Columns now have a default value specified in an enum rather than arbitrary values in various places in the code
- Minor code refactors in `wxGameList.cpp` (Removed unused function `OnColumnDrag()`, merged `CreateListColumns()` back into the constructor.

### DIFF
--- a/src/config/CemuConfig.cpp
+++ b/src/config/CemuConfig.cpp
@@ -89,12 +89,26 @@ void CemuConfig::Load(XMLConfigParser& parser)
 	auto gamelist = parser.get("GameList");
 	game_list_style = gamelist.get("style", 0);
 	game_list_column_order = gamelist.get("order", "");
-	column_width.name = gamelist.get("name_width", -3);
-	column_width.version = gamelist.get("version_width", -3);
-	column_width.dlc = gamelist.get("dlc_width", -3);
-	column_width.game_time = gamelist.get("game_time_width", -3);
-	column_width.game_started = gamelist.get("game_started_width", -3);
-	column_width.region = gamelist.get("region_width", -3);
+
+	/*
+	 * gamelist.get() relies on TinyXML that itself relies on sscanf to read
+	 * data. sscanf does succeed when reading a negative number into a
+	 * uint32, it's simply cast to an unsigned. This is not what we want
+	 * so we read the signed integer and return the default column width if
+	 * the value in the config is negative.
+	 */
+	auto loadColumnSize = [&gamelist] (const char *name, uint32 defaultWidth) {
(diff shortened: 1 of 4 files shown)


bitscher:
Beware that this change has NOT been tested on the Windows version.

src/config/CemuConfig.cpp
```diff
@@ -89,12 +89,12 @@ void CemuConfig::Load(XMLConfigParser& parser)
 	auto gamelist = parser.get("GameList");
 	game_list_style = gamelist.get("style", 0);
 	game_list_column_order = gamelist.get("order", "");
-	column_width.name = gamelist.get("name_width", -3);
-	column_width.version = gamelist.get("version_width", -3);
-	column_width.dlc = gamelist.get("dlc_width", -3);
-	column_width.game_time = gamelist.get("game_time_width", -3);
-	column_width.game_started = gamelist.get("game_started_width", -3);
-	column_width.region = gamelist.get("region_width", -3);
+	column_width.name = std::max(gamelist.get("name_width", column_width.name), 0u);
```
bitscher, line 92: Please don't merge for now. This actually needs fixing.
tinyxml (and XMLConfig.h) seem to silently cast the `-3` to an unsigned...

Exzap:
Looks mostly ok on Windows except for the icon column being a few pixels too wide:
image

bitscher:
This is due to the +8 pixels I had to add on gnome othewise the image was cropped from the left,
I couldn't find a function to find the internal margins for cells but I will dig deeper.
Alternatively I can remove the vertical lines that this change enabled but did not exist before to make it less obvious.

lijunyu-cn:
There is still a problem somewhere. I tried to adjust the column width in the WinSize event like you did. Most of the time, my code seems to work fine, but when I try something on the window...
image
1. Reduce the height of the window until the vertical scroll bar appears
2. Drag the scroll bar to scroll the topmost item out of the list
3. Increase the height of the window until the vertical scroll bar disappears
Now you will see an empty row on the top of the gamelist, and the gamelist will not redraw correctly. A similar situation occurs when you reduce the window width and then drag the horizontal scroll bar. I tried the 2.0-2 version you released, and the same problem occurred.

Crementif:
Superseded by #265 which is now merged in https://github.com/cemu-project/Cemu/commit/551f82110974b515f07c58130dd29447e305e1b1. Thanks for your contribution!

PR #226, opened 2022-09-09
Check if raw keycode is out of bounds

SSimco:
According to the [wxWidgets documentation](https://docs.wxwidgets.org/3.0/classwx_key_event.html#a6fddcd170d05b0852a7eb2a0cb730795) :
> Under GTK, the raw key code is the keyval field of the corresponding GDK event.

And some keyevents have a raw key code that is greater than 256, which leads to undefined behavior.
In some cases, it crashes the app. For example, when I press the fn+F12 to increase the volume, which has a keyval value of 0x1008ff13 (according to the [GTK source code](https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gdk/gdkkeysyms.h)), the app crashes.

### DIFF
--- a/src/gui/CemuApp.cpp
+++ b/src/gui/CemuApp.cpp
@@ -170,12 +170,16 @@ int CemuApp::FilterEvent(wxEvent& event)
 	{
 		const auto& key_event = (wxKeyEvent&)event;
 		wxGetKeyState(wxKeyCode::WXK_F17);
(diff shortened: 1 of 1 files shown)


Exzap:
This is ok as a quick fix for the out-of-bounds writes. In the future we should look into substituting the keydown array with a `unordered_map` or similar to support any integer keycode.

PR #233, opened 2022-09-10
Fix overlay crash and display correct stats

goeiecool9999:


goeiecool9999:
#142 I didn't even see this before I started working

PR #234, opened 2022-09-10
Fix DSUController on Linux

capitalistspz:
DSUController wasn't showing as an option in input settings, due to an error being thrown about invalid socket options. `SO_RCVTIMEO` on Linux needs to use a `timeval` struct rather than an integer as a parameter.
[POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html#tag_15_10_16)


Exzap:
Why not implement this via `setsockopt()` on unix platforms? As far as I know, you can get the sockfd from a boost socket via `.native_handle()`.

capitalistspz:
> Why not implement this via `setsockopt()` on unix platforms? As far as I know, you can get the sockfd from a boost socket via `.native_handle()`.

I thought it might be cleaner without having inline `#if`/`#elif`/`#endif` directives

Exzap:
Looks good to me now. Thanks!

> I thought it might be cleaner without having inline #if/#elif/#endif directives

Generally yes. But when it's only in one or two places then the shorter solution should be preferred.

PR #238, opened 2022-09-10
More l10n improvements and add workflow to generate POT file

Fs00:
In addition to a few small UI strings improvements, this PR makes the gamepad button names translatable (home, left, right, click), except A, B, X, Y etc. which are always the same in all languages.
To accomplish that, I've used the [wxTRANSLATE](https://docs.wxwidgets.org/3.2/group__group__funcmacro__string.html#gab8862de0f74049708bfea75ae9e14c45) macro to mark button names as translatable (so that they get included in the POT file) and then called [wxGetTranslation](https://docs.wxwidgets.org/3.2/group__group__funcmacro__string.html#ga49dcdc17d82a627b7a50d401f1a37d69) in the GUI (actually, in the `InputPanel`s) to retrieve the translated names.
While I was at it, I've deduplicated a lot of the code that creates the input fields for each controller button in `InputPanel` classes. 

Last but not least, I've added a new GH Actions workflow that generates the .pot file using the `xgettext` command. You can refer to its [man page](https://linux.die.net/man/1/xgettext) to find out what the parameters do, I'm too lazy to explain all of them here rn. :sweat_smile:
The generated POT can be uploaded to the Cemu-Language repo as-is.

### DIFF
--- a/.github/workflows/generate_pot.yml
+++ b/.github/workflows/generate_pot.yml
@@ -1,0 +1,42 @@
+name: Generate translation template
+
+on:
+  pull_request:
+    paths-ignore:
+      - "*.md"
+    types:
+      - opened
+      - synchronize
+      - reopened
+  push:
+    paths-ignore:
+      - "*.md"
+    branches:
+      - main
+
+jobs:
+  generate-pot:
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Checkout repo"
+        uses: actions/checkout@v3
+
+      - name: "Install gettext"
+        run: |
+          sudo apt update -qq
+          sudo apt install -y gettext
+
+      - name: "Generate POT file using xgettext"
+        run: >
+          find src -name *.cpp -o -name *.hpp -o -name *.h | 
+          xargs xgettext --from-code=utf-8
+          -k_ -kwxTRANSLATE -w 100
+          --check=space-ellipsis --omit-header
+          -o cemu.pot
+
+      - name: Upload artifact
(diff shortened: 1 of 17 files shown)


.github/workflows/generate_pot.yml
```diff
@@ -0,0 +1,42 @@
+name: Generate translation template
+
+on:
+  pull_request:
+    paths-ignore:
+      - "*.md"
+    types:
+      - opened
+      - synchronize
+      - reopened
```
Exzap, line 10: Any reason the workflow triggers for pull requests? I don't think we want to generate translation templates for unmerged PRs and the `push` trigger already covers any merged PRs.

Fs00: I think there is value in running the workflow for unmerged pull requests, since I've added the `--check=space-ellipsis` xgettext option. That option makes the command fail if there are any translatable strings that contain spaces before ellipsis (which is considered a punctuation mistake). This way we can see and fix those errors before merging.
Obviously we can also avoid adding this kind of "lint" checks (or add other checks that are listed in the xgettext man page), the decision is up to you :)

Exzap: Alright, we can give this a try and see if it catches any such mistakes. Everything else looks good so I'll merge this now.

Fs00:
Great! Looking forward to have the updated POT on the Cemu-Language repo.
I might open a PR there to add a workflow that builds .mo files, so that you and other maintainers have just to upload them here on this repository to include the updated translations in the app.
Thanks!

PR #239, opened 2022-09-10
Fix analog triggers for DSUController

capitalistspz:


Exzap:
I'm not super familiar with our input code but this looks good to me. Thanks!

PR #240, opened 2022-09-10
Fix Super Smash Bros. SD-card access.

goeiecool9999:
Append function was not implemented and FSGetMountSourceNext wasn't returning the right data.
Append behaviour is probably incorrect like this. I have no idea how I would even begin to find out what is correct.
I also have no clue what anything in the FSGetMountSourceNext function means. I just saw the number was mismatched when Smash Bros. was calling it.


Exzap:
For API changes like with FS AppendFile you should at least analyse/test a few games that use this API. Making changes based on blind faith is pretty risky. To actually figure out API behavior there are many ways, you can always reverse-engineer the affected games to see what kind of behavior they expect or look at what IOSU does.

The SD fix seems quite arbitrary. Granted, the logic we have in place is super hacky as well, but I'd prefer not making it even hackier. It's also unknown how other games will handle this change.

goeiecool9999:
I'm using ``cmd->cmdDefault.destBufferMPTR`` both as destOffset and fileHandle in this patch... I copy-pasted it from writefile without looking. Issue closed because this patch is garbage.

Issue #236, opened 2022-09-10
[Linux] Dark mode display issues


Shoegzer:
If a system-wide dark theme is used, two main issues can be observed:

1. The main window displays title text as white against a white background (white horizontal strip).  The title icons display on the left just fine, however, and the rest of the main window honors the system-wide dark theme.
  
2. Check boxes are not visible, although they can be selected.  For example, under "General", text for options such as "Remember main window position" can be seen although the box itself does not appear.  Clicking the mouse on the text itself does have the intended effect of remembering the window position, however.

Relevant Information:

* OS: Linux Mint 21 (Ubuntu 22.04 base) / kernel 5.15
* System theme: Mint-Y-Dark
* GPU: Nvidia GTX 3070 / driver 515.65
* Commit: 3349d7b

methatronc:
I concur, same specs, successful build, same observation : #2 for me as shown in the attachment.
![image](https://user-images.githubusercontent.com/58328740/189474695-3d90276f-34a5-4c6f-b099-3cc4a13ca3a5.png)

Shoegzer:
Thanks for the screenshot, and yes that captures what I'm seeing.  Just text, no boxes. I assume you're also seeing the same with my first point i.e. white-on-white text in the main window?

kturques:
I'm not a developer, but it seems that Cemu currently have "Game List" background colors hardcoded.
Also, text colors are not set by default, only the background.

I changed the file "Cemu/src/gui/components/wxGameList.cpp" like this:
```
void wxGameList::UpdateItemColors(sint32 startIndex)
{
	wxWindowUpdateLocker lock(this);
	for (int i = startIndex; i < GetItemCount(); ++i)
	{
		const auto titleId = (uint64)GetItemData(i);
		if (GetConfig().IsGameListFavorite(titleId))//entry->favorite)
		{
			SetItemBackgroundColour(i, kFavoriteColor);
			SetItemTextColour(i, 0x000000UL);
		}
		else if ((i&1) != 0)
		{
			SetItemBackgroundColour(i, wxSystemSettings::GetColour(wxSYS_COLOUR_WINDOW));
			SetItemTextColour(i, wxSystemSettings::GetColour(wxSYS_COLOUR_WINDOWTEXT));
		}
		else
		{
			SetItemBackgroundColour(i, wxSystemSettings::GetColour(wxSYS_COLOUR_WINDOWFRAME));
[...]

PR #241, opened 2022-09-10
Add system color definitions to wxGameList

Shoegzer:
As per issue #236, this defines system background and text colors rather than rely on hard-coded values.  This improves display on Linux systems with dark themes.  Please test particularly against light themes and Windows platforms.

Thanks @kturques.


Exzap:
image

Looks like this on Windows. The alternating colors are swapped and the blue is far too intense. It may be difficult to get this right by just relying on system colors. An alternative idea would be to read `wxSYS_COLOUR_MENUBAR` and then dynamically calculate the secondary color from that. Using additive/subtractive logic depending on whether it's dark or bright.

Also, for comparison, the original look:

image

Shoegzer:
Thanks @Exzap!  Digging further, using `wxSYS_COLOUR_MENU` for the secondary row color seems like it would be less intense on Windows.   I also switched the alternating row color so it isn't swapped.  Is this any better?

Exzap:
`wxSYS_COLOUR_MENU` and `wxSYS_COLOUR_MENUBAR` same to be the same color:
image

kturques:
Using "BTNFACE" and "WINDOW" colors with this:
```
wxSYS_COLOUR_BTNFACE
wxSYS_COLOUR_BTNTEXT

wxSYS_COLOUR_WINDOW
wxSYS_COLOUR_WINDOWTEXT
```
I don't know if this is good enough, but I couldn't find the original "blueish" tone, that's the best I got :)

Windows 10
cemu-win
Linux (Adwaita)
cemu-adwaita
Linux (Adwaita Dark)
cemu-adwaita-dark

Shoegzer:
@Exzap Thanks again, that's too bad, those values look different from each other on Linux but it needs to be consistent.   
@kturques Thanks, the contrast is very slight but it least it seems to be consistent across platforms.  I just committed the change, though static values don't seem ideal; as Exzap mentioned it should be reading one of the values and dynamically adjusting based on theme.  I've been trying to find a way to do that and help out but I'm kinda new to wx myself.

Shoegzer:
Okay, after a quick crash course in wx I added some logic to adjust the row colors dynamically based on an attempt to determine the theme.  I really hope this looks decent (it does for me in Linux at least).

@Exzap I added some comments about how to adjust the values, so if it still isn't to your liking for Windows you can either make the changes or if you dont have time just let me know if it needs to be lighter or darker and I'm happy to do it.

Shoegzer:
Thanks for your feedback, that helps.  I did look into the input panel and can probably fix that too, I just want to be sure background and text colors look decent across platforms before taking further steps, with the goal to have consistent colors that are generally agreed to look good no matter the user's system.  

So far I've noticed the following issues impacting visibility / usability on dark themes. I assume most stem from a hard-coded white background color while reading a system font color (which is typically white or light grey for dark themes):

* Main game window row colors (this PR so far)
* Input panel button assignment box colors (the issue you raised above)
* Title manager alternate row colors
* Logging window
* View PPC threads main frame
* PPC Debugger / Memory dump / register view windows
* Texture cache window bottom frame
* About Cemu window

Also, selection boxes are invisible for dark themes (or linux in general?) and need to be fixed.  This is the biggest usability issue in my mind, as it's not possible to tell what selections are active without observing through use.

As a side note I'm curious if there's been any thought or discussion on moving from wx to Qt as other projects have done for various reasons (and wx is kinda rusty comparatively speaking).

Crementif:
> As a side note I'm curious if there's been any thought or discussion on moving from wx to Qt as other projects have done for various reasons (and wx is kinda rusty comparatively speaking).

It's definitely been brought up a few times, but it'd be a considerable undertaking to decouple and rewrite the entire GUI. And while wxWidgets is indeed a bit more clunky, there's definitely downsides to QT that other projects are also now facing. They've done some [unfavorable moves](https://www.qt.io/blog/qt-offering-changes-2020) towards their open-source licensing and it's generally a very bloated library (although it does mean that it doesn't rely on system components to render the GUI which means that you avoid these types of issues :p).

Exzap:
The current Windows look with this PR:
image

I'll merge this now since the original goal of the PR was achieved with the game list looking good on both bright and dark themes. But I do miss the blue tint we had so I'll do a follow-up commit to reintroduce that for very bright themes.

Shoegzer:
@Crementif Thanks for your explanation. I can understand the work involved in such a move and there are no doubt higher priority targets right now anyway. I wasn't aware of the licensing issue you referenced but that's an unfortunate direction.  And yeah it IS bloated, though I'm sure you'd agree the results speak for themselves - it's quite a nice looking library especially with Qt6.  I'll offer my dev help in any way I can should the effort gather steam because I really love this project, though I'd rather not spearhead it.  

@Exzap Thanks for the merge. When I have some more time I'll open an issue and corresponding PR to attempt to address some of the other disparaties with dark themes as well, unless someone else takes it up before then.  If you'd like me to accommodate the blue tint in any of that just let me know and I'll do my best.

PR #247, opened 2022-09-12
Remove cubeb and vulkan-headers submodules

Zopolis4:
Slims down the git repo, and if we are going to use vcpkg we should use it all the way.

Independently, once all the dependencies are on vcpkg and major package managers, can we move to using system package managers on linux and visual studio integration for vcpkg?

That way we dont need to have a dependencies folder at all.


halotroop2288 (approved):

Zopolis4:
Build works on my local machine, appears to be a failure due to vcpkg caching on buildbots.

Zopolis4:
Windows build failure is unrelated to my changes™.

Exzap:
There is a pretty significant downside to using vcpkg ports over git submodules. System headers may not always be up to date for those that compile without vcpkg. It's especially problematic for vulkan-headers where we often need the cutting edge version because of new Vulkan extensions.

I'd would maybe make sense to use vcpkg for cubeb but since it is pretty niche it may not be available on most Linux package managers?

PR #248, opened 2022-09-13
Use unordered_map for keydown

SSimco:
Use an unordered_map instead of using atomic_bool for keeping the states of the key.
Also, instead of using raw key codes I used virtual key codes from wxWidgets, which are mapped to ImGuiKey. This also makes easier to get the button names on Windows and Linux, where the ImGuiKey values are mapped to their string equivalent.


src/gui/guiWrapper.h
```diff
@@ -41,14 +43,44 @@ struct WindowInfo
...
+	};
+	bool get_keystate(uint32 keycode)
+	{
+		const std::lock_guard<std::mutex> lock(keycode_mutex);
+		auto result = m_keydown.find(keycode);
+		if (result == m_keydown.end())
+			return false;
+		return result->second;
+	}
+	std::unordered_map<uint32, bool> get_keystates() const
```
Exzap, line 59: Creating a temporary copy of the entire container is pretty inefficient and in this form it's also not thread-safe.
A better design would be to pass the target map for the copy as a reference, like this:
`void get_keystates(std::unordered_map<uint32, bool>& buttonsOut)`
Then inside the function you can handle the copying/assignment while also locking `keycode_mutex`.

src/input/emulated/EmulatedController.cpp
```diff
@@ -280,7 +280,9 @@ bool EmulatedController::is_mapping_down(uint64 mapping) const
 	if (it != m_mappings.cend())
 	{
 		if (const auto controller = it->second.controller.lock()) {
-			return controller->get_state().buttons.test(it->second.button);
+			auto buttons=controller->get_state().buttons;
```
Exzap, line 283:
```suggestion
			auto& buttons=controller->get_state().buttons;
```
Use a reference here to avoid creating a copy

Exzap (changes requested):
Thanks for looking into this. I gave this a quick test on Windows and it works, but it also breaks compatibility with existing keyboard profiles which is something that needs to be resolved before we can merge this. There are also some minor changes that should be made to the code, see code comments below.

Another concern I have with this is that hardcoding the key names means we have to manually localize them into each language. API like `GetKeyNameTextA` automatically returns the name in the language matching the keyboard layout. It's not a deal breaker but a solution to get the standardized key names from the OS (in a cross-platform way, if possible) would be preferred.

SSimco:
Now it doesn't break compatibility with existing keyboard profiles on windows.

I've changed it back to use `GetKeyNameTextA` to get the name in the language matching the keyboard layout on windows.

On linux I've used gdk_keyval_name to get the key name.

Exzap:
Looks good to me now. I tested this on Windows for a bit and it works as expected with no issues. Thanks!

PR #249, opened 2022-09-14
Properly list files in CMake

Zopolis4:
No more glob :)


src/config/CMakeLists.txt
```diff
@@ -1,8 +1,18 @@
...
-
-file(GLOB CPP_FILES *.cpp)
-file(GLOB H_FILES *.h)
-add_library(CemuConfig ${CPP_FILES} ${H_FILES})
+add_library(CemuConfig
+    ActiveSettings.cpp
+    ActiveSettings.h
+    CemuConfig.cpp
+    CemuConfig.h
+    CMakeLists.txt
```
Exzap, line 6: Copy & paste mistake?

Zopolis4: Yeah-- I used `tree --noreport -i -f` but that kept directories without files and CMakeLists.txt. Fixed.

Exzap:
LGTM

Tachi107:
I'm leaving this comment here for future reference:

Listing header files in `add_library()` and `add_executable()` is only necessary to make them show up in IDEs, but it is not needed for compilation (and it is kinda confusing in my opinion). New CMake versions allow to split headers from source files while still allowing IDEs to mark them as associated to a particular target with [`target_sources(FILE_SET)`](https://cmake.org/cmake/help/latest/command/target_sources.html#file-sets); it requires CMake 3.23.

Using file sets allows CMake to understand which files are really source files and which are really headers, but apart from making the code a bit clearer and making `install()`s more concise I don't know if/how it can improve things during the building phase.

PR #251, opened 2022-09-15
Implement proper microphone support

AGraber:
This PR implements microphone support by more or less completing the existing mic HLE lib alongside a new `IAudioInputAPI` interface similar to the normal one for output. Only supports cubeb for now but can be extended to DirectSound with another `IAudioInputAPI` implementation in the future.

A more detailed overview of the changes:

- Made a new `IAudioInputAPI` interface for handling audio input, since there were too many differences to how stuff is handled to include it `IAudioAPI` interface. Currently only implemented by `CubebInputAPI` but can be implemented by other APIs.
- Implemented feeding microphone samples from `IAudioInputAPI`, falls back to generating samples with controller button if device is disabled.
- Changed `MIC_HANDLE_DRC0` and `MIC_HANDLE_DRC1` from 100/101 to 0/1 respectively, Mario Kart 8 considers non-zero values as an error and stops initializing the rest of the microphone. It saves the value returned by `MICInit` (which is effectively 0) and uses it for all other mic functions as the index so I'm pretty sure the handle games expect is 0. 
- Implement settings for the `IAudioInputAPI` stuff
- Rearranged the logic for recreating audio devices a bit, this fixes IAudio(Input)API pointers not being destroyed when selecting "Disabled" on the audio settings menu.

### DIFF
--- a/src/Cafe/CafeSystem.cpp
+++ b/src/Cafe/CafeSystem.cpp
@@ -5,6 +5,7 @@
 #include "Cafe/HW/Espresso/Interpreter/PPCInterpreterInternal.h"
 #include "Cafe/HW/Espresso/Recompiler/PPCRecompiler.h"
 #include "audio/IAudioAPI.h"
+#include "audio/IAudioInputAPI.h"
 #include "Cafe/HW/Espresso/Debugger/Debugger.h"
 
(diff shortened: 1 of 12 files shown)


shydert:
Why is this not merged yet?

PR #253, opened 2022-09-15
Clean up more Cemuhook leftovers

Fs00:
This PR aims to remove the last unnecessary mentions and traces of Cemuhook from the codebase.
I've taken this opportunity to remove a few fields that have been kept around only for compatibility with Cemuhook, please tell me if I have removed too much :smile:


Fs00:
Btw, while removing dead code I've noticed some weird logic here:
https://github.com/cemu-project/Cemu/blob/e8702fed67a0bde8370e1ca73ae0f681d14cbf31/src/Cafe/OS/libs/coreinit/coreinit_DynLoad.cpp#L87-L104
The `if` at line 103 will never be true since the `if` at line 97 sets the `rplHandle` to 0 if it's invalid... what's the correct logic to have here?

Exzap:
Looks good

> The `if` at line 103 will never be true since the `if` at line 97 sets the `rplHandle` to 0 in case it's invalid... what's the correct logic to have here?

Ah good catch. Seems like a regression that happened at some point, not sure if any games actually care about the return value tho. I'll fix it in a follow-up commit.

PR #255, opened 2022-09-15
[Linux] [CMake] Add Findglslang / FindSprivTools

rtissera:
Hi,

Adding these two cmake files permissively licensed ease the Linux build a lot in my scenario (cross-compiling within buildroot).
Hope it can be merged and will help.

### DIFF
--- a/cmake/FindSpirvTools.cmake
+++ b/cmake/FindSpirvTools.cmake
@@ -1,0 +1,174 @@
+#.rst:
+# Find SPIRV-Tools
+# ----------------
+#
+# Finds the SpirvTools library. This module defines:
+#
+#  SpirvTools_FOUND          - True if SPIRV-Tools are found
+#  SpirvTools::SpirvTools    - SpirvTools imported target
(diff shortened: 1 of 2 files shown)


Tachi107:
Hi @rtissera, what issue are you actually trying to solve? Adding 400 lines of CMake code to solve one issue means having to debug 400 more lines of code if they stop working...

By taking a quick look at the Find modules I saw that they try to work around limitations and issues of old CMake and glslang versions, and I'm not sure that's needed for Cemu; the minium required CMake version here is 3.21.1, and while I don't know what the minimum required glslang version is, the Find modules work around issues found in ancient glslang versions like 7.x, adding unneeded complexity.

Could you please paste a log showing the issues you're having, possibly in a new issue? It may be possible that there's a simpler solution, or that the issue is in your build environment.

rtissera:
Sorry for the trouble, I was badly outdated on my glslang setup, latest stable release works fine without adding such CMake additionals files. Closing PR :)

Tachi107:
Nice to hear, thanks :)

PR #258, opened 2022-09-15
Code refactor, fixes and minor optimizations

GermanAizek:
Hi @Exzap, @Crementif,
This PR will be correcting the code for better readability, bringing it to the latest C++ standard, using some optimizations and fixing bugs.

### DIFF
--- a/src/Cafe/CafeSystem.cpp
+++ b/src/Cafe/CafeSystem.cpp
@@ -638,7 +638,7 @@ namespace CafeSystem
 		FSCDeviceHostFS_Mount(std::string("/internal/code/").c_str(), _pathToUtf8(executablePath.parent_path()), FSC_PRIORITY_BASE);
 		std::string internalExecutablePath = "/internal/code/";
 		internalExecutablePath.append(_pathToUtf8(executablePath.filename()));
-		_pathToExecutable = std::move(internalExecutablePath);
+		_pathToExecutable = internalExecutablePath;
 		// since a lot of systems (including save folder location) rely on a TitleId, we derive a placeholder id from the executable hash
 		auto execData = fsc_extractFile(_pathToExecutable.c_str());
 		if (!execData)
--- a/src/Cafe/CafeSystem.cpp
+++ b/src/Cafe/CafeSystem.cpp
@@ -577,7 +577,7 @@ namespace CafeSystem
 					const auto file = fsc_open(rpxPath.c_str(), FSC_ACCESS_FLAG::OPEN_FILE | FSC_ACCESS_FLAG::READ_PERMISSION, &status);
 					if (file)
 					{
-						_pathToExecutable = rpxPath;
+						_pathToExecutable = std::move(rpxPath);
 						fsc_close(file);
 					}
 				}
@@ -638,7 +638,7 @@ namespace CafeSystem
 		FSCDeviceHostFS_Mount(std::string("/internal/code/").c_str(), _pathToUtf8(executablePath.parent_path()), FSC_PRIORITY_BASE);
 		std::string internalExecutablePath = "/internal/code/";
 		internalExecutablePath.append(_pathToUtf8(executablePath.filename()));
-		_pathToExecutable = internalExecutablePath;
+		_pathToExecutable = std::move(internalExecutablePath);
 		// since a lot of systems (including save folder location) rely on a TitleId, we derive a placeholder id from the executable hash
 		auto execData = fsc_extractFile(_pathToExecutable.c_str());
 		if (!execData)
--- a/src/Cafe/GraphicPack/GraphicPack2PatchesApply.cpp
+++ b/src/Cafe/GraphicPack/GraphicPack2PatchesApply.cpp
@@ -125,6 +125,7 @@ double _cbResolveConstant(std::string_view varname)
 		if (idxDot != std::string_view::npos)
 		{
 			std::string moduleName = importName.substr(0, idxDot);
+			std::string functionName = importName.substr(idxDot + 1);
 			uint32 rplHandle = RPLLoader_GetHandleByModuleName(moduleName.c_str());
 			if (rplHandle == RPL_INVALID_HANDLE)
 			{
@@ -132,7 +133,6 @@ double _cbResolveConstant(std::string_view varname)
 			}
 			else
 			{
-				std::string functionName = importName.substr(idxDot + 1);
 				MPTR exportResult = RPLLoader_FindModuleOrHLEExport(rplHandle, false, functionName.c_str());
 				if (exportResult)
 				{
--- a/src/Cafe/GraphicPack/GraphicPack2PatchesApply.cpp
+++ b/src/Cafe/GraphicPack/GraphicPack2PatchesApply.cpp
@@ -125,7 +125,6 @@ double _cbResolveConstant(std::string_view varname)
 		if (idxDot != std::string_view::npos)
 		{
 			std::string moduleName = importName.substr(0, idxDot);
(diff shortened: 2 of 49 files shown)


src/Cafe/CafeSystem.cpp
```diff
@@ -638,7 +638,7 @@ namespace CafeSystem
 		FSCDeviceHostFS_Mount(std::string("/internal/code/").c_str(), _pathToUtf8(executablePath.parent_path()), FSC_PRIORITY_BASE);
 		std::string internalExecutablePath = "/internal/code/";
 		internalExecutablePath.append(_pathToUtf8(executablePath.filename()));
-		_pathToExecutable = internalExecutablePath;
+		_pathToExecutable = std::move(internalExecutablePath);
```
Exzap, line 641: Needlessly verbose and may trip up someone in the future who tries to use either `rpxPath` or `internalExecutablePath`. This code runs once on game launch so it's not a worthy candidate for shaving of CPU microseconds anyway.

src/Cafe/GraphicPack/GraphicPack2PatchesApply.cpp
```diff
@@ -125,14 +125,14 @@ double _cbResolveConstant(std::string_view varname)
...
 			std::string moduleName = importName.substr(0, idxDot);
-			std::string functionName = importName.substr(idxDot + 1);
 			uint32 rplHandle = RPLLoader_GetHandleByModuleName(moduleName.c_str());
 			if (rplHandle == RPL_INVALID_HANDLE)
 			{
 				importError = " (module not found)";
 			}
 			else
 			{
+				std::string functionName = importName.substr(idxDot + 1);
```
Exzap, line 135: Ripping apart the two similar `.substr()` lines worsens readability imho. I assume this was done as an optimization but the else case is taken virtually always as it's the non-error case.

src/Cafe/HW/Latte/Core/LatteRenderTarget.cpp
```diff
@@ -486,38 +486,41 @@ bool LatteMRT::UpdateCurrentFBO()
...
-		// set effective size
-		sint32 effectiveWidth, effectiveHeight;
-		LatteTexture_getEffectiveSize(colorAttachmentView->baseTexture, &effectiveWidth, &effectiveHeight, nullptr, colorAttachmentView->firstMip);
-		if( rtEffectiveSize->width == 0 && rtEffectiveSize->height == 0 )
-		{
-			rtEffectiveSize->width = effectiveWidth;
-			rtEffectiveSize->height = effectiveHeight;
-		}
-		else if( rtEffectiveSize->width != effectiveWidth && rtEffectiveSize->height != effectiveHeight )
+		if (colorAttachmentView)
```
Exzap, line 489: Unnecessary null pointer check. GetColorAttachmentTexture() will never return null.

src/Cafe/HW/Latte/LegacyShaderDecompiler/LatteDecompilerAnalyzer.cpp
```diff
@@ -850,9 +850,12 @@ void LatteDecompiler_analyze(LatteDecompilerShaderContext* shaderContext, LatteD
 			cemu_assert_unimplemented();
 	}
 	// analyze subroutines
-	for (auto subroutineAddr : list_subroutineAddrs)
+	if (!list_subroutineAddrs.empty())
```
Exzap, line 853: What is the advantage of checking if non-empty for a range based loop?

src/Cafe/OS/libs/vpad/vpad.cpp
```diff
@@ -905,11 +905,11 @@ void vpadExport_VPADStopMotor(PPCInterpreter_t* hCPU)
 void vpadExport_VPADSetTPCalibrationParam(PPCInterpreter_t* hCPU)
 {
 	ppcDefineParamU32(channel, 0);
-	ppcDefineParamMEMPTR(params, VPADTPCalibrationParam, 1);
 
 	debugBreakpoint();
 	if (channel < VPAD_MAX_CONTROLLERS)
 	{
+		ppcDefineParamMEMPTR(params, VPADTPCalibrationParam, 1);
```
Exzap, line 912: Parameter initialization should be at the beginning of the function. There is also no performance gained as this code path is taken 99.99% of time.

src/Cemu/FileCache/FileCache.cpp
```diff
@@ -287,7 +287,10 @@ uint8* _fileCache_compressFileData(const uint8* fileData, uint32 fileSize, sint3
 	Bytef* compressedData = (Bytef*)malloc(4 + compressedLen);
 	int zret = compress2(compressedData + 4, &compressedLen, uncompressedInput, uncompressedLen, 4); // level 4 has good compression to performance ratio
 	if (zret != Z_OK)
+	{
+		delete compressedData;
```
Exzap, line 291:
```suggestion
		free(compressedData);
```
Delete is wrong here as the memory is allocated with `malloc`

src/Cemu/nex/nexFriends.h
```diff
@@ -175,10 +172,10 @@ class nexPrincipalBasicInfo : public nexType
 	}
 
 	nexPrincipalBasicInfo(uint32 principalId, char* nnid, const nexMiiV2& mii)
+		: mii(mii)
```
Exzap, line 175: `principalId` should also be moved into the initializer list

src/Cemu/nex/nexFriends.h
```diff
@@ -218,10 +215,10 @@ class nexNNAInfo : public nexType
 	}
 
 	nexNNAInfo(uint8 countryCode, uint8 countrySubCode, const nexPrincipalBasicInfo& principalBasicInfo)
+		: principalInfo(principalBasicInfo)
```
Exzap, line 218: `countryCode` and `countrySubCode` should also be part of the initializer list

src/input/InputManager.cpp
```diff
@@ -225,14 +225,14 @@ bool InputManager::migrate_config(const fs::path& file_path)
...
 		const auto api_string = m_data.get<std::string>("General.api");
-		auto uuid_opt = m_data.get_optional<std::string>("General.controller");
 		const auto display_name = m_data.get_optional<std::string>("General.display");
 
 		std::string uuid;
 		if (api_string == to_string(InputAPI::Keyboard))
 			uuid = to_string(InputAPI::Keyboard);
 		else
 		{
+			auto uuid_opt = m_data.get_optional<std::string>("General.controller");
```
Exzap, line 235: Seems like an unnecessary micro-optimization that only sacrifices readability?

Exzap (changes requested):
Overall this is good. Some of it is pretty subjective but still acceptable. However, a few of the proposed changes are micro-optimizations at the cost of code quality which need to be addressed.

src/Cafe/HW/Latte/LegacyShaderDecompiler/LatteDecompilerAnalyzer.cpp
GermanAizek: Yes, you're right, after C++11, don't need write this
Reference: https://stackoverflow.com/questions/52828423/in-c11-and-later-is-it-still-optimal-to-do-an-empty-check-before-a-range-f

GermanAizek:
@Exzap, fixed your changes here https://github.com/cemu-project/Cemu/pull/258/commits/89c0945f23a31f2a55bae75f22f03d49cba7d785

Exzap:
Looks good. Thanks!

PR #264, opened 2022-09-17
GX2: Add workaround for FFL uninitialized texture parameters

Exzap:


AMDBartek:
I can confirm that this commit fixed the issue, thank you and the Cemu team for the hard work.

PR #268, opened 2022-09-18
Fix some issues with building on macOS

Tillsunset:
This fixes some issues when building on macOS and for macOS

Reasoning behind changes
1. For some reason when building on macOS, OBJC and OBJCXX need to be explicitly enabled else an error involving [precompiled.h](https://github.com/cemu-project/Cemu/tree/main/src/Common) occurs. Is only enabled on macOS, no side effects. Credits to OatmealDome
2. changing strong ordering to weak ordering. There exists implicit conversion from strong to weak and in our case it shouldn't produce any side effects. Three way comparison on strings produces weak_ordering which can't be converted to strong_ordering. Functionality in all cases should be kept the same while adding support for macOS.
3. Uses the raw PNG bytes found in [resources/embedded](https://github.com/cemu-project/Cemu/tree/main/src/resource/embedded). Was tested and produces no side effects on Windows and Linux

To Do list for a separate PR 
1. Some string encoding issues involving boost::iequals
2. GitHub workflow for macOS

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,6 +54,10 @@ if (MSVC)
 	endif()
 endif()
 
+if (APPLE)
+    enable_language(OBJC OBJCXX)
+endif()
+
 option(ENABLE_OPENGL "Enables the OpenGL backend" ON)
 option(ENABLE_VULKAN "Enables the Vulkan backend" ON)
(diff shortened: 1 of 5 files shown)


Exzap:
LGTM. Thanks!

PR #271, opened 2022-09-19
Refactor CI

Zopolis4:
Refactor the files, and add builds for gcc+make and clang-cl+ninja.

This uses a different form of caching, one that dosen't pollute the Cemu repo githubside with packages, and also saves build folders to speed up compilation even more.


Exzap:
This PR is still marked as draft so I assume it isn't finished, but even now it's already going far beyond just "refactoring". This has multiple unrelated changes:
- Changing handling of PUBLIC_RELEASE flag
- Adding new CI build targets
- Replacing CI steps with external actions
- Actual refactoring of the CI scripts

The CI is pretty critical so we have to be extra meticulous when reviewing. Please submit as separate PRs so each one can be reviewed, discussed and merged individually.

PR #277, opened 2022-09-20
fixed the string encoding issue on macOS

Tillsunset:
this fixes the last remaining issue when building on macOS using clang 14


UltraHDR:
@Tillsunset Can you please update the BUILD.md? Thanjs

PR #285, opened 2022-09-23
AppImage Build Support

donn:
This adds support for building AppImages to Cemu, now that #382 and #356 have been merged.

You can build an AppImage by `cd`ing into `appimage` and just running `make`.

I compiled Cemu on a Docker image running CentOS 7 as it's reasonably old but still supported for another year or so (the older your Glibc is, the more compatible your binaries are), wrote a script to package all the shared libraries, and then created an AppImage that's ready for use. I tested it on my Manjaro box and an Ubuntu VM, both of which booted fine.

Two caveats though:

* The user will still manually need to install libpulse (or a compatible library like pipewire-pulse) and libopengl. I don't think there's a way around that.
* There is this quirk on Manjaro:

image

It doesn't occur on my Pop! VM and idk what to make of it exactly, but it's worth pointing out.

If you want my honest opinion, I think AppImages, despite being the most "out-of-the-box" compatible of all the Linux distros, are a bit jank. I will keep this PR open for discussion/posterity.
 
---
## Original Post

I tried my hand at this to see how far I'd get.

I compiled Cemu on a Docker image running CentOS 7 as it's reasonably old (the older your Glibc is, the more compatible your binaries are), wrote a script to package all the shared libraries, and then tested that on my host machine.

So far, the binary actually works and boots into games, but there are some graphical glitches with the UI itself: 
image

That alone would be tolerable (if annoying), but there are some more pressing issues:

* ~~There is no audio at all. Cubeb shows up as an audio backend, but it doesn't appear to be playing back anything.~~
   * ~~Matter of fact, it doesn't look like the binary is linked to any audio library, period.~~
   * Fixed by installing pulseaudio-devel on the Docker image it's being compiled in
* ~~It's not detecting controllers.~~
  * nvm, it is, the dropdown button wasn't rendering properly
* I'm sure this discussion is being had in relation to the Mac ports somewhere (I can't find it) but using the binary path to save settings.xml and other files is not going to work: AppImages are mounted as a read-only filesystem, with the intention to have these kinds of files to a directory under Documents or `~/.cemu` or something: https://github.com/cemu-project/Cemu/blob/b720d17a977d4d090926ddbb6b6e6e81cb44d37e/src/config/ActiveSettings.cpp#L17
  * EDIT: Yeah, this: https://github.com/cemu-project/Cemu/issues/107#issuecomment-1230678642

Anyway, I'm gonna keep working on this and see if I get anywhere eventually.


qurious-pixel:
In `src/config/ActiveSettings.cpp` it is enough to change 
`s_path = s_full_path.parent_path();` 
to 
`s_path = std::getenv("XDG_DATA_HOME");`

You may run into an issue loading the wxWidget file picker due to libpng mismatch. I was able to resolve by building wxWidgets from source instead of vcpkg.

The final issue was gamepad window not loading the canvas on OpenGL, but working correctly with Vulkan.

That and Ctrl or Ctrl+Tab not invoking it as it did on Wine/Windows.

donn:
Thanks for the tip on `s_path`: guessing we'd have to implement this as OS-specific behavior though. 

> You may run into an issue loading the wxWidget file picker due to libpng mismatch. I was able to resolve by building wxWidgets from source instead of vcpkg.

I'll have to look into how I can exclude libpng from vcpkg- it's my first time dealing with this build system as you can probably tell. 

> Edit: if you are using a newer compiler than the target OS, you may have to bundle the libs and patch the appimage or it will have cxx errors when loading.

AppImage doesn't recommend bundling libstdc++, but I'll test it on a couple distros and see if we ultimately need to anyway. I did have to modify their excludelist.

PR #300, opened 2022-09-25
Fix crash on GTK when a preset value is changed

SSimco:


Exzap:
I suggest two improvements:
- Move workaround code into a separate function to keep it more isolated
- Adding a short comment explaining what the workaround is for

SSimco:
Now the workaround code is in a separate function and I've added an explanation for the need of the workaround.

Exzap:
Thanks!

PR #302, opened 2022-09-25
Added Native Pretendo and custom URLs support

MythicalPlayz:
This PR adds support for third party servers (currently Pretendo only but can be expanded later) support natively and Indication on top of which Network Service the user is using.

This also Adds an option to disable SSL Verification checkbox if third party servers do not have SSL (like Pretendo atm).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ bin/Cemu_*.ilk
 bin/Cemu.exe.backup
 bin/mlc01/*
 bin/settings.xml
+bin/network_services.xml
 bin/title_list_cache.xml
 bin/debugger/*
 bin/sdcard/*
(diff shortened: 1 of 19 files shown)


MythicalPlayz:
image
This is what the Account menu Looks

(if anyone can improve the placement of the new Items that would be great)

jonbarrow:
Maybe it would be cleaner to replace the parts of the URL string during the request instead of returning new ones? This would likely reduce the number of checks needed. So when Cemu goes to make a request to the servers, do a check for the active then and run a replace on the URL to change `nintendo.net` to `pretendo.cc`?

Alecerzea:
How can I download or compile these PR??

MythicalPlayz:
Look in my Actions and find the compiled versions (Except MacOS)

Exzap (changes requested):
Structurally this is pretty good and it integrates well with the existing config code.
Some smaller improvements can be made for cleaner code, see comments.

latitudeh:
Any updates on when this will be merged?

Exzap (approved):

Exzap:
LGTM. Not sure if the MacOS compilation error is resolved but if not it should be an easy fix which I will look into after merging.

PR #304, opened 2022-09-25
fix ninja binary path in macos instructions

emiyl:
This PR fixes the path of the ninja binary, which is different on macOS than it is on Linux.


Pedro-Beirao:
/usr/local is the place of the x64 brew
/opt/homebrew is the place of the arm64 brew

Cemu can only be compiled on x64, so /usr/local is the correct path

PR #310, opened 2022-09-28
service: nfp: Fix corruption, correct structs and use write counters

german77:
Improves the amiibo implementation by correcting the NFC structs, fixing the corrupted bytes of `dynamic lock` and incrementing/reporting the write counters.

With this change amiibos should be compatible with yuzu. Some games that reported `different amiibo detected` after writing or deleting some data should be fixed as well.


Exzap:
Looks good and I found no issues while testing. Thanks!

PR #312, opened 2022-09-28
Linux: Print demangled symbols on backtrace

goeiecool9999:
This PR does the following:
Crash backtraces go from looking like this:
image
To looking like this:
image

The PR also increases the amount of lines cemu tries to retrieve because I've encountered a situation where the backtrace was incomplete.

Caveats:
- For a process to be able to get symbol information from linux's backtrace functions the linker flag `-rdynamic` needs to be enabled. This increases Release executable size by 4.4 MiB.
- Symbols are exported implicitly.
- I'm not sure the addr2line tool supports de-mangled symbol names so people using it may run into trouble. This could be a problem if a bug is hard to reproduce and all they have is a stacktrace, though AFAIK that tool requires a debug build to function anyway and I don't think it'll be common for people to find hard to reproduce bugs while using a debug build because I expect most people to play games using release builds. Of course they'll still have some indication of the general area where the crash happens.

### DIFF
--- a/src/Common/ExceptionHandler/ExceptionHandler_posix.cpp
+++ b/src/Common/ExceptionHandler/ExceptionHandler_posix.cpp
@@ -1,8 +1,42 @@
 #include <signal.h>
 #include <execinfo.h>
 #include <string.h>
+#include <string>
 
 #include "config/CemuConfig.h"
+
+void demangleAndPrintBacktrace(char** backtrace, size_t size)
+{
+	for (char** i = backtrace; i < backtrace + size; i++)
+	{
+		std::string traceLine{*i};
+		size_t parenthesesOpen = traceLine.find_last_of('(');
+		size_t parenthesesClose = traceLine.find_last_of(')');
+		size_t offsetPlus = traceLine.find_last_of('+');
+		if (!parenthesesOpen || !parenthesesClose || !offsetPlus ||
+			 offsetPlus < parenthesesOpen || offsetPlus > parenthesesClose)
+		{
+			// something unexpected was read. fall back to default string
+			std::cerr << traceLine << std::endl;
+			continue;
+		}
+
+		std::string symbolName = traceLine.substr(parenthesesOpen+1,offsetPlus-parenthesesOpen-1);
+		int status = -1;
+		char* demangled = abi::__cxa_demangle(symbolName.c_str(), nullptr, nullptr, &status);
+		if (demangled)
+		{
+			std::cerr << traceLine.substr(0, parenthesesOpen+1);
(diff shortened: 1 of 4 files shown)


src/Common/ExceptionHandler/ExceptionHandler_posix.cpp
```diff
@@ -1,9 +1,43 @@
...
+			 offsetPlus < parenthesesOpen || offsetPlus > parenthesesClose)
+		{
+			// something unexpected was read. fall back to default string
+			std::cerr << traceLine << std::endl;
+			continue;
+		}
+
+		std::string symbolName = traceLine.substr(parenthesesOpen+1,offsetPlus-parenthesesOpen-1);
+		int status = -1;
+		char* demangled = abi::__cxa_demangle(symbolName.c_str(), nullptr, nullptr, &status);
```
Tachi107, line 26: This is GNU libstdc++ specific, shouldn't it be `ifdef`d? I believe macOS users use LLVM libc++

goeiecool9999: Clang implements the same function so it should be okay. GitHub CI does build for MacOS and they seem to have passed so I don't expect any issues.

goeiecool9999:
-rdynamic breaks vulkan loading.

src/Common/ExceptionHandler/ExceptionHandler_posix.cpp
Tachi107: Does this still work when compiling with `-stdlib=libc++`?

goeiecool9999: I tried to compile with that option but a vcpkg library is compiled without it causing linker errors. I don't know how to change the compiler options for vcpkg dependencies.

goeiecool9999: I found out how to change the compiler option for vcpkg libraries but I can't use clang to compile them and my GCC installation is not compiled with -stdlib support. So I can't test this. However a bigger problem for this code on Mac OS is problably the format of the output of backtrace_symbols. Best case scenario if it's different the sanity checks will probably fail and it will just print the raw output.

Exzap:
As discussed on Discord, we'd prefer a solution that would read the debug info from the ELF symbol table rather than force the export of every function. But since this would require using a more sophisticated back-trace library I think this PR is an acceptable intermediate solution. 

The only request I have is that `-rdynamic` should not be emitted for Apple platforms since it is not necessary there.

PR #314, opened 2022-09-30
Workaround for pixel format issue on MacOS

Tillsunset:
Workaround for pixel format on MacOS, R5G6B5_UNORM_PACK16 not supported on Molten-VK. This just an initial fix, there are more format that also aren't supported


emiyl (approved):
Seems to work great and fixes BOTW crashing on boot 👍

Exzap:
`TextureDecoder_R8_G8_B8_A8` expects R8G8B8A8 format as input on the Wii U side and outputs R8G8B8A8 on the Vulkan side which is not what you want. Currently it just breaks the texture and reads pixels out of bounds since the input is still R5B6G5 even if you substitute on the Vulkan side. If the format has to change on the fly (from R5G6B5 to R8G8B8A8) it gets a bit more tricky and a specialized decoder has to be written. There isn't a great example for this but you can look at the usage and implementation of `TextureDecoder_R4_G4_UNORM_toRGBA4444_vk` as a base. You can try implementing this yourself for `R5G6B5->R8G8B8A8` but if it's too difficult I can help with this. Just let me know. For workaround purposes you can also just set `formatInfoOut->decoder = nullptr;` because no decoder is better than the wrong one. That would be good enough for merging.

We also have a mechanism for dynamically detecting supported texture formats at runtime. Example code [here](https://github.com/cemu-project/Cemu/blob/37672572204230a2c4cd6d1282ccf48b51afb259/src/Cafe/HW/Latte/Renderer/Vulkan/VulkanRenderer.cpp#L2566) and [here](https://github.com/cemu-project/Cemu/blob/37672572204230a2c4cd6d1282ccf48b51afb259/src/Cafe/HW/Latte/Renderer/Vulkan/VulkanRenderer.cpp#L1930). You should use this over hardcoding `#if BOOST_OS_MACOS` because then Cemu can leverage native format support if it ever gets added to Metal/MoltenVk in the future.

PR #316, opened 2022-09-30
Add support for non portable mode

SSimco:
This PR adds support for reading/saving files in standard directories locations.
By default all files are read/saved in the same directory where the executable file is located.

If portable mode is turned off then the files locations will be:

Config path:
- Linux: `$XDG_CONFIG_HOME/cemu`
- Windows: `C:\Users\username\AppData\Roaming\cemu` or `C:\Documents and Settings\username\Application Data\cemu`
- Mac: `~/Library/Application Support/cemu`

User data path:
- Linux: `$XDG_DATA_HOME/cemu`
- Windows: `C:\Users\username\AppData\Roaming\cemu` or `C:\Documents and Settings\username\Application Data\cemu`
- Mac: `~/Library/Application Support/cemu`

Data path (application global files, e.g. gameProfiles default files or MO files):
- Linux: `prefix/share/cemu`
- Windows: same directory where the executable file is located
- Mac: `cemu.app/Contents/SharedSupport` bundle subdirectory

Cache path:
- Linux: `~/.cache/cemu`
- Windows: `C:\Users\username\AppData\Local\cemu`
- Mac: `~/Library/Caches/cemu`

### DIFF
--- a/src/Cafe/GraphicPack/GraphicPack2.cpp
+++ b/src/Cafe/GraphicPack/GraphicPack2.cpp
@@ -63,7 +63,7 @@ void GraphicPack2::LoadGraphicPack(fs::path graphicPackPath)
 void GraphicPack2::LoadAll()
 {
 	std::error_code ec;
-	fs::path basePath = ActiveSettings::GetPath("graphicPacks");
+	fs::path basePath = ActiveSettings::GetUserDataPath() / "graphicPacks";
 	for (fs::recursive_directory_iterator it(basePath, ec); it != end(it); ++it)
 	{
 		if (!it->is_directory(ec))
@@ -93,7 +93,7 @@ bool GraphicPack2::LoadGraphicPack(const std::wstring& filename, IniParser& rule
 		if (it == config_entries.cend())
 		{
 			// check for relative path
-			it = config_entries.find(MakeRelativePath(gp->GetFilename2()).lexically_normal());
+			it = config_entries.find(MakeRelativePath(ActiveSettings::GetUserDataPath(), gp->GetFilename2()).lexically_normal());
 		}
 
 		if (it != config_entries.cend())
--- a/src/Cafe/GraphicPack/GraphicPack2.cpp
+++ b/src/Cafe/GraphicPack/GraphicPack2.cpp
@@ -63,7 +63,7 @@ void GraphicPack2::LoadGraphicPack(fs::path graphicPackPath)
 void GraphicPack2::LoadAll()
 {
 	std::error_code ec;
-	fs::path basePath = ActiveSettings::GetUserDataPath() / "graphicPacks";
+	fs::path basePath = ActiveSettings::GetUserDataPath("graphicPacks");
 	for (fs::recursive_directory_iterator it(basePath, ec); it != end(it); ++it)
 	{
 		if (!it->is_directory(ec))
--- a/src/config/ActiveSettings.cpp
+++ b/src/config/ActiveSettings.cpp
@@ -37,7 +37,7 @@ void ActiveSettings::LoadOnce(const fs::path& user_data_path,
 
 	s_filename = s_full_path.filename();
 
-	g_config.SetFilename(GetPath(GetConfigPath(), "settings.xml").generic_wstring());
+	g_config.SetFilename(GetConfigPath("settings.xml").generic_wstring());
 	g_config.Load();
 
 	std::wstring additionalErrorInfo;
--- a/src/config/ActiveSettings.cpp
+++ b/src/config/ActiveSettings.cpp
@@ -247,6 +247,6 @@ fs::path ActiveSettings::GetMlcPath()
 
 fs::path ActiveSettings::GetDefaultMLCPath()
 {
-	return GetPath(GetUserDataPath(), "mlc01");
+	return GetUserDataPath("mlc01");
 }
 
--- a/src/config/ActiveSettings.cpp
+++ b/src/config/ActiveSettings.cpp
@@ -1,6 +1,7 @@
 #include "config/ActiveSettings.h"
 
 #include "Cafe/GameProfile/GameProfile.h"
+#include "Cemu/Logging/CemuLogging.h"
 #include "LaunchSettings.h"
 #include "util/helpers/helpers.h"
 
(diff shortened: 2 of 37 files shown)


src/gui/ChecksumTool.cpp
```diff
@@ -186,7 +186,7 @@ void ChecksumTool::LoadOnlineData() const
 
 		std::string latest_commit;
 
-		const auto checksum_path = ActiveSettings::GetPath("resources/checksums");
+		const auto checksum_path = GetPath(ActiveSettings::GetUserDataPath(), "resources/checksums");
```
goeiecool9999, line 189: I see this all over the code changes. Why not leave the GetPath functions in the ActiveSettings class and add more functions like ActiveSettings::GetPathUserData and ActiveSettings::GetPathConfig that just call GetPath and prepend the corresponding paths? Would make for a cleaner patch since ActiveSettings.cpp doesn't get modified as much and make the code more readable in my opinion.

goeiecool9999: There's probably some clever way to roll GetMLCPath and the various other Path functions into one. Looks like GetPath and GetMLCPath already share a lot of code. Beware of the changes in commit 168ecf8825813c570793510a0314130c3fb18385. It made some Path functions use the _utf8ToPath function and others not. If we make it all consistent it should fit really neatly I think.

goeiecool9999: I have a PR in the works that will address this.

goeiecool9999: Nevermind you beat me to it!

SSimco: Now all GetXPath methods can be called in a similar way as the original GetPath method.

src/gui/CemuApp.cpp
```diff
@@ -70,6 +73,38 @@ void unused_translation_dummy()
 
 bool CemuApp::OnInit()
 {
+	fs::path user_data_path, config_path, cache_path, data_path;
+	auto standardPaths = wxStandardPaths::Get();
+#ifdef PORTABLE
+	fs::path exePath(standardPaths.GetExecutablePath().ToStdString());
+	user_data_path = config_path = cache_path = data_path = exePath.parent_path();
+#else
+	SetAppName("cemu");
```
UltraHDR, line 82:
```suggestion
	SetAppName("Cemu");
```
Should this be capital?

src/config/ActiveSettings.cpp
```diff
@@ -12,13 +13,31 @@
...
+			fs::create_directories(path, ec);
+		}
+		if (!TestWriteAccess(path))
+			cemuLog_log(LogType::Force, "Failed to write to {}", path.generic_string());
+	}
+
 	s_filename = s_full_path.filename();
 
-	g_config.SetFilename(GetPath("settings.xml").generic_wstring());
+	g_config.SetFilename(GetPath(GetConfigPath(), "settings.xml").generic_wstring());
```
goeiecool9999, line 40:
```suggestion
	g_config.SetFilename(GetConfigPath("settings.xml").generic_wstring());
```

SSimco: Thanks! I missed that.

goeiecool9999 (approved):
LGTM!

src/Cafe/GraphicPack/GraphicPack2.cpp
```diff
@@ -63,7 +63,7 @@ void GraphicPack2::LoadGraphicPack(fs::path graphicPackPath)
 void GraphicPack2::LoadAll()
 {
 	std::error_code ec;
-	fs::path basePath = ActiveSettings::GetPath("graphicPacks");
+	fs::path basePath = ActiveSettings::GetUserDataPath() / "graphicPacks";
```
goeiecool9999, line 66:
```suggestion
	fs::path basePath = ActiveSettings::GetUserDataPath("graphicPacks");
```

src/main.cpp
```diff
@@ -242,8 +242,6 @@ void unitTests()
 
 int mainEmulatorHLE()
 {
-	if (!TestWriteAccess(ActiveSettings::GetPath()))
-		wxMessageBox("Cemu doesn't have write access to it's own directory.\nPlease move it to a different location or run Cemu as administrator!", "Warning", wxOK|wxICON_ERROR); // todo - different error messages per OS
```
goeiecool9999, line 246: Should this check be reinserted somewhere else?

src/gui/CemuApp.cpp
SSimco: I'm not sure if it really matters, but I've seen apps on Linux use either `App` or `app` as the config folder name.
I just put `cemu` as a placeholder since wxWidgets sets the app name as the executable name.

src/main.cpp
SSimco: It tests write access in ActiveSettings::LoadOnce for all paths (except data_path, since it should only read from it), although it doesn't display a message but it writes in the log file.

goeiecool9999: I think it would be a good idea to show the message box popup in addition to logging.

SSimco: Now it should also display a message box saying which folder failed the write access test.

goeiecool9999:
There's another issue. The settings GUI is unchanged when portable mode is off so it still has the option for you to select your MLC path location which does nothing when you use these changes.

goeiecool9999:
Another question that I have is regarding the Data path. What's the value of `prefix` in wxWidgets? Running cemu with these changes results in the Data path being `/home/<myuser>/build/Cemu/share/cemu` Which doesn't exist at the moment. I guess getting the files in the right location would be the responsibility of a package maintainer.

SSimco:
> Another question that I have is regarding the Data path. What's the value of `prefix` in wxWidgets? Running cemu with these changes results in the Data path being `/home/<myuser>/build/Cemu/share/cemu` Which doesn't exist at the moment. I guess getting the files in the right location would be the responsibility of a package maintainer.

What wxWidgets sets the `prefix` value to depends on where the executable is. If it's in `/usr/bin/cemu` then the prefix is `/usr`, if it's in `/home/<myuser>/build/Cemu/bin/cemu` then the prefix is `/home/<myuser>/build/Cemu`. Then the data path is just the `prefix/share/cemu`. 
You can override the prefix (only on Unix) by setting the `WX_APPNAME_DATA_DIR` to the directory to use (where `APPNAME` is the upper-cased value of `GetAppName()`).

src/config/ActiveSettings.cpp
```diff
@@ -222,6 +247,6 @@ fs::path ActiveSettings::GetMlcPath()
 
 fs::path ActiveSettings::GetDefaultMLCPath()
 {
-	return GetPath("mlc01");
+	return GetPath(GetUserDataPath(), "mlc01");
```
goeiecool9999, line 250:
```suggestion
	return GetUserDataPath("mlc01");
```

SSimco:
Closed by mistake.

Exzap:
Looks good! But I do think that we should use `Cemu` over `cemu` for the the directory paths. At least that way it matches the official capitalization that we also use for the titlebar, executable name and most other places.

SSimco:
I will open another pull request since I can't edit this anymore after I deleted the branch.

PR #319, opened 2022-09-30
bugfix: title manager doesn't update when clearing MLC path in settings.

goeiecool9999:
- Update title manager when you clear the mlc path in settings and show a message when it's not accessible.
- Make sure TestWriteAccess never throws an exception
- Rewrote the infinite loop in SelectMLCPath using a goto which I know might be controversial. It can be changed back if needed but I think the code makes more sense like this.

### DIFF
--- a/src/gui/CemuApp.cpp
+++ b/src/gui/CemuApp.cpp
@@ -338,6 +338,21 @@ void CemuApp::CreateDefaultFiles(bool first_start)
 }
 
 
+bool CemuApp::TrySelectMLCPath(const std::wstring& path)
+{
+	if (!TestWriteAccess(fs::path{ path }))
+		return false;
+
+	GetConfig().SetMLCPath(path);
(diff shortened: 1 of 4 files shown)


Exzap:
> Rewrote the infinite loop in SelectMLCPath using a goto which I know might be controversial. It can be changed back if needed but I think the code makes more sense like this.

It's literally the only `goto` in the entire project outside of external dependencies. It's just not worth it to break the existing convention for whatever minor subjective advantage `goto` would have here.

Other than that, this PR looks good.

goeiecool9999:
I've reverted the rewrite of the infinite for loop and changed the wording of the new message boxes to reuse the same prompt as the one that already existed. It does split up the translation into multiple strings though.

Exzap:
I believe it's better to have the whole message be one string as then anyone doing translation will have more context and freedom for editing, even if it means a bit of redundancy.

PR #321, opened 2022-10-02
Use correct Texture clear in LatteDraw_handleSpecialState8_clearAsDepth

Tillsunset:
Fixes Yoshi's Woolly World

### DIFF
--- a/src/Cafe/HW/Latte/Renderer/OpenGL/OpenGLRendererCore.cpp
+++ b/src/Cafe/HW/Latte/Renderer/OpenGL/OpenGLRendererCore.cpp
@@ -580,7 +580,10 @@ void LatteDraw_handleSpecialState8_clearAsDepth()
 		}
 		else
 		{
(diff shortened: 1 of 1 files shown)


Exzap:
This ignores the clear color and breaks the operation for color textures.
There is more research required for this function so I dont know what the expected behavior should be for depth textures, but on the original hardware it's meant for fast color texture clears via hardware accelerated depth clears.

I recommend using this solution instead:
```
if (!view->baseTexture->isDepth)
   g_renderer->texture_clearColorSlice(view->baseTexture, sliceIndex + view->firstSlice, mipIndex + view->firstMip, clearColor[0], clearColor[1], clearColor[2], clearColor[3]);
```
It avoids the error and the behavior should be the same because on non-MoltenVk the incorrect depth-as-color clear is just ignored anyway.

Tillsunset:
Thanks for the recommendation

PR #328, opened 2022-10-03
ih264 per-function target attribute on clang and GCC

emiyl:
Continuation of #152.


emiyl:
Just curious - #152 targets sse4.1, however the functions says "sse42". Should I be targeting sse4.2, or is sse4.1 fine?

emiyl:
As per Exzap's comment on #152, I'm putting sse42.

PR #330, opened 2022-10-04
Add check for alternative glm target

SSimco:


CMakeLists.txt
```diff
@@ -100,6 +100,10 @@ if (NOT TARGET glslang::SPIRV AND TARGET SPIRV)
 	add_library(glslang::SPIRV ALIAS SPIRV)
 endif()
 
+if (NOT TARGET glm::glm AND TARGET glm)
+	add_library(glm::glm ALIAS glm)
+endif()
```
Tachi107, line 105: Why is this needed? Upstream glm only defines `glm::glm`, in both the [latest release](https://github.com/g-truc/glm/blob/0.9.9.8/cmake/glm/glmConfig.cmake#L16) and in the [master branch](https://github.com/g-truc/glm/blob/cc98465e3508535ba8c7f6208df934c156a018dc/CMakeLists.txt#L26)

PR #331, opened 2022-10-04
dedicated decoder for R4G4 and R4G4B4A4 to R8G8B8A8

emiyl:
This _should_ decode `R4_G4_UNORM` and `R4_G4_B4_A4_UNORM` to `VK_FORMAT_R8G8B8A8_UNORM`. This is my first time writing something like this so I'm not entirely sure it's accurate, and I don't actually know where these textures are visible so I can't see them either. I did a few tests and it _seems_ to be okay though?

This is useful on macOS, where MoltenVK does not support the two texture formats.

### DIFF
--- a/src/Cafe/HW/Latte/Core/LatteTextureLoader.h
+++ b/src/Cafe/HW/Latte/Core/LatteTextureLoader.h
@@ -556,7 +556,7 @@ public:
 	}
 };
 
-class TextureDecoder_R4_G4_UNORM_toRGBA4444 : public TextureDecoder, public SingletonClass<TextureDecoder_R4_G4_UNORM_toRGBA4444>
+class TextureDecoder_R4_G4_UNORM_To_RGBA4 : public TextureDecoder, public SingletonClass<TextureDecoder_R4_G4_UNORM_To_RGBA4>
 {
 public:
 	sint32 getBytesPerTexel(LatteTextureLoaderCtx* textureLoader) override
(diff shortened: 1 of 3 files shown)


src/Cafe/HW/Latte/Core/LatteTextureLoader.h
```diff
@@ -632,6 +632,50 @@ class TextureDecoder_R4_G4_UNORM_toRGBA4444_vk : public TextureDecoder, public S
...
+
+				uint8 red4 = (v0 >> 4) & 0xF;
+				uint8 green4 = (v0 & 0xF);
+
+				red4 = (red4 << 4) | red4;
+				green4 = (green4 << 4) | green4;
+
+				*(uint8*)(outputData + pixelOffset + 0) = red4;
+				*(uint8*)(outputData + pixelOffset + 1) = green4;
+			}
```
Exzap, line 662: You should add:
```
*(uint8*)(outputData + pixelOffset + 2) = 0;
*(uint8*)(outputData + pixelOffset + 3) = 0;
```
otherwise the extra channels have undefined values.

emiyl: shouldn't `pixelOffset + 3` be 255?

emiyl:
image

R4G4 decoder seems to be working wonderfully

emiyl:
Since the R4G4 decoder works, I think it's safe to assume that the R4G4B4A4 decoder works too, since they're nearly identical.

emiyl:
Class names were a lil inconsistent and it was bugging me so @Tillsunset and I agreed on a consistent way to name them.

```
TextureDecoder_R5_G5_B5_A1_UNORM_swappedRB_To_RGBA8
TextureDecoder_A1_B5_G5_R5_UNORM_To_RGBA8
TextureDecoder_R5_G6_B5_UNORM_To_RGBA8
TextureDecoder_R4_G4_UNORM_To_RGBA8
TextureDecoder_R4_G4_UNORM_To_RGBA4_vk```

PR #345, opened 2022-10-08
Input panel color fix

Tillsunset:
before and after
Screen Shot 2022-10-08 at 2 20 32 AM
Screen Shot 2022-10-08 at 2 20 58 AM


Crementif:
Before:
image
After:
image

This might just be an unintentional change but it seems like they default to a more disabled-field looking color on light themes at least. You'll have to see how well it works but a solution could be to just find two constant colors for either dark/black modes or light modes. Just try and see what works.

The comments also make this code feel copy-pasted from the game list at the moment.

Exzap:
Instead of duplicating the theme checking code (the part that determines the value of isDarkTheme) it would probably be better to move it into an utility function in `wxHelpers.h`

Amnesia1000:
Not only the error is in the input panel, also in the title manager, showing an item with black background and white text and the next with white background and white text.

PR #350, opened 2022-10-09
memory optimization, using posix_memalign

Tillsunset:
replace aligned_alloc with posix_memalign, no more allocating padding bytes.

Really sorry for the multiple PRs, I found this as the last one merged.


Tillsunset:
still works just as fine Screen Shot 2022-10-09 at 2 11 16 AM

Exzap:
According to the man page for posix_memalign, `alignment` has to be a multiple of `sizeof(void*)`.
Adding something like `alignment = std::max<WORD32>(alignment, sizeof(void*));` should suffice.

Tillsunset:
added all the safety we ever wish for

PR #354, opened 2022-10-10
Add bin/ to .gitignore

S41L0R:
Ok this one is really minor, I admit it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@
 
 build/
 out/
+bin/
 .cache/
 bin/Cemu_*
 bin/Cemu_*.exe


Crementif:
We intentionally keep the /bin folder in since it contains the initial binary files like gameprofiles, resources and shader cache. See here: https://github.com/cemu-project/Cemu/tree/main/bin.

emiyl:
Could I suggest:

```
bin/*
!bin/gameProfiles
!bin/resources
!bin/shaderCache
```

Exzap:
Excluding `bin/gameProfiles` as a whole doesn't work because it contains both static `gameProfiles/default/` and dynamic files.
Anyway the gitignore has pretty much everything inside `/bin` already covered, I think it's only missing the screenshot folder.

PR #355, opened 2022-10-10
Follow imgui recommendation and streamline build dependencies

Exzap:
The imgui repository doesn't come with a build script since it's meant to be embedded into projects by just copying the source files. Package managers work around this by inventing their own build script, which is suboptimal because everyone adds their own spin on things. But more critically, for flatpak (and appimage?) this means we would also have to create a custom `CMakeLists.txt` or find some other roundabout way to build imgui.
So instead, this PR changes it so we follow their recommendation. imgui is built as part of ImguiImpl and we source their files via a git submodule. Doing it this way means no workaround is required for any of the platforms we currently target, including the future flatpak.

This PR also simplifies the dependencies of `RendererShaderVk.cpp` by getting rid of the check for legacy versions of glslang, which we don't support anymore anyway.


abouvier:
Aren't you also supposed to use the files in [imgui/backends](https://github.com/ocornut/imgui/tree/master/backends) instead of vendor them in [Cemu/src/imgui](https://github.com/cemu-project/Cemu/tree/main/src/imgui)?

Exzap:
Generally yes, but we use customized backends.

PR #356, opened 2022-10-11
Add support for non portable mode

SSimco:
Now the app name is `Cemu` instead of `cemu`.


kturques:
Don't know if this is the right place to comment, but "gameProfiles" are not getting saved. This should go in "$XDG_CONFIG_HOME/Cemu/gameProfiles"

SSimco:
> Don't know if this is the right place to comment, but "gameProfiles" are not getting saved. This should go in "$XDG_CONFIG_HOME/Cemu/gameProfiles"

It wasn't saving because it was not creating a `gameProfiles` folder. Also, it was trying to save `gameProfiles` to `$XDG_DATA_HOME/Cemu/gameProfiles`.

I changed it now so it saves to `$XDG_CONFIG_HOME/Cemu/gameProfiles` and it creates the folder if it doesn't exist.

Exzap:
Looks good now. Thanks!

emiyl:
In the original PR (#316), it says the macOS config and user data paths should be at `~/Library/Application Support/Cemu`, however this is not the case. In this PR, config files are stored in `~/.config/Cemu` and user data files are stored `/.local/share/Cemu` on macOS, same as Linux. Is this intentional?

emiyl:
> No, it's not intentional. Maybe BOOST_OS_LINUX is defined somehow and it's using the paths for Linux?

Yes, there should be an extra check for BOOST_OS_MACOS when setting paths.

Tatsh:
This does not cover everything that #130 was covering since when I start Cemu, it immediately shows this and hangs:

image

With #130 this was not the case and the Cemu binary could live anywhere as long as system data paths were known. There should not be a need to have a wrapper that would set up the correct working directory before execution, assuming that is the workaround here.

PR #364, opened 2022-10-13
Create Cemu .app bundle for macOS

emiyl:
This PR:

- Creates a .dmg file when building releases, containing a Cemu.app bundle and an alias to `/Applications`
- Disables portable mode when compiling for macOS, using `~/Application Support/Cemu` instead, and `Cemu.app/Contents/SharedSupport` for default game profiles and langauge files
- Copies the MoltenVK dylib from the macOS build VM to the bundle, so it can be ran without installing anything from brew

This brings the macOS releases more in line with most other macOS apps.


Tillsunset:
just something of note, when using an app bundle and looking at relative paths (such as for controller profiles). Cemu might look relative to the executable not relative to the app bundle itself. The app bundle is closer to a folder than anything else. So you might wanna look [here](https://github.com/cemu-project/Cemu/blob/main/src/config/ActiveSettings.cpp#L22) and iteratively get the parent path until we get the folder where everything is at.

OatmealDome:
While this does work, it's *very* sub-optimal because it relies on GitHub Actions to build the bundle. In addition, the built bundle does not include an `Info.plist` file. macOS expects all apps to have this file and for it to be populated with metadata.

Please consider using [CMake build options](https://cmake.org/cmake/help/latest/prop_tgt/MACOSX_BUNDLE.html) instead.

Also, I don't believe you need to include `MoltenVKShaderConverter`.

Tillsunset:
Something along the lines of [this](https://github.com/Tillsunset/Cemu/blob/f89104173fc7ca5bc34c7666e2668abc546a5b17/src/CMakeLists.txt#L82-L90), I don't have the iconset but it should be easy to create one based on a high quality base icon. 

I also fixed working with resource files when using app bundles with [this](https://github.com/Tillsunset/Cemu/blob/f89104173fc7ca5bc34c7666e2668abc546a5b17/src/gui/CemuApp.cpp#L80-L82).

I'm not quite sure how this should all be put together, but this are small changes you can take as reference

emiyl:
Thank you for the feedback, I'm now using the CMake build system to generate the app bundle and I created this icon, which I think fits in pretty well with macOS:

cemu

emiyl:
While this doesn't help with controller profiles, I think that the language resources should be moved to be inside the bundle, not in `~/Library/Application Support`.

DTibor1986:
hi Emiyl, thanks for your effort/work for bundling Cemu into Mac app.
Probably you are aware of it, but a brief feedback: I still manually need to chmod Cemu app in Contents in order to run it. (at least on my Studio)
Other than that I can not give the app permission to be able to read the folder where I store my game dump, the app is simply not showing up in data security settings where I can give the folder permission to it. (this could be because due to my lack of knowledge :-))

emiyl:
>  I still manually need to chmod Cemu app in Contents in order to run it.

This should now be fixed.

> Other than that I can not give the app permission to be able to read the folder where I store my game dump

Hmm that's odd. I get the prompt when opening to access my folders when opening the app, but not when I directly open the binary from the MacOS folder. Are you opening the .app bundle?

DTibor1986:
Accessing the games folder is OK now, just needed to update my keys.txt in Application Support library. (sorry my mistake)
As for the chmod, still necessary even with the latest build, compiled less than an hour ago.

AndreaCicca:
It would be great to have the .app file into a .dmg like any other third party app for MacOS

emiyl:
For releases, it will be packaged in .dmg

emiyl:
> As for the chmod, still necessary even with the latest build, compiled less than an hour ago.

I see the issue :( Unfortunately I'm not sure. It's definitely running chmod on the binary before it's released as an artifact, but it doesn't seem to carry across.

.github/workflows/build.yml
```diff
@@ -213,19 +215,32 @@ jobs:
...
-      run: chmod a+x bin/Cemu_release && mv bin/Cemu_release bin/Cemu
+      run: |
+        mkdir bin/Cemu
+        mv bin/Cemu_release.app bin/Cemu/Cemu.app
+        mv bin/Cemu/Cemu.app/Contents/MacOS/Cemu_release bin/Cemu/Cemu.app/Contents/MacOS/Cemu
+        sed -i '' 's/Cemu_release/Cemu/g' bin/Cemu/Cemu.app/Contents/Info.plist
+        mkdir bin/Cemu/Cemu.app/Contents/SharedSupport
+        for folder in gameProfiles resources
+        do
+          cp -R ./bin/${folder} ./bin/Cemu/Cemu.app/Contents/SharedSupport
```
OatmealDome, line 234: Looks like this part is still dependent on GitHub Actions. It would be best if this (and the stuff with MoltenVK below) are done with CMake so that builds can be produced locally.

Tillsunset:
It seems when using zip, file permissions are not preserved. Some other format must be used. [Link](https://stackoverflow.com/questions/2846387/file-permissions-changed-between-os-x-and-windows)

.github/workflows/build.yml
emiyl: Done :) not too familiar with cmake but it seems to be working

galad87:
To create a zip that keeps all the right macOS bits you can use:

ditto -c -k --sequesterRsrc --keepParent

emiyl:
I've worked around the permissions issue by just packaging the artifact in a dmg, and then mounting the dmg in the release actions.

DTibor1986:
Hi Emiyl, 

the dmg in latest nightly is working great, thank you!! :-)
If it not a big request could you please make an app bundle with moltenVK 1.2 included?

emiyl:
moltenvk 1.2 should already be included

Issue #297, opened 2022-09-25
No localization on macOS


FRtranslator:
Localization is not loaded on macOS, Cemu build 2.0-4: the UI sticks to English even if you choose "French" for instance.

goeiecool9999:
I want to add to that that it also doesn't work on Arch Linux and it crashes when you select a different language. If anyone knows if it works on Ubuntu that would be helpful.

jamescastells:
May be related, but if I select "English" as a UI language I get this pop-up every time I start Cemu:

<img width="281" alt="image" src="https://user-images.githubusercontent.com/5696029/192436578-306ec8bd-1816-426d-b4d4-e6d9de41b659.png">

PR #366, opened 2022-10-13
Fix Cannot set locale to ""

### LINKED ISSUE #349: [macOS] Cannot set locale to language "".
Everytime I open Cemu on macOS, I get the same message: "Cemu Warning. Cannot set locale to language ""."

image

My general settings contain both "Default" and "English". If I select English, upon restarting it reverts to "Default" while keeping the error message.

image

Using the latest version on M1 Macbook pro.

Tillsunset:
Add en resource and change language selection for macOS user that don't have US as their system region. Fixes #349 and #297


Fs00:
@Tillsunset Did the addition of the English .mo file alone fix any issues? Because as far as I know it's common practice not to include the .mo catalog for the source language.

Tillsunset:
No, it was only when I included the English files and changed the behavior in the backup case to use it specifically.

PR #370, opened 2022-10-14
Linux/MacOS: Greatly improve performance

Exzap:
This PR replaces `std::unordered_set` in LatteBufferCache with a custom high-performance sparse bitset implementation.

Benchmark from my Steam Deck. Before:
image

With these new changes I get a 60% improvement in FPS:
image

The FPS boost will not be the same for each game as different workloads have different bottlenecks, but at least some improvement should be seen across the board. It also does not increase performance on Windows noticeably because MSVC's implementation of unordered_set already behaved in just the way we needed for good performance.

In depth technical explanation:
Seems like `std::unordered_set` cannot be blindly trusted for performance critical code. The performance characteristics differ based on which STL is used as each one comes with a different underlying algorithm. `libstdc++` seems to implement unordered_set as flat hashtable and consequentially .clear() is generally pretty expensive since it invokes `memset` on the entire table, while on MSVC it's much lower cost due to being implemented as a bucketed hashtable. Our render backend calls `.clear()` frequently so we rely on this operation being very fast.

This isn't to say that one implementation is superior over the other. It's just that we have to be careful with certain assumptions about performance for STL containers since they may not hold true on all platforms. When in doubt and high performance is of critical importance, it's better to use boost (same implementation for all platforms) or robin_hood, or any of the many other optimized container implementations.

Anyway, after benchmarking `std::unordered_set` vs `boost::unordered_set` vs `robin_hood::unordered_set` I came to the conclusion that none of them are perfectly suitable for simulating a sparse bitset of invalidated pages. I ended up writing a custom SparseBitset container that performs a bit better than the previous best case of using MSVC's `std::unordered_set`

### DIFF
--- a/src/Cafe/HW/Latte/Core/LatteBufferCache.cpp
+++ b/src/Cafe/HW/Latte/Core/LatteBufferCache.cpp
@@ -1005,8 +1005,67 @@ void LatteBufferCache_getStats(uint32& heapSize, uint32& allocationSize, uint32&
 }
 
 FSpinlock g_spinlockDCFlushQueue;
-std::unordered_set<uint32>* g_DCFlushQueue = new std::unordered_set<uint32>(); // queued pages
-std::unordered_set<uint32>* g_DCFlushQueueAlternate = new std::unordered_set<uint32>();
+
+class SparseBitset
+{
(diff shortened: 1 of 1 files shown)


amayra:
how about  libc++ ?

PR #372, opened 2022-10-14
Added Boot Image for Gamepad

MythicalPlayz:


src/Cafe/HW/Latte/Core/LatteShaderCache.cpp
```diff
@@ -52,6 +52,11 @@ struct
 	sint32 pipelineFileCount;
 }g_shaderCacheLoaderState;
 
+struct
+{
+	ImTextureID textureId;
+}d_shaderCacheLoaderState;
```
Exzap, line 58: Creating a new structure seems unnecessary. You can just add `textureId` to `g_shaderCacheLoaderState`.
I also suggest renaming both the TV and GamePad `textureId` to something more descriptive like `bootTVTexId` and `bootDRCTexId` which avoids the name collision and matches the filenames.

src/Cafe/HW/Latte/Core/LatteShaderCache.cpp
```diff
@@ -251,7 +256,27 @@ void LatteShaderCache_load()
 
 		fsc_close(fscfile);
 	}
+	//get game background loading image for DRC
+	TGAFILE file2{};
+	d_shaderCacheLoaderState.textureId = nullptr;
+
+	std::string drcTexPath = fmt::format("{}/meta/bootDRCTex.tga", CafeSystem::GetMlcStoragePath(CafeSystem::GetForegroundTitleId()));
+	sint32 status2;
+	auto fscfile2 = fsc_open(drcTexPath.c_str(), FSC_ACCESS_FLAG::OPEN_FILE | FSC_ACCESS_FLAG::READ_PERMISSION, &status2);
```
Exzap, line 265: I suggest reusing `status` and `fscFile` instead of defining a new set of variables. Although this is subjective and not necessary.
At the very least, you should avoid numbered variable names (status, status2) and use more descriptive names.

src/Cafe/HW/Latte/Core/LatteShaderCache.cpp
```diff
@@ -484,6 +541,36 @@ void LatteShaderCache_ShowProgress(const std::function <bool(void)>& loadUpdateF
...
+				g_renderer->BeginFrame(false);
+				if (g_renderer->ImguiBegin(false)) {
+				ImGui::SetNextWindowPos({ 0,0 }, ImGuiCond_Always);
+				ImGui::SetNextWindowSize(io.DisplaySize, ImGuiCond_Always);
+				ImGui::PushStyleVar(ImGuiStyleVar_WindowBorderSize, 0);
+				ImGui::PushStyleVar(ImGuiStyleVar_WindowPadding, { 0,0 });
+				if (ImGui::Begin("Background texture2", nullptr, kPopupFlags))
+				{
+					if (d_shaderCacheLoaderState.textureId)
+				{
```
Exzap, line 553: The `if` and it's `{` should be at the same indentation. But also the entire block is too far indented.
I noticed that in general there is a bit of weird formatting. Try to match the coding style of the surrounding code :)

Exzap (changes requested):
Looks good, but some formatting changes are required

src/Cafe/HW/Latte/Core/LatteShaderCache.cpp
MythicalPlayz: Probably Just VS code acting weird.

src/Cafe/HW/Latte/Core/LatteShaderCache.cpp
MythicalPlayz: I did actually did try this but it would result in an error (cant remember if it was a runtime or a compiler error).

src/Cafe/HW/Latte/Core/LatteShaderCache.cpp
MythicalPlayz: Anyways I tried fixing it as best as I can

Exzap:
Looks good, thanks! Indentation is still off in some places but clang-format will take care of that in the future.

PR #375, opened 2022-10-14
Fixes the CRC errors

Tillsunset:
The logs saying CRC mismatch on macOS is fixed. ~~I searched for the "\_\_BYTE\_ORDER" macro [here](https://gcc.gnu.org/onlinedocs/cpp/Common-Predefined-Macros.html) but found "\_\_BYTE_ORDER\_\_"~~
Now using [std::endian](https://en.cppreference.com/w/cpp/types/endian) and handle mixed endian

### DIFF
--- a/src/util/crypto/crc32.cpp
+++ b/src/util/crypto/crc32.cpp
@@ -322,7 +322,7 @@ unsigned int crc32_calc_slice_by_8(unsigned int previousCrc32, const void* data,
 	// process eight bytes at once (Slicing-by-8)
 	while (length >= 8)
 	{
-		if (std::endian::native == std::endian::big){
+		if constexpr (std::endian::native == std::endian::big){
 			uint32_t one = *current++ ^ swap(crc);
 			uint32_t two = *current++;
 			crc = Crc32Lookup[0][two & 0xFF] ^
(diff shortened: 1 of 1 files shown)


Tillsunset:
[before](https://github.com/cemu-project/Cemu/files/9790725/log.txt)
[after](https://github.com/cemu-project/Cemu/files/9790719/log.txt)

Crementif:
Could also use https://en.cppreference.com/w/cpp/types/endian (see example at the bottom) to detect endianness. Not sure if there's any downsides over using a header-defined macro, but detecting endianness seems to vary quite a lot so I think it'd work better.

Tillsunset:
Thanks for the advice, now using std function

src/util/crypto/crc32.cpp
```diff
@@ -322,29 +322,34 @@ unsigned int crc32_calc_slice_by_8(unsigned int previousCrc32, const void* data,
...
-		crc = Crc32Lookup[0][(two >> 24) & 0xFF] ^
-			Crc32Lookup[1][(two >> 16) & 0xFF] ^
-			Crc32Lookup[2][(two >> 8) & 0xFF] ^
-			Crc32Lookup[3][two & 0xFF] ^
-			Crc32Lookup[4][(one >> 24) & 0xFF] ^
-			Crc32Lookup[5][(one >> 16) & 0xFF] ^
-			Crc32Lookup[6][(one >> 8) & 0xFF] ^
-			Crc32Lookup[7][one & 0xFF];
-#endif
+		if (std::endian::native == std::endian::big){
```
Exzap, line 325:
```suggestion
		if constexpr (std::endian::native == std::endian::big) {
```
Without constexpr, the compiler may not optimize the conditional away during compile time. Also needs to be added to the `else if` below, but not the else.

Crementif:
I did note some egyptian/weirdly formatted braces, but seems good to me!

PR #376, opened 2022-10-15
Fix crashes from libpng version conflicts

goeiecool9999:
As discussed in discord statically linking libpng exports function symbols for the static version which are also wrongly used by dynamic libraries that dynamically link to the system version of libpng which can lead to crashes.


goeiecool9999:
I've added the options to the linux specific section because I don't know if macOS builds have the same issue. I've also added some comments explaining why the options are there.

goeiecool9999:
Using both a statically linked version of libpng for wxwidgets and the system version for everything else is obviously less than ideal but I don't see a straightforward way of resolving that.

Tachi107:
> Using both a statically linked version of libpng for wxwidgets and the system version for everything else is obviously less than ideal but I don't see a straightforward way of resolving that.

Could you please be more clear? You mean that when using `-rdynamic` you can't dynamically link to some libraries?

goeiecool9999:
> > Using both a statically linked version of libpng for wxwidgets and the system version for everything else is obviously less than ideal but I don't see a straightforward way of resolving that.
> 
> Could you please be more clear? You mean that when using `-rdynamic` you can't dynamically link to some libraries?

Cemu statically links a libpng version for wxWidgets from vcpkg, however some other dynamic dependency ignores the static library that's present in the cemu executable and instead loads the system version. Even without rdynamic cemu exports the same symbols that a libpng .so file normally would so that leads to problems when the dynamically loaded version is a different version from the one that is statically linked with cemu.
One crash I encountered was where a function pointer in a libpng object originating from the dynamic version of libpng was set to a function from the statically linked version of libpng.
It's my hope that by removing the static library symbols from the export table the symbols will be separated and any potential conflicts are resolved.

Tachi107:
> Cemu statically links a libpng version for wxWidgets from vcpkg, however some other dynamic dependencies ignore the static library that's present in the cemu executable and instead loads the system version.

How can this happen? If, say, libpng is statically linked, the executable simply cannot link to the shared version on the system.

    $ cc -o pngtest /usr/share/doc/libpng-dev/examples/pngtest.c -lpng -lm -lz
    $ ldd pngtest
    	linux-vdso.so.1 (0x00007ffcee78e000)
    	libpng16.so.16 => /lib/x86_64-linux-gnu/libpng16.so.16 (0x00007f79c1b42000)
    	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f79c1800000)
    	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f79c1b25000)
    	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f79c1a46000)
    	/lib64/ld-linux-x86-64.so.2 (0x00007f79c1b99000)

    $ cc -o pngtest /usr/share/doc/libpng-dev/examples/pngtest.c /usr/lib/x86_64-linux-gnu/libpng.a -lm -lz
    $ ldd pngtest
    	linux-vdso.so.1 (0x00007ffe9f39c000)
    	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f3b28fcb000)
    	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f3b28fae000)
    	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f3b28c00000)
    	/lib64/ld-linux-x86-64.so.2 (0x00007f3b290f6000)

(note the missing `libpng16.so.16` in the second example)

goeiecool9999:
> How can this happen?

Try also linking/using libraries like libcairo, libfreetype, or libgdk_pixbuf, all of those shared libraries try to link to libpng and they will try to load the dynamic version of libpng.

goeiecool9999:
I'm pretty certain that I'm right about what's happening.
image
In this image of the stracktrace of my crash you can clearly see libpng functions in both a dynamically loaded library and cemu itself.
removing the symbols like this PR does fixes the crash.

Tachi107:
Makes sense. Wouldn't it be easier to always link to the shared 
version, then? libpng is available pretty much everywhere.

-- 

goeiecool9999:
I agree but I think there's a reason that they use a specific libpng version for statically linked libraries like wxwidgets. That's why instead of going that route I fixed the symbol conflicts instead.

goeiecool9999:
I think they statically link libpng because the vcpkg version of for wxwidgets does it that way. I guess that makes sense for windows if libcairo, libfreetype, and libgdk_pixbuf aren't used on that platform.

Exzap:
Superseded by #381
Thanks for looking into this! As mentioned on Discord, going the route of always dynamically linking libpng even with vcpkg seems to cause the least headache.

Issue #311, opened 2022-09-28
Compiles, but doesn't start on Ubuntu 22.10


jagiella:
I successfully compiled cemu with clang-14 following the respective steps of the documentation https://github.com/cemu-project/Cemu/blob/main/BUILD.md

```bash
sudo apt install -y git curl cmake ninja-build nasm libgtk-3-dev libsecret-1-dev libgcrypt20-dev libsystemd-dev freeglut3-dev libpulse-dev
sudo apt install -y clang
git clone --recursive https://github.com/cemu-project/Cemu
cd Cemu
cmake -S . -B build -DCMAKE_BUILD_TYPE=release -DCMAKE_C_COMPILER=/usr/bin/clang -DCMAKE_CXX_COMPILER=/usr/bin/clang++ -G Ninja
cd build
ninja -j 2
cd ..
```

I get some strange linker warnings:
```bash
[480/480] Linking CXX executable /home/nickj/Dokumente/GitHub/Cemu/bin/Cemu_release
/usr/bin/ld: Warnung: Typ des Symbols 'glDeleteProgramPipelines' geändert von 2 nach 1 in /tmp/lto-llvm-3f1852.o
/usr/bin/ld: Warnung: Typ des Symbols 'glDisable' geändert von 2 nach 1 in /tmp/lto-llvm-3f1852.o
/usr/bin/ld: Warnung: Typ des Symbols 'glDepthMask' geändert von 2 nach 1 in /tmp/lto-llvm-3f1852.o
/usr/bin/ld: Warnung: Typ des Symbols 'glPrimitiveRestartIndex' geändert von 2 nach 1 in /tmp/lto-llvm-3f1852.o
/usr/bin/ld: Warnung: Typ des Symbols 'glStencilMaskSeparate' geändert von 2 nach 1 in /tmp/lto-llvm-3f1852.o
/usr/bin/ld: Warnung: Typ des Symbols 'glClipControl' geändert von 2 nach 1 in /tmp/lto-llvm-3f1852.o
/usr/bin/ld: Warnung: Typ des Symbols 'glDisablei' geändert von 2 nach 1 in /tmp/lto-llvm-3f1852.o
/usr/bin/ld: Warnung: Typ des Symbols 'glLogicOp' geändert von 2 nach 1 in /tmp/lto-llvm-3f1852.o
[...]

PR #378, opened 2022-10-15
Fix building with system libraries

vieiraa:
This fixes building Cemu with vcpkg disabled.
Tested only in Arch Linux. 
Workaround for #311.


Tachi107:
Arch Linux is currently shipping a broken glm package, the issue should be fixed there instead.

vieiraa:
The SIGSEGV is not caused by glm.
Apparently it is caused by the libpng package in vcpkg (at least it crashes there, according to gdb).
I removed glm from the link libraries because it is not necessary to link against it.

Tachi107:
> I removed glm from the link libraries because it is not necessary to link against it.

Yes, if the headers are not installed in the default include path then linking to the imported target is required

Exzap:
Closing this since it's a workaround and not an actual fix for the problem. I also suspect this may have been fixed by #381 and #382

PR #382, opened 2022-10-17
Remove libpng dependency from wxwidgets for linux platform.

goeiecool9999:
After #381 cemu was still linking libpng statically but this fixes it.


Exzap:
Ah right. I missed this. Thanks!

PR #383, opened 2022-10-17
Fixed Discord Rich Presence not working on games that are on MLC

MythicalPlayz:
This PR makes it so Discord Rich Presence loads the name of the title instead of title ID and also loads it with your Cemu Language  Preference.


Exzap:
thanks!

PR #388, opened 2022-10-18
Auto update game title name when Language is changed and fallback to English name if not available

MythicalPlayz:
This PR updates the short name of the titles in the title list when the user switches language without requiring a restart to update.

This PR also updates the Window/Discord Name by having it fallback to the English Strings if there is not an available string in the selected language.

### DIFF
--- a/src/Cafe/CafeSystem.cpp
+++ b/src/Cafe/CafeSystem.cpp
@@ -68,6 +68,7 @@
 
 std::string _pathToExecutable;
 std::string _pathToBaseExecutable;
+std::string applicationName;
 
 RPLModule* applicationRPX = nullptr;
 uint32 currentBaseApplicationHash = 0;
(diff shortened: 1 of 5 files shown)


src/Cafe/CafeSystem.cpp
```diff
@@ -68,6 +68,7 @@
 
 std::string _pathToExecutable;
 std::string _pathToBaseExecutable;
+std::string applicationName;
```
Exzap, line 71: This should not be a global. It's used only in `GetForegroundTitleName()` and it should be defined there.

src/Cafe/TitleList/TitleInfo.cpp
```diff
@@ -9,6 +9,8 @@
 #include <zarchive/zarchivereader.h>
 #include "config/ActiveSettings.h"
 
+std::string titleNameCfgLanguage;
```
Exzap, line 12: This should also not be a global.

PR #399, opened 2022-10-21
Vulkan: Refactor swapchain code.

goeiecool9999:
Changes in this PR:
* Change all variants of mainWindow (isMainWindow, main_window, mainwindow) to mainWindow
* Change all variants of chainInfo (swapInfo, swap_info, swap_chain_info, swapchain_info) to chainInfo
* Eliminate all capitalisations of swapchain that aren't Swapchain or swapchain (treated as a single word in Vulkan)
* Renamed SwapChainInfo struct SwapchainInfoVk and put it in separate files
* Separate out a lot of swapchain related code into SwapchainInfoVk
* Put ternary operator to choose between SwapchainInfoVk structures in dedicated function
* Remove a bunch of branches for picking between the two SwapchainInfoVk instances
* Call existing WaitDeviceIdle member function instead of vkWaitDeviceIdle

PR #401, opened 2022-10-21
GeneralSettings: Update vsync after settings close

goeiecool9999:


Exzap:
The behavior is a bit inconsistent right now, at least on Vulkan. When it recreates the swapchain it will use whatever vsync mode is currently in the config (see `VulkanRenderer::ChooseSwapPresentMode`). Your message says a restart is required, but the new vsync will apply when the window is resized or anything else causes the swapchain to be recreated.

Imho we should just get rid of `renderer->EnableVSync` and have the render backends compare the active vsync mode against the one stored in the config once per frame. If it differs then switch to new vsync. At least in theory this should avoid any multi-threading issues.

goeiecool9999:
I've changed it according to your suggestions. Will probably cause a merge conflict with #399 but this should be far easier to review. Verified that it works for vulkan but I haven't tested OpenGL as I would have to use windows for that.

Exzap:
Looks good. No issues found while testing.

PR #402, opened 2022-10-22
add check for "." in path and unit test.

### LINKED ISSUE #373: [FSC] FSCPath refactor broke the ability to boot WUA copies of Donkey Kong Country: Tropical Freeze
Opening this to track the issue reported on discord where https://github.com/cemu-project/Cemu/commit/b8462cec8b8b6d7adb75d621268ddb187b358f25 introduced an issue causing the banana to peel into infinity and not make it in game.

Game region is EUR and update is packed into the WUA.

Tillsunset:
Fixes #373

### DIFF
--- a/bin/gameProfiles/default/0005000010101a00.ini
+++ b/bin/gameProfiles/default/0005000010101a00.ini
@@ -1,7 +1,1 @@
 # LEGO City Undercover (USA)
-
-[General]
-loadSharedLibraries = false
-
-[Graphics]
-GPUBufferCacheAccuracy = 0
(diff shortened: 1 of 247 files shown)


Exzap:
Good catch.
I think the robustness of this can be improved, it breaks some theoretically well formed paths:
`/vol/content/..` Should produce the nodes "vol", "content" and "..". With this PR the trailing ".." becomes "."
`/vol/content/filename.` For this example it alters the filename by cutting off the dot
It also currently doesn't handle something like:
`/vol/content/./file.txt`

Exzap:
Looks good now. Thanks!

PR #404, opened 2022-10-23
Fix incorrect title ID (00050000-1011000? -> 00050000-10111000)

kescherCode:
For some reason, a 1 was omitted in the title ID for Ben 10 OMNIVERSE™.

I noticed this when taking a directory listing:

the title ID's game profile .ini noticeably has a character less than all others in a directory listing


Exzap:
Thanks!

PR #408, opened 2022-10-24
DSU: fix build error when enable DEBUG_DSU_CLIENT

schspa:
We need this ec variable for debug message print

Signed-off-by: Schspa Shi <schspa@gmail.com>


Exzap:
Thanks!

PR #410, opened 2022-10-25
DSUMessages: fix bad crc calculate

schspa:
We need to set m_crc32 to zero before calculate crc value. m_crc32 is zero when the peer calculate crc value.

Signed-off-by: Schspa Shi <schspa@gmail.com>


Exzap:
Ah I overlooked the crc field hashing itself when I changed this code. Thanks!

PR #413, opened 2022-10-25
FSA: Add missing path translations

GaryOderNichts:
Translates paths for remove, makeDir and rename.
This fixes issues where the working directory was not respected for those commands.


Exzap:
Thanks!

PR #414, opened 2022-10-25
OpenGL: Bind back buffer before clearing (fixes #222)

### LINKED ISSUE #222: [Linux] [OpenGL] Maximizing the Cemu window makes the whole screen dark
In full-screen mode this does not happen, the game is displayed correctly.

gurrgur:
This fixes #222.
The back buffer is not attached when clearing, causing the previously bound framebuffer object to be cleared instead. The solution is to bind the back buffer before clearing. In the vulkan renderer it is done in the same way.

### DIFF
--- a/src/Cafe/HW/Latte/Renderer/OpenGL/OpenGLRenderer.cpp
+++ b/src/Cafe/HW/Latte/Renderer/OpenGL/OpenGLRenderer.cpp
@@ -548,6 +548,9 @@ void OpenGLRenderer::DrawBackbufferQuad(LatteTextureView* texView, RendererOutpu
 	renderstate_resetDepthControl();
 	attributeStream_reset();
 
+	// bind back buffer
+	rendertarget_bindFramebufferObject(nullptr);
+
 	if (clearBackground)
 	{
 		int windowWidth, windowHeight;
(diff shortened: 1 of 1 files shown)


gurrgur:
Marking this as WIP for now since this likely needs some testing. It would be great if someone could test this on windows with nvidia or intel drivers as I only have access to AMD hardware right now.

Squall-Leonhart:
Window resize issue confirmed fixed on nvidia.

gurrgur:
Thanks for testing! I also tested with my AMD card on linux and windows 10 and both seem to work fine so this should be ready now.

Exzap:
Looks good to me. Thanks!

Romano44:
I have still this issue on 1.27.1 on Windows 8.1 with latest GTX1060 drivers as follows: 
When I resize the window or go fullscreen, screen goes black. This only occur with OpenGL, not Vulkan.
Also, it happen with keeping aspect ratio option ON. If I turn it off to just "stretch", issue disappear. However aspect ratio is then broken on my 4:3. In any case it help I have a scaling set to be done on display in the drivers CP.

Exzap:
1.27.1 is from June 2022. Get the latest release from <https://github.com/cemu-project/Cemu/releases/tag/v2.0-32> which includes the fix from this pull request.

PR #416, opened 2022-10-26
Fixes some compiler warnings

Tillsunset:
replaces volatile with std::atomic
always true assert
bitwise operation on enum with equivalent builtin enum
printf format with proper unsigned long long
counter operation that always returns false

### DIFF
--- a/src/Cafe/HW/Latte/Core/Latte.h
+++ b/src/Cafe/HW/Latte/Core/Latte.h
@@ -64,8 +64,8 @@ struct LatteGPUState_t
 		{
 			bool isEnabled;
 			MPTR physPtr;
-			volatile uint32 flipRequestCount;
-			volatile uint32 flipExecuteCount;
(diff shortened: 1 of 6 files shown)


Tillsunset:
https://github.com/cemu-project/Cemu/blob/c9dd8ef46de459639c7d2554f1d340b67b2d67ff/src/gui/CemuUpdateWindow.cpp#L512-L517
counter is always used as 1, 11, 21...
I'm not 100% sure if that is intended over being used as 0, 10, 20...

Exzap:
> counter is always used as 1, 11, 21...
I'm not 100% sure if that is intended over being used as 0, 10, 20...
or even 10, 20, 30..

Since this only has negligible side effects we can leave it like that for now. Personally I'm not a fan of altering `counter` in the middle of the conditional so my proposed solution would be to increment `counter` after the if.

PR #418, opened 2022-10-26
Update macOS instructions to use LLVM 15

UltraHDR:
This compiles fine for me


UltraHDR:
Thanks @Tillsunset for the detailed explanation. You know better than me and I'll close this and wait untl December.

PR #425, opened 2022-10-28
macOS: Set values for empty keys in Info.plist

### LINKED ISSUE #422: [macOS] "Process name" in Activity monitor appears blank
Instead of appearing as "Cemu" in Activity monitor, its process name appears blank:

image

OatmealDome:
A lot of these keys were mistakenly left blank, so I've filled them out (and removed a few deprecated ones).

Fixes #422.

image


src/CMakeLists.txt
```diff
@@ -75,6 +75,13 @@ if (MACOS_BUNDLE)
 	set(RESOURCE_FILES "${CMAKE_SOURCE_DIR}/src/resource/cemu.icns")
 	target_sources(CemuBin PRIVATE "${RESOURCE_FILES}")
 
+	set(MACOSX_BUNDLE_ICON_FILE "cemu.icns")
+	set(MACOSX_BUNDLE_GUI_IDENTIFIER "info.cemu.Cemu")
+	set(MACOSX_BUNDLE_BUNDLE_NAME "Cemu")
+	set(MACOSX_BUNDLE_SHORT_VERSION_STRING ${CMAKE_PROJECT_VERSION})
+	set(MACOSX_BUNDLE_BUNDLE_VERSION ${CMAKE_PROJECT_VERSION})
+	string(TIMESTAMP MACOSX_BUNDLE_COPYRIGHT "Copyright © %Y Cemu Project")
```
Tachi107, line 83: I think it'd be better to replace this dynamic year with a static one, as it makes bit-by-bit build reproducibility a bit harder. Also, copyright applies from when you created (or significantly updated) a certain work, not when it was transformed in binary form.

OatmealDome: Yeah, that makes sense. Fixed.

emiyl (approved):

PR #431, opened 2022-10-29
Vulkan: Experimental fix for self-dependency barriers during render pass. (performance testing needed)

### LINKED ISSUE #267: [Linux] [Radv vulkan driver] [BOTW] Tall grass like graphical bugs
Zelda: Breath of the wild has wierd texture issues with vulkan. It includes long grass like textues and some missing ones. 

This is inside that grass texture

same bug on building

inside the above glitch

some more 

Oh and a opengl screenshot, it doesn't occur in opengl

It is like this for me since forever really, the frequency of these tall grass has decreased though, in v1.26, it was way worse. I suspected bad dumb, but it works correctly with opengl, so i guess we can rule that out.

Anyway, I couldn't find anyone else with same problem that is also one of the reasons why i never posted about it.

I am using linux build with commit hash 
12b6830546e28e6a697c5dbcbc3aa24882ec7fb3

Artix linux 
Mesa version 22.1.7
vulkan version 1.3.211
opengl version 4.6
vulkan driver radv
cpu ryzen 3 3300u
gpu vega  6
ram 6 gb
vram 2gb

Oh and about log file, can someone give directions how to obtain them, i Have no idea how to.

goeiecool9999:
This change ends the current render pass when cemu detects a self-dependency within it. This fixes #267.
However the performance implications of separating the draw calls into multiple render passes and the broader barriers are unknown.
~~My test in breath of the wild showed an acceptable reduction in worst %.1 framerate from 69.7fps to 56.3fps. with 97th percentile being down only 2 fps from 102 to 100.~~ (Bad testing methodology)
Different GPUs/OS's/drivers may get different outcomes so this change requires performance testing.
**Accurate barriers must be enabled for this to work**

### DIFF
--- a/src/Cafe/HW/Latte/Renderer/Vulkan/VulkanRendererCore.cpp
+++ b/src/Cafe/HW/Latte/Renderer/Vulkan/VulkanRendererCore.cpp
@@ -1214,6 +1214,9 @@ void VulkanRenderer::draw_setRenderPass()
 	if (m_state.descriptorSetsChanged)
 		sync_inputTexturesChanged();
 	
+	// assume that FBO changed, update self-dependency state
+	m_state.hasRenderSelfDependency = fboVk->CheckForCollision(m_state.activeVertexDS, m_state.activeGeometryDS, m_state.activePixelDS);
+
 	sync_RenderPassLoadTextures(fboVk);
 
 	if (m_featureControl.deviceExtensions.dynamic_rendering)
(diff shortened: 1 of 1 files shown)


Exzap:
Performance seems unchanged on Windows Nvidia and Intel Xe drivers, at least nothing that would be discernible beyond random noise. If the impact is this low on three completely different implementations then this should be safe to merge, at least performance-wise.

PR #433, opened 2022-10-30
Vulkan: Fix freeze when closing pad view.

goeiecool9999:
Explanation:
The pad VulkanCanvas has a Semaphore wrapped in a shared_ptr.
Added a boolean variable to VulkanRenderer that indicates if the pad window should be stop being updated. It also has a weak_ptr for a semaphore.
Swapchain operations only take place if IsSwapchainInfoValid returns true. We add the boolean to the check for the pad window.
When the pad canvas's destructor is called it calls a function in VulkanRenderer that sets the boolean that stops swapchain use, and as an argument it passes it's shared_ptr to the semaphore. The function saves this semaphore as a weak reference. 
When the function returns the canvas destructor awaits the semaphore.
When the rendering thread reaches the next IsSwapchainInfoValid for the pad window it checks to see if the weak reference is valid and if it is it notifies the semaphore. This causes the Canvas destructor to finish, invalidating the swapchain. Then it returns false because of the boolean flag which means the invalid swapchain is never used again.
The next time the pad view is opened it resets the flag and recreates the swapchain like the first Initialize call.

### DIFF
--- a/src/Cafe/HW/Latte/Renderer/Vulkan/VulkanRenderer.cpp
+++ b/src/Cafe/HW/Latte/Renderer/Vulkan/VulkanRenderer.cpp
@@ -658,18 +658,18 @@ void VulkanRenderer::Initialize(const Vector2i& size, bool isMainWindow)
 	}
 	else
 	{
-		m_padCanvasDestroying = false;
-		m_usingPadFrameSurface.acquire();
+		m_padCloseReadySemaphore = nullptr;
+		m_forceStopPadUse = false;
+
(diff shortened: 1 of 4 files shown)


goeiecool9999:
I just realised the whole passing pointers back and forth is just a way to avoid removing the constness of IsSwapchainInfoValid. It would be bad to remove it's constness but it would get rid of the odd passing around of pointers cause the semaphore could just be a member of VulkanRenderer. @Exzap Already pointed out that having side effects to a IsSomething function is bad and that's effectively what I've done now with this ownership workaround.

Issue #278, opened 2022-09-20
Cemu crashes when closing seperate GamePad view


Her0brineCreeper:
Whenever the seperate "GamePad view" window is closed while a game is running, the Cemu emulator crashes immediately. It doesn't matter wether the X (close) button of the window or the Options->Seperate GamePad view button is used. When no game is running, the window can be opened and closed without any problems. The crash occures with every game I tested.
No crash dump is being created eventhough the option is set in the settings. Maybe the attached [log.txt](https://github.com/cemu-project/Cemu/files/9608526/log.txt) file contains any useful information.

System details:
Cemu 2.0-2 (Experimental)
Windows 10 21H2 64-Bit
AMD Ryzen 5 2600X
16 GB RAM
NVIDIA GeForce RTX 2060 SUPER

Squall-Leonhart:
This is vulkan specific, and only affects nvidia drivers.

goeiecool9999:
Can you try again with the latest version see if the crash still occurs?

Her0brineCreeper:
The issue is fixed for me with the latest version of Cemu (2.0-15). Closing the windows works fine now according to my tests.
Thank you very much for your work in fixing this issue :D

SessionLoops:
@goeiecool9999 Hi, could you share what might cause the crash? I'm a developer and facing same issues too. Thanks in advance!

Squall-Leonhart:
review the referenced changes

PR #435, opened 2022-10-30
Vulkan: Fix freeze when closing pad view

goeiecool9999:
I think this way is a lot cleaner than #433
likely fix for #278


MaranBr:
This fixes the problem for me.

PR #439, opened 2022-10-31
Linux/macOS: Fix various colour values for dark themes.

### LINKED ISSUE #296: [Linux] Text doesn't display correctly on 'About Cemu' window while using system's Dark theme
- Current Version: `2.0-4` (compiled manually, commit [101ff77](https://github.com/cemu-project/Cemu/commit/101ff7783ea8d39bbf0d8f2ec7c3f65168870d65), issue present since `2.0`) 
- issue is GUI related, can't find a relevant log to show

Not really much to say here, when opening the `About Cemu` tab, it's missing the majority of text, only displaying links for the libraries/utils. No version number, no compile date, no Patreon supporter list, mostly blank
Screenshot_20220924_162852

This however isn't an issue when running the Windows version via Wine
Screenshot_20220924_163535

goeiecool9999:
This change fixes up some colour codes for dark themes on linux in places that others missed.
I moved the existing code for calculating accent colours to the wxwidgets helpers file as a function so the title manager list can use the same code.
besides that I removed the white background colour in the about dialog and made the input UI also use the background colour for the buttons.
I haven't checked how this affects the look of other platforms
Fixes #296

Screenshot_2022-10-31_00-56-03
Screenshot_2022-10-31_00-54-44
Screenshot_2022-10-31_00-55-38
Screenshot_2022-10-31_00-52-55
Screenshot_2022-10-31_00-55-47
Screenshot_2022-10-31_00-52-40


Exzap:
Quick test on Windows. Before:
image
This PR:
image

The background being slightly darker gives it a bit of a greyed-out look.

Title Manager looks the same. About window looks ok:
image

goeiecool9999:
Is this an acceptable to way to solve it for now?

goeiecool9999:
image
We could keep it in the oven a little longer. It's not perfect yet.

goeiecool9999:
This is what it looks like using the CalculateAccentColour function for the selected items.
image
There is a third colour for input options from a different controller than the one currently selected which is still too light though.

goeiecool9999:
I made it so the background colour doesn't change when an input is selected. And use the accent colour to show inputs mapped to a different controller than the one selected. Looks okay on regular adwaita:
image
And with adwaita-dark
image

Theming is a big controversy in the GNOME/GTK space. I think we should check if this looks good on Windows/Ubuntu/macOS and then call it a day for now.

jamescastells:
How can I test this on macOS? I can't seem to find a binary to test for this specific commit.

goeiecool9999:
> How can I test this on macOS? I can't seem to find a binary to test for this specific commit.

You can download the artifact by clicking on the green circle next to the latest revision and then go to the summary. It's at the bottom of the page. I also couldn't figure it out the first time I tried. 😉

goeiecool9999:
Another thing that is TODO is switching between themes while Cemu is running. Currently that's not possible because it never updates the colours.

jamescastells:
> You can download the artifact by clicking on the green checkmark next to the latest revision and then go to the summary. It's at the bottom of the page. I also couldn't figure it out the first time I tried. 😉

Do you mean this page? Because there are no artifacts below.

image

jamescastells:
I see it ok on macOS. The only complain I'd have is that when selected a text-box, it's not obvious in the UI. For example, I'm selecting the "A" textbox here, but it's not notable. 

image

goeiecool9999:
> The only complain I'd have is that when selected a text-box, it's not obvious in the UI.

That's what I was afraid of.

jamescastells:
Yes, I agree. Even with that "complaint" on macOS I think this should be merged and released ASAP.

PR #442, opened 2022-11-01
Non-windows: Link fontawesome into executable.

goeiecool9999:
On linux we don't have access to resource files so we have to resort to converting files to C source code.
For fontawesome this would result in a 1MB C source file from a 160KB .ttf file. Clearly not acceptable.
This is the next best thing. Short assembly file that exposes symbols to the .ttf file data.
I'm creating this pull request to trigger CI to check if mac supports it too. ~If not I'll make it linux only.~
Confirmed working on macOS and Linux.

### DIFF
--- a/src/imgui/imgui_extension.cpp
+++ b/src/imgui/imgui_extension.cpp
@@ -43,6 +43,10 @@ void ImRotateEnd(float rad, ImVec2 center)
 uint8* extractCafeDefaultFont(sint32* size);
 sint32 g_font_size = 0;
 uint8* g_font_data = nullptr;
+#if !BOOST_OS_WINDOWS
+extern int const g_fontawesome_size;
+extern char const g_fontawesome_data[];
+#endif
 std::unordered_map<int, ImFont*> g_imgui_fonts;
 std::stack<int> g_font_requests;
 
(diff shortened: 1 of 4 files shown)


goeiecool9999:
Confirmed working on macOS and Linux.

Tachi107:
Could you please add an option to avoid embedding them? This would be undesirable in distribution packages and possibly also Flatpak

goeiecool9999:
> Could you please add an option to avoid embedding them? This would be undesirable in distribution packages and possibly also Flatpak
> 
> Edit: also, I think it would be necessary to see if FontAwesome's license allows embedding, and under which conditions.

Cemu already embeds it in the windows .exe so I assume they sorted out the licensing. Initially I was gonna try to figure out fontconfig but then I thought it might cause issues if the system font is different from what cemu is expecting so I went the route of embedding it. Speaking for myself here I wouldn't mind a package including a 100kb font in an executable. Are you anticipating issues for packages/flatpack other than size and data duplication?

Tachi107:
Yes, distributions generally try to avoid embedding stuff into the 
executable for various reasons/policies, and as far as I understand 
flatpak is similar in this regard.

Exzap:
> Could you please add an option to avoid embedding them? This would be undesirable in distribution packages and possibly also Flatpak

We don't officially support distro packaging so this isn't something we should take into account. Embedding works fine on every platform and there are no licensing issues with fontawesome. Flatpak also has no issues with this

Tachi107:
> We don't officially support distro packaging so this isn't something we should take into account

Cemu also didn't have to account for vcpkg-less builds, but in the end that turned out to be needed for Flatpak 🙃

Of course you don't _need_ to add the option, it's just nice if you do.

goeiecool9999:
Just in case you still want to use the patch to use fontconfig
https://github.com/goeiecool9999/Cemu/commit/cbbef62314fc573813790ac184c768003de730d0
This is an updated version I messed something up in the cmake file

Tachi107:
Thanks! It will surely be useful to somebody :)

PR #452, opened 2022-11-03
Appimage build

qurious-pixel:
Add AppImage to artifacts and release builds


.github/workflows/build_check.yml
```diff
@@ -11,7 +11,7 @@ on:
     paths-ignore:
       - "*.md"
     branches:
-      - main
+      #- main
```
qurious-pixel, line 14: to-do: uncomment this line.

Exzap:
Solid work!
I spent a bit of time testing this and it works well. There are some minor UI issues that I didn't have in native builds, but we'll iron these out as we go.

Thanks!

PR #453, opened 2022-11-04
Linux: use Feral GameMode if available

### LINKED ISSUE #259: [Feature Request] preventing linux to go in to standby while playing
i have archlinux with x11 gnome and cemu built from source running mariokart 8 and i am using a dualshock ps4 controller over bluetooth but after the set ammount of inactivity time the screen turns off and cemu doesnt prevent that

lisa-wolfgang:
Resolves #259.

[Feral GameMode](https://github.com/FeralInteractive/gamemode) is a service that, when enabled for a given process, requests that certain game-enhancing optimizations are applied to the OS. One of these optimizations includes inhibiting the screen saver.

This PR adds a setting (under General) to enable this service while a game is running in Cemu. It begins during game initialization (i.e. when `log.txt` is first written to.) The required package (`gamemode`, or `gamemode-git` on AUR) has also been added to the build instructions and workflows.


goeiecool9999:
To comply with the licensing terms the license should be added to the about window

Exzap:
I suspect not every users wants it turned on all the time since it increases power consumption. There should be an option to enable/disable it.

goeiecool9999:
Is it necessary to add the packages to the installation instructions/workflows? If you include the header file shouldn't compilation work just fine without installing game mode?

src/main.cpp
```diff
@@ -262,6 +266,14 @@ int mainEmulatorHLE()
...
 	// log Cemu startup info
 	infoLog_cemuStartup();
+	// enable gamemode (Linux-only library for standby prevention and some OS optimization)
+	#if BOOST_OS_LINUX
+		if (GetConfig().gamemode && gamemode_request_start() < 0)
+			cemuLog_force("Failed to enable gamemode, it may not be installed");
+			// TODO: disable gamemode when only the game, not Cemu, is closed (a feature not yet implemented)
+			// currently unnecessary because this will happen automatically when Cemu closes
+			// gamemode_request_end();
+	#endif
```
goeiecool9999, line 276: mainEmulatorHLE is a function that runs when cemu starts and in the settings you only start gamemode if a game is running. Should probably move this somewhere like MainWindow::FileLoad so it only runs when a game is started.

src/main.cpp
```diff
@@ -262,6 +266,14 @@ int mainEmulatorHLE()
 	PPCCore_init();
 	// log Cemu startup info
 	infoLog_cemuStartup();
+	// enable gamemode (Linux-only library for standby prevention and some OS optimization)
+	#if BOOST_OS_LINUX
+		if (GetConfig().gamemode && gamemode_request_start() < 0)
```
goeiecool9999, line 271: Short-circuiting controls whether gamemode is started or not. I think it's worth the extra lines to make it clearer what is happening.

goeiecool9999:
Just a suggestion but maybe you could also disable the config option when you detect that gamemode isn't installed with something like when ```gamemode_query_status``` returns -1 and show a tooltip explaining feral gamemode needs to be installed.

lisa-wolfgang:
> Is it necessary to add the packages to the installation instructions/workflows? If you include the header file shouldn't compilation work just fine without installing game mode?

Theoretically, yes, but the idea was to have screen saver inhibition included by default for convenience.

goeiecool9999:
> > Is it necessary to add the packages to the installation instructions/workflows? If you include the header file shouldn't compilation work just fine without installing game mode?
> 
> Theoretically, yes, but the idea was to have screen saver inhibition included by default for convenience.

Ideally the build instructions are geared more towards developers. Most users should download a build that cemu provides. Since gamemode isn't in the downloads the users who want to use this feature would still need to install gamemode themselves (that's also why I suggested showing the user visually that an additional install is necessary in the settings menu). Having an optional thing that is only required at runtime in the `BUILD.md` instructions seems confusing to me. Also if the github actions can pass without installing gamemode we should _definitely_ remove it from that. It would only serve to slightly slow down build times.

As an aside:
Does anyone know if this would work without modifying the appimage?

lisa-wolfgang:
> Most users should download a build that cemu provides. Since gamemode isn't in the downloads the users who want to use this feature would still need to install gamemode themselves

True. If possible, I think we should look into a way to have at least screen saver inhibition not require an external installation.

This is how GameMode itself does it: https://github.com/FeralInteractive/gamemode/blob/b11d2912e280acb87d9ad114d6c7cd8846c4ef02/daemon/gamemode-dbus.c#L711
This is Dolphin's implementation: https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/UICommon/X11Utils.cpp#L48

Tachi107:
> the idea was to have screen saver inhibition included by default for convenience.

I'd say that using gamemode just for that is a bit overkill. Linux desktop environments have APIs to signal that the screen saver shouldn't be activated

Tachi107:
> > the idea was to have screen saver inhibition included by default for convenience.
> 
> I'd say that using gamemode just for that is a bit overkill. Linux desktop environments have APIs to signal that the screen saver shouldn't be activated

Edit:

> This is how GameMode itself does it

Yes, there's a cross-desktop D-Bus interface to do this (I'm sure it works on GNOME, can't say anything about others environments). Alternatively there's a Wayland protocol to do this, but it is of course Wayland-only and not supported on GNOME.

Bitwolfies:
Steam and Firefox can block the freedesktop screensaver api, though I am unsure as to how.

Screenshot_20221107_195034

lisa-wolfgang:
Closing this PR in favor of an implementation that wouldn't require an external installation. Continue discussion in https://github.com/cemu-project/Cemu/issues/259. I am not currently working on a follow-up PR.

PR #463, opened 2022-11-06
Add commit hash to build name on build check workflow

Tillsunset:
replaces 999999 in build name with the short hand commit hash of build check. Only affects pushes and PRs
Screenshot 2022-11-06 at 5 25 15 PM


Exzap:
It might be more sensible to get the git commit hash via cmake script?

Exzap:
Thanks!

CMakeLists.txt
```diff
@@ -6,7 +6,14 @@ option(MACOS_BUNDLE "The executable when built on macOS will be created as an ap
 set(EXPERIMENTAL_VERSION "" CACHE STRING "") # used by CI script to set experimental version
 
 if (EXPERIMENTAL_VERSION)
-  add_definitions(-DEMULATOR_VERSION_MINOR=${EXPERIMENTAL_VERSION})
+	add_definitions(-DEMULATOR_VERSION_MINOR=${EXPERIMENTAL_VERSION})
+	execute_process(
+			COMMAND git log --format=%h -1
```
jn64, line 11: This will get the wrong hash if user is not working on the latest commit. I believe correct way is:

```
git rev-parse --short HEAD
```

PR #470, opened 2022-11-08
ImGui: match all begin/end

lijunyu-cn:
1. As commented in imgui.h, should call a matching End() for each Begin()  regardless of its return value. Tested in BotW, this seems to improve performance very slightly.
2. Fix softkeyboard, click on background overlay will bring it to front and cover the softkeyboard.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@
 *.out
 *.app
 .vs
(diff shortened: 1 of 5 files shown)


Exzap:
LGTM. Thanks!

PR #474, opened 2022-11-10
Linux OpenGL: Fix gamepad not updating

goeiecool9999:
Apparently in order for a canvas to be able to swap buffers, on linux the context needs to be current for that canvas.


qurious-pixel:
**Master:**
gl-pad3

**PR:**
gl-pad4

PR #475, opened 2022-11-10
Linux: Save 10.7MiB of useless debug info

goeiecool9999:
The GitHub actions artifacts include 10.7MiB worth of debugging info which is most likely of no use to anyone because cemu's code itself does not include debug information.
The data consists largely of debug information about some boost code and discord-rpc.


Tachi107:
`cmake --install` has a `--strip` option. Wouldn't it be better to use that when preparing the GitHub action artifact instead? "useless" debug information might be useful sometimes.

goeiecool9999:
> `cmake --install` has a `--strip` option. Wouldn't it be better to use that when preparing the GitHub action artifact instead? "useless" debug information might be useful sometimes.

It is useless because it only covers some boost libraries and discord-rpc. So if you're gonna attach a debugger with this info you'd only benefit from it in that code. That's why I called it useless. If you have no info for the cemu code calling that stuff it'll be harder use the debugger. Either include all debug info or none at all was my thinking.

goeiecool9999:
> `cmake --install` has a `--strip` option.

Do you know if it keeps symbol information? The DWARF debug info won't be of much use but the symbols make the crash logs more informative.

PR #476, opened 2022-11-11
Disable Discord checkbox based on build option

jn64:
Disable the Discord checkbox when compiled with `-DENABLE_DISCORD_RPC=OFF`.

I'm not a programmer (but I play one on TV), so please let me know where the code should go if not here.

screenshot


jn64:
> If true the check box should not be clickable also.

I don't know what you mean.

Masamune3210:
They mean the checkbox next to the greyed out option should be disabled as well to prevent people from enabling it

Exzap:
The text being grayed out means the checkbox is disabled. Box and text are part of the same checkbox widget

Anyway LGTM. Thanks!

Masamune3210:
Aah, I see. Thanks for the info

PR #480, opened 2022-11-11
Linux/macOS: overlay CPU/memory statistics

goeiecool9999:
This is a continuation of #142.
~~It now reads a different linux file that reports memory use more accurately~~ and it cleans up some of the conditional compiling for different platforms. When #142 was opened the CMake file used a GLOB so they couldn't use logic to exclude source files. Some other files also used macros so I made those use the same cmake logic.
macOS support is also implemented

For testers: Enable the overlay with RAM usage, CPU usage, and CPU per-core usage enabled. Then load a game and check if the metrics are correct.


goeiecool9999:
Screenshot showing the overlay working on linux:
image

Tillsunset:
This currently causes a seg fault on macOS when enabling the stats overlay

goeiecool9999:
I misunderstood the parameters. It seems that processorMsgCount is the size of the object pointed to by cpuLoad. It's probably okay now.

Tillsunset:
LGTM
Screenshot 2022-11-12 at 7 06 49 AM

goeiecool9999:
Looked at some other code that uses the mac function and apparently the stackoverflow answer has some resource leaka so now I changed it to try to fix that. Your comment was after that but the github build wasn't finished yet. Did you try the latest version?

Tillsunset:
yeah built it myself

goeiecool9999:
Someone pointed out that the linux version has inaccurate FPS metrics when the overlay is enabled. This is because the code that updates statistics and timestamps is assumed to consume zero time. Reading smaps_rollup apparently consumes enough time to cause a 0.2 difference in reported fps. The kernel manual pages mentioned it was slower but I didn't expect that slowdown to be actually significant.

goeiecool9999:
To calculate how much unexpected runtime there is in milliseconds you can solve 60÷(1−x÷1000) = 60.2.
Solving gives x = 3.32226. Over 3 milliseconds just to compute an accurate metric for the amount of memory used seems quite steep. I'll revert to using statm

Exzap:
On Windows this PR causes the CPU usage to show as 50 - 100%, instead of the full 0 - 100% range. Idle cores show up as 50%.
image
Any idea why this happens?

goeiecool9999:
> Any idea why this happens?

I didn't touch the windows code from the previous PR. But I notice that compared to the old version a comment went missing with this content: ``// kernel time already includes idletime`` So I guess whoever wrote the new version must have forgotten about that.

Exzap:
That fixed it. Thanks!

tomlally:
> So I guess whoever wrote the new version must have forgotten about that.

Wasn't forgotten; I just wanted to test on windows, unfortunately circumstances interfered. Sorry about that. https://github.com/cemu-project/Cemu/pull/142#issuecomment-1285325085

PR #494, opened 2022-11-14
Fix two if statements checking range of number wrong

goeiecool9999:
These were found by enabling every compiler warning and gradually disabling the unhelpful ones.
I believe the compiler has correctly identified these mistakes and this corrects them.
image
image


Exzap:
Looks good to me. MULLO is used by DS Virtual Console JITed code, so this might fix some issues with those titles.

PR #497, opened 2022-11-16
Add cross-platform "disable screen saver" setting

### LINKED ISSUE #259: [Feature Request] preventing linux to go in to standby while playing

lisa-wolfgang:
Resolves #259.

This PR adds a "disable screen saver" setting to the General section of the general settings. Disabling the screen saver now also works on macOS and Linux, in addition to Windows.

(The following is no longer true for the latest commit) ~~The one caveat is that the approach is taken from Dolphin's source code. The specific section used is licensed under GPL v2. This means that the section may need to be rewritten. I'm not sure if the ways in which the APIs are used are general enough for this to matter, but better safe than sorry with licensing.~~


infirit:
xdg-screensaver doesn't work in both Xfce [1] and KDE plasma [2]. 

[1] https://gitlab.freedesktop.org/xdg/xdg-utils/-/merge_requests/38
[2] https://gitlab.freedesktop.org/xdg/xdg-utils/-/issues/114

src/util/ScreenSaver/ScreenSaver.h
```diff
@@ -0,0 +1,55 @@
...
+        IOPMAssertionRelease(s_power_assertion);
+        s_power_assertion = kIOPMNullAssertionID;
+      }
+    }
+#endif
+
+#if BOOST_OS_LINUX
+    char id[11];
+    char *argv[4] = {(char *)"xdg-screensaver", (char *)(inhibit ? "suspend" : "resume"), id, nullptr};
+    posix_spawnp(nullptr, "xdg-screensaver", nullptr, nullptr, argv, environ);
```
Tachi107, line 52: Calling the `xdg-screensaver` binary looks like a hack. Since you already include `sd-bus.h`, why not calling the D-Bus method directly? And by the way, as this is a really minor D-Bus usage, it might be better to use the more minimal, reference libdbus; it is more widely available (does not depend on libsystemd etc)

vieiraa:
You can use D-Bus to implement this on Linux.
Here's RPCS3's implementation (they use Qt): https://github.com/RPCS3/rpcs3/blob/04df3928667d40f17fca9d15f97858ec2e4259b0/rpcs3/display_sleep_control.cpp.
You'll have to implement it using the native library.
I find this sample: https://github.com/makercrew/dbus-sample/blob/master/dbus.cpp

infirit:
> Here's RPCS3's implementation (they use Qt): https://github.com/RPCS3/rpcs3/blob/04df3928667d40f17fca9d15f97858ec2e4259b0/rpcs3/display_sleep_control.cpp.

And this lacks at least Xfce and xscreensaver support. And if gnome-screensaver is not installed it will loose Gnome support as well.

vieiraa:
org.freedesktop.ScreenSaver should work with Xfce. 
Source: https://github.com/xfce-mirror/xfce4-session/blob/master/xfce4-session/xfce-screensaver.c

infirit:
> org.freedesktop.ScreenSaver should work with Xfce. Source: https://github.com/xfce-mirror/xfce4-session/blob/master/xfce4-session/xfce-screensaver.c

That's only for inhibiting other people's screensaver. XFCE doesn't implement `org.freedesktop.ScreenSaver` itself, it uses `org.xfce.ScreenSaver`. And tomorrow people will ask you to inhibit on `org.cinnamon.ScreenSaver` and a month later `com.deepin.ScreenSaver` and it will go on and on, it'll never end.

src/util/ScreenSaver/ScreenSaver.h
infirit: also note that xdg-screensaver doesn't work on wayland.

j8r: It will also not work in Flatpak as-is without https://github.com/Unrud/xdg-screensaver-shim.

lisa-wolfgang:
The `org.freedesktop.ScreenSaver` API is now called instead of `xdg-screensaver`. Relevant errors are printed in `log.txt` if Cemu cannot access this API.

As mentioned earlier, this is the standard interface for screen saver inhibition. As such, this should provide out-of-the-box functionality for the most users possible.

src/util/ScreenSaver/ScreenSaver.h
```diff
@@ -0,0 +1,126 @@
+#include "Cemu/Logging/CemuLogging.h"
+
+#if BOOST_OS_MACOS
+#include <IOKit/pwr_mgt/IOPMLib.h>
+#endif
+
+#if BOOST_OS_LINUX
+#include <spawn.h>
```
Tachi107, line 8: It seems that including [`spawn.h`](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/spawn.h.html) is unneeded.

src/util/ScreenSaver/ScreenSaver.h
```diff
@@ -0,0 +1,126 @@
...
+#ifdef USE_ELOGIND
+#include <elogind/sd-bus.h>
+#include <elogind/sd-daemon.h>
+#else
+#include <systemd/sd-bus.h>
+#include <systemd/sd-daemon.h>
+#endif
```
Tachi107, line 15: Instead of using sd-bus, I think it'd be better to use plain [libdbus](https://dbus.freedesktop.org/doc/api/html/). It is a bit less developer-friendly, but it is more widely available (not Linux-only) and not part of the systemd project, which some users may want to avoid for whatever reason; here we're also only making a couple of dbus calls, so it shouldn't be that big of a pain.

src/util/ScreenSaver/ScreenSaver.h
```diff
@@ -0,0 +1,126 @@
...
+    // Adapted from https://github.com/FeralInteractive/gamemode/blob/b11d2912e280acb87d9ad114d6c7cd8846c4ef02/daemon/gamemode-dbus.c#L711, available under the BSD 3-Clause license
+    static unsigned int screensaver_inhibit_cookie = 0;
+
+    const char *service = "org.freedesktop.ScreenSaver";
+    const char *object_path = "/org/freedesktop/ScreenSaver";
+    const char *interface = "org.freedesktop.ScreenSaver";
+    const char *function = inhibit ? "Inhibit" : "UnInhibit";
+
+    sd_bus_message *msg = NULL;
+    sd_bus *bus_local = NULL;
```
Tachi107, line 61: Cemu's codebase generally puts `*` near the type, not the variable name.

```suggestion
    const char* service = "org.freedesktop.ScreenSaver";
    const char* object_path = "/org/freedesktop/ScreenSaver";
    const char* interface = "org.freedesktop.ScreenSaver";
    const char* function = inhibit ? "Inhibit" : "UnInhibit";

    sd_bus_message* msg = NULL;
    sd_bus* bus_local = NULL;
```

src/util/ScreenSaver/ScreenSaver.h
```diff
@@ -0,0 +1,126 @@
...
+    sd_bus_error err;
+    memset(&err, 0, sizeof(sd_bus_error));
```
Tachi107, line 63: This can be more cleanly zero-initialized with:

```suggestion
    sd_bus_error err = {0};
```

But I believe that `err` doesn't need to be initialized anyway, as it is written by the `sd_bus_call_method()` calls.

src/util/ScreenSaver/ScreenSaver.h
```diff
@@ -0,0 +1,126 @@
...
+
+    int result = -1;
+
+    // Open the user bus
+    int ret = sd_bus_open_user(&bus_local);
+    if (ret < 0)
+    {
+      cemuLog_force("Could not disable screen saver (user bus error)");
```
Tachi107, line 71: Indentation is a bit messed up. Before opening the `if` body you use four spaces, but here you use two.

Cemu seems to generally use tabs.

src/util/ScreenSaver/ScreenSaver.h
```diff
@@ -0,0 +1,126 @@
...
+    }
+    else
+    {
+      result = 0;
+    }
+
+    sd_bus_unref(bus_local);
+#endif
+  };
+};
```
Tachi107, line 126: Missing newline at the end of the file.

Tachi107 (changes requested):
Thanks for your work on this! Using `org.freedesktop.ScreenSaver` instead of calling a random binary is certainly cleaner :)

Still, I have a few suggestions below.

src/util/ScreenSaver/ScreenSaver.h
lisa-wolfgang: I will try my best, but DBus is outside of my area of expertise, which is why I adapted implementations from other software.

src/util/ScreenSaver/ScreenSaver.h
lisa-wolfgang: Unless I'm missing something, the indentation seems to be correct. The four spaces are from being nested inside both a class and a method.

src/util/ScreenSaver/ScreenSaver.h
lisa-wolfgang: The one thing I'm confused about is how to connect to the user bus using libdbus like sd-bus does. I haven't been able to find documentation about this.

vieiraa: I think libdbus calls it DBUS_BUS_SESSION. Found an example program that uses libdbus:
https://github.com/wware/stuff/blob/master/dbus-example/dbus-example.c
You'll problably need to do something similar to the query function in this example.

lisa-wolfgang:
This PR was initially working for me on Ubuntu, but at some unknown point in time after testing the latest commit, I began getting `Could not disable screen saver (org.freedesktop.ScreenSaver not available)` errors in `log.txt`, which means that Cemu was able to connect to the user bus, but using the `org.freedesktop.ScreenSaver` interface failed in some way. Not sure why this is happening.

lisa-wolfgang:
@Tachi107 What other means do you have in mind?

Tachi107:
You could use something like busctl, dbus-send or gdbus from the command line.

Shamus397:
Just thought I'd chime in here with a small suggestion: libSDL2 inhibits screensavers in a cross-platform way and does disabling by default since v2.0.4 (since they figure most developers use it for games or media players); as such, you might want to look at their code to see how they do it.

 For Linux, the relevant files (from v2.26.2 sources) are src/video/x11/SDL_x11events.c (for X) and src/video/wayland/SDL_waylandwindow.c (for Wayland).  It might be worthwhile looking at how they handle MacOS and Windows, too, FWIW (it may just be identical to what you do in your patch).

Shamus397:
Now that I think about it, libSDL2 is a dependency of Cemu, and so disabling the screensaver may be as simple as a call to `SDL_DisableScreenSaver()` at the appropriate place.  : )

Shamus397:
Having played with this a bit, it looks like it works just fine—at least under Linux.  In order for it to work, though, you have to init the SDL video subsystem like so (I stuck it in `CemuApp::OnInit()`):

`SDL_InitSubSystem(SDL_INIT_VIDEO);`

However, since screen saving is disabled by default this way, if you want to leave it up to the user whether or not to disable the screen saver you'll have to follow up with:

`SDL_EnableScreenSaver();`

No need to reinvent the wheel, just leverage what's available.  : )

lisa-wolfgang:
Thanks for your research on this! Using SDL's built-in screen saver functionality is certainly much simpler than trying to figure out D-Bus, and hopefully it supports more desktop environments without the caveat of numerous namespaces.

I'll keep the Windows and macOS implementations until it can be tested that those work properly.

lisa-wolfgang:
At least on my end, it seems that neither `SDL_InitSubSystem(SDL_INIT_VIDEO)` nor `SDL_DisableScreenSaver()` can manage to disable the screen saver. `SDL_InitSubSystem(SDL_INIT_VIDEO)` runs without error, but checking `SDL_IsScreenSaverEnabled()` after running `SDL_DisableScreenSaver()` returns `SDL_TRUE` (meaning the screen saver has not been disabled.)

Shamus397:
It might be a bug in the version of libSDL2 that is linked in vcproj (2.0.22 I believe) that causes `SDL_IsScreenSaverEnabled()` to report `SDL_TRUE` after calling `SDL_DisableScreenSaver()`.  Have you tried running a game in Cemu after putting `SDL_InitSubSystem(SDL_INIT_VIDEO)` in `CemuApp::OnInit()`?  That alone is sufficient on my system to engage in the default behavior of disabling the screen saver here.  Did you check the return value of `SDL_InitSubSystem(SDL_INIT_VIDEO)` to see if it returned an error (anything < 0)?  If it does show an error, you can find what it is by calling `SDL_Error()` (it will return a string with the error).

Does any SDL2 based game/media player disable the screen saver on your platform? 

I probably should mention that I'm running KDE/Plasma on Linux.  : )

lisa-wolfgang:
Yes, I built Cemu from the latest commit and ran a game, letting it sit on the title screen until the screen saver kicked in. When I checked `log.txt`, there was no error for `SDL_InitSubSystem`.

Shamus397:
Here's a little test program to see if libSDL2 works at all on your system:

`#include <SDL2/SDL.h>`
`#include <stdio.h>`

`int main(int argc, char * argv[])`
`{`
`        int response = SDL_InitSubSystem(SDL_INIT_VIDEO);`

`        if (response < 0)`
`        {`
`                printf("Could not initialize SDL video: %s\n", SDL_GetError());`
`                return -1;`
`        }`

`        SDL_bool b = SDL_IsScreenSaverEnabled();`

`        printf("SDL reports screensaver is %s...\n", (b == SDL_TRUE ? "ACTIVE" : "inactive"));`
`        printf("Press [ENTER] key to quit...\n");`

`        char c = fgetc(stdin);`

`        SDL_Quit();`
`        return 0;`
`}`

You can compile this with (assuming the above code is put into "testsdl.cpp"):

`g++ testsdl.cpp -o testsdl -lstdc++ -lSDL2`

Assuming it tells you that the screensaver is inactive, you can let it run until the screensaver is supposed to kick in to see if it works for you (i.e., don't touch the mouse and/or keyboard until the screensaver is supposed to kick in).  :-)

lisa-wolfgang:
Running the test program gives me the following output:
```
SDL reports screensaver is inactive...
Press [ENTER] key to quit...
```
I let it run until the screen saver was supposed to kick in (modified to 1 minute for testing purposes) and the screen saver inhibition seemed to work as intended.

Shamus397:
OK, now that we've established that it works on your system, maybe where the `SDL_InitSubSystem()` call is done is sensitive to where it's called from.  I put it directly after:

`m_mainFrame = new MainWindow();`

in `CemuApp.cpp`; maybe give that a try?  :-)

lisa-wolfgang:
The issue was that the subsystem was never being initialized due to a mistakenly inverted status check. Everything is now working as expected on my end! Thanks for working through this with me.

src/util/ScreenSaver/ScreenSaver.h
```diff
@@ -0,0 +1,73 @@
...
+#include <SDL2/SDL.h>
+#endif
+
+class ScreenSaver
+{
+public:
+  static void SetInhibit(bool inhibit)
+  {
+    // Adapted from Dolphin's implementation: https://github.com/dolphin-emu/dolphin/blob/e3e6c3dfa41d377520f74ec2488fc1f7b6c05be3/Source/Core/UICommon/UICommon.cpp#L454
+#if BOOST_OS_WINDOWS
```
Tachi107, line 17: Can't the SDL screensaver inhibitor be used on all platforms? It would simplify code quite a bit.

Shamus397: It absolutely can be used on all platforms; after all, that's SDL's ultimate purpose.  :-)  Would still need testing on the other platforms though, just to be sure.  ;-)

lisa-wolfgang: I was initially hesitant about removing the platform-specific implementations, but now that I think about it again, I'm not exactly sure why.

src/gui/GeneralSettings2.cpp
```diff
@@ -867,6 +873,20 @@ void GeneralSettings2::StoreConfig()
...
+	if (CafeSystem::IsTitleRunning())
+	{
+		if (config.disable_screensaver)
+		{
+			ScreenSaver::SetInhibit(true);
+		}
+		else
+		{
+			ScreenSaver::SetInhibit(false);
+		}
```
Tachi107, line 887: Can be simplified to

```cpp
ScreenSaver::SetInhibit(config.disable_screensaver)
```

Tachi107 (commented):
One last nit

lisa-wolfgang:
Everything looks to be in a good state at this point. I doubt that Windows and macOS would ignore SDL's screen saver inhibition -- we're not doing it in any particularly special way.

lisa-wolfgang:
@Exzap I hope I'm not being a bother, but I just wanted to make sure you knew that this PR is now in a solid state. All outstanding issues from the past couple of months have been resolved. SDL is now used for screen saver inhibition, so cross-platform (and hopefully cross-distro) support is much better than it was with the original approach.

Exzap:
Looks good to me as-is, no changes necessary. Thanks for looking into this.
Can you merge with upstream/rebase one more time? Github says there are conflicts

lisa-wolfgang:
Hmm, it looks like it's already up-to-date with upstream/master. I'm not seeing any conflicts on my end. What are the conflicts that you're seeing?

Exzap:
Seems fine now. Thanks anyway

PR #143, opened 2022-09-01
WIP: Add wayland support

jambonmcyeah:
Only Vulkan support is added so far I'll do OpenGL soon.

It currently only works on fullscreen and it's really glitchy in windowed mode.

PR #500, opened 2022-11-16
Input settings: Fix combobox dropdown not working on linux

### LINKED ISSUE #460: [Linux] Controller profiles cannot be selected
In the input settings, the dropdown to select controller profiles is unselectable, even though I have multiple profiles for the selected controller. I've confirmed that saving profiles still works -- they appear in the `controllerProfiles` folder.

Cemu version: https://github.com/cemu-project/Cemu/commit/a40b226e003b7ad6a21612f6a1a72ba09f102faa
Operating system: Ubuntu 22.10

goeiecool9999:
Fixes #460
Cemu doesn't add the controller profile entries to the combobox until a dropdown is triggered but (at least on my system) GTK grays out the combobox if it's empty, so there's no way to trigger a dropdown event to populate it.


jn64:
Works here, finally can load profiles (and immediately found a different bug with it :bug: )

goeiecool9999:
There's one more issue with this. When there are no profiles and you save for the first time the newly created profile doesn't get added to the combobox so the dropdown is not available when it should be.

goeiecool9999:
Probably caused by this being added to GTK version 2.14
https://lazka.github.io/pgi-docs/Gtk-3.0/classes/ComboBox.html#Gtk.ComboBox.set_button_sensitivity

jn64:
Also...

- Save n profiles, restart Cemu and open the dropdown (it should be n height)
- Save 1 more profile
- Open the dropdown, it's still n height, with scroll arrows
- Close and open the dropdown again, now it's n+1 height

goeiecool9999:
@Tachi107 I've used your FindGTK3.cmake file from #143 is the license notice okay that way?

cmake/FindGTK3.cmake
```diff
@@ -0,0 +1,21 @@
+# SPDX-FileCopyrightText: 2022 Andrea Pappacoda <andrea@pappacoda.it>
+# SPDX-License-Identifier: ISC
+
+include(FindPackageHandleStandardArgs)
+
+find_package(PkgConfig REQUIRED)
+if (PKG_CONFIG_FOUND)
```
Tachi107, line 7: `REQUIRED` should be dropped here

cmake/FindGTK3.cmake
```diff
@@ -0,0 +1,21 @@
...
+    pkg_search_module(GTK3 REQUIRED gtk+-3.0)
+    if (GTK3_FOUND)
```
Tachi107, line 9: `REQUIRED` should be dropped here too

cmake/FindGTK3.cmake
```diff
@@ -0,0 +1,21 @@
...
+        target_link_libraries(GTK3::gtk INTERFACE ${GTK3_LIBRARIES})
+        target_link_directories(GTK3::gtk INTERFACE ${GTK3_LIBRARY_DIRS})
+        target_include_directories(GTK3::gtk INTERFACE ${GTK3_INCLUDE_DIRS})
+    endif()
+    find_package_handle_standard_args(GTK3
+            REQUIRED_VARS
+            GTK3_LINK_LIBRARIES
+            GTK3_FOUND
+            VERSION_VAR GTK3_VERSION
+    )
```
Tachi107, line 20: I'd suggest using `pkg_search_module()`'s `IMPORTED_TARGET` option, like this:

```suggestion
    pkg_search_module(GTK3 IMPORTED_TARGET gtk+-3.0)
    if (GTK3_FOUND)
        add_library(GTK3::gtk ALIAS PkgConfig::GTK3)
    endif()
    find_package_handle_standard_args(GTK3
            REQUIRED_VARS GTK3_LINK_LIBRARIES
            VERSION_VAR GTK3_VERSION
    )
```

cmake/FindGTK3.cmake
```diff
@@ -0,0 +1,21 @@
...
+find_package(PkgConfig REQUIRED)
+if (PKG_CONFIG_FOUND)
+    pkg_search_module(GTK3 REQUIRED gtk+-3.0)
+    if (GTK3_FOUND)
+        add_library(GTK3::gtk IMPORTED INTERFACE)
+        target_link_libraries(GTK3::gtk INTERFACE ${GTK3_LIBRARIES})
+        target_link_directories(GTK3::gtk INTERFACE ${GTK3_LIBRARY_DIRS})
+        target_include_directories(GTK3::gtk INTERFACE ${GTK3_INCLUDE_DIRS})
+    endif()
+    find_package_handle_standard_args(GTK3
```
Tachi107, line 15: I usually put `find_package_handle_standard_args()` unconditionally at the end of the Find module, but I'm not sure if it really matters at all.

Tachi107 (changes requested):
> @Tachi107 I've used your FindGTK3.cmake file from #143 is the license notice okay that way?

Sure - the ISC license doesn't require you to ask the author for usage, by the way :)

I actually don't remember writing that piece of code, but I have some suggestions below.

goeiecool9999:
> I actually don't remember writing that piece of code,

I have a feeling the person who created the Wayland PR took one of your other CMake files as a base and didn't take the copyright header out. That explains the confusion.

amayra:
I cant even select UP and DOWN key arrow in native Linux bin unlike wine

Tachi107:
> Thank you very much. Do you know any good resources to better understand CMake?

CMake is huge, complex, and hard to get right. I somewhat understand it because I have to deal with lots of C and C++ projects being a Debian packager, but I don't know of any _good_ guide that gives truly useful advice, and even if it existed it would eventually get outdated. What I can suggest is to read the documentation of the various [CMake commands](https://cmake.org/cmake/help/latest/manual/cmake-commands.7.html) (functions), as it often hints at what the "right" thing to do is, and it's always up to date.

For example, the [`install()`](https://cmake.org/cmake/help/latest/command/install.html) docs give some useful advice: it says to avoid absolute destination paths, it [praises](https://cmake.org/cmake/help/latest/command/install.html#installing-files) [file sets](https://cmake.org/cmake/help/latest/command/target_sources.html#file-sets), it encourages the use of components, etc.

Luckily there are [saner build systems](https://mesonbuild.com) out there :)

Issue #343, opened 2022-10-06
[macOS] Cemu UI Rendering at half Retina resolution


Veemyu:
macOS's Retina Resolutions work by displaying the UI in a logical, for example 1920x1080 resolution ie. also resulting in a size 1080P would normally be displayed on the Monitor. Yet Retina is supposed to render the actual UI in 3840x2160, always double the logical resolution.
Cemu currently renders, like the console prints out, at Scale 1.0

This Results in a 1280x720 Window (logically) taking up 2560x1440 Real pixels but the underlying rendering is only at 1280x720
This results in a really blurry and pixelated look. The UI is mostly fine, the problem is with the Vulkan Canvas.

This should maybe help, the all too familiar Scale 1.0 needs to be 2.0 on Retina screens, not on the whole Apple Platform though!
https://github.com/KhronosGroup/MoltenVK/issues/428 
They have the exact same issue solved up there.

`[mvk-info] Created 3 swapchain images with initial size (1280, 720) and contents scale 1.0 for screen Built-in Retina Display.`

Here are comparison pictures between 1512x982 with Retina (HiDPI) which should display as 3024x1964 everywhere!!
And also 3024x1964 Native in fullscreen which is also supposed to display as 3024x1964 obviously.
<img width="650" alt="Menu Native" src="https://user-images.githubusercontent.com/38016914/194294894-478b75d8-7264-4d5d-86e3-9b8cddd3ca03.png">
<img width="650" alt="Menu HiDPI" src="https://user-images.githubusercontent.com/38016914/194294916-94f5a5ce-c329-4067-bf26-04071593451b.png">
<img width="650" alt="Game Native" src="https://user-images.githubusercontent.com/38016914/194294924-8ec637bc-fba3-4ab1-b449-68ba251dece3.png">
<img width="650" alt="Game HiDPI" src="https://user-images.githubusercontent.com/38016914/194294959-cd2b2653-80cd-415a-88b5-c43915deb3ae.png">

556644O:
I also found the same issue. Some other softwares have the same issue, such as PPSSPP and Streets of Rage 4. At present, I use Easyres to solve this issue by setting the resolution to 1X. Or is there a better approach out there to solve this issue?
![截屏2022-11-14 11 25 46](https://user-images.githubusercontent.com/116966347/201569957-ac19ff94-9264-4ac8-ad8d-3e98eae6b64d.png)

greybaron:
[...]

PR #502, opened 2022-11-17
Vulkan: Further simplify swapchain code

goeiecool9999:
- Fix use after free bug in CreateSwapchainCreateInfo
- Always use minImageCount. Unclear why it was min+1. This also reduces input lag on double buffered vsync by one frame.
- Remove unnecessary semaphore, it is unnecessary because the image is awaited by means of a fence.
- Don't submit command buffer in AcquireNextSwapchainImage
- Don't wait for fence when recreating; already waited for device to be idle.
- Remove barrier before present. VkQueuePresentKHR is set to wait for the presentSemaphore which becomes signalled when the command buffer that writes to the swapchain image finishes execution. No barrier is required because: ```Any writes to memory backing the images referenced by the pImageIndices and pSwapchains members of pPresentInfo, that are available before vkQueuePresentKHR is executed, are automatically made visible to the read access performed by the presentation engine. This automatic visibility operation for an image happens-after the semaphore signal operation, and happens-before the presentation engine accesses the image.```
- Remove SwapBuffer error handling loop and replace with code similar to AcquireNextSwapchainImage error handling

> it is unnecessary because the image is awaited by means of a fence.

Clarification: while working on this PR I had a version where I deleted the fence instead and relied on the semaphore to synchronize access to the swapchain image. This worked as well but the specification has an informative note in vkAcquireNextImageKHR saying: ```Applications should not rely on vkAcquireNextImageKHR blocking in order to meter their rendering speed. The implementation may return from this function immediately regardless of how many presentation requests are queued, and regardless of when queued presentation requests will complete relative to the call. Instead, applications can use fence to meter their frame generation work to match the presentation rate.```
> Applications should not rely on vkAcquireNextImageKHR blocking in order to meter their rendering speed.

Getting rid of the fence does exactly this. But if we use the fence anyway the semaphore becomes redundant because we do not submit the command buffer that writes to the acquired swapchain image before the fence is signalled.

Remaining questions:
These statements seem to contradict eachother:
Normative text:
> If timeout is UINT64_MAX, the timeout period is treated as infinite, and vkAcquireNextImageKHR will block until an image is acquired or an error occurs.

Informative text:
> The implementation may return from this function immediately regardless of how many presentation requests are queued, and regardless of when queued presentation requests will complete relative to the call. 

It would be nicer if we _could_ rely on vkAcquireNextImageKHR to block because then all the access synchronization happens on-device although in practice this probably makes little difference.

### DIFF
--- a/src/Cafe/HW/Latte/Renderer/Vulkan/SwapchainInfoVk.cpp
+++ b/src/Cafe/HW/Latte/Renderer/Vulkan/SwapchainInfoVk.cpp
@@ -118,6 +118,8 @@ void SwapchainInfoVk::Create(VkPhysicalDevice physicalDevice, VkDevice logicalDe
 	result = vkCreateFence(logicalDevice, &fenceInfo, nullptr, &m_imageAvailableFence);
 	if (result != VK_SUCCESS)
 		UnrecoverableError("Failed to create fence for swapchain");
+
+	hasDefinedSwapchainImage = false;
 }
 
 void SwapchainInfoVk::Cleanup()
--- a/src/Cafe/HW/Latte/Renderer/Vulkan/SwapchainInfoVk.cpp
+++ b/src/Cafe/HW/Latte/Renderer/Vulkan/SwapchainInfoVk.cpp
@@ -14,7 +14,7 @@ void SwapchainInfoVk::Create(VkPhysicalDevice physicalDevice, VkDevice logicalDe
 	m_actualExtent = ChooseSwapExtent(details.capabilities);
 
 	// calculate number of swapchain presentation images
-	uint32_t image_count = details.capabilities.minImageCount + 1;
+	uint32_t image_count = details.capabilities.minImageCount;
 	if (details.capabilities.maxImageCount > 0 && image_count > details.capabilities.maxImageCount)
 		image_count = details.capabilities.maxImageCount;
 
@@ -115,16 +115,6 @@ void SwapchainInfoVk::Create(VkPhysicalDevice physicalDevice, VkDevice logicalDe
 			UnrecoverableError("Failed to create semaphore for swapchain present");
 	}
 
-	m_acquireSemaphores.resize(m_swapchainImages.size());
-	for (auto& semaphore : m_acquireSemaphores)
-	{
-		VkSemaphoreCreateInfo info = {};
-		info.sType = VK_STRUCTURE_TYPE_SEMAPHORE_CREATE_INFO;
-		if (vkCreateSemaphore(logicalDevice, &info, nullptr, &semaphore) != VK_SUCCESS)
-			UnrecoverableError("Failed to create semaphore for swapchain acquire");
-	}
-	m_acquireIndex = 0;
-
 	VkFenceCreateInfo fenceInfo = {};
 	fenceInfo.sType = VK_STRUCTURE_TYPE_FENCE_CREATE_INFO;
 	fenceInfo.flags = VK_FENCE_CREATE_SIGNALED_BIT;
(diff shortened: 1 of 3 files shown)


Veemyu:
Just a quick question, if this is even in your power.
[issues/343] (https://github.com/cemu-project/Cemu/issues/343)
Concerning Retina Resolutions not being applied correctly.
Console states this.
`[mvk-info] Created 3 swapchain images with initial size (1280, 720) and contents scale 1.0 for screen Built-in Retina Display.`
While the logical resolution is 1280x720
The physical resolution is 2560x1440

So CEMU renders at half the resolution on Macs with scaling eg. 80% of them.
This might also be a MoltenVK issue. But maybe some part of the swap-chain code affects this too.
It's a long standing issue, I hope if something rings a bell you might be able to tackle this issue.

goeiecool9999:
Considering that Linux has similar issues with display scaling I have a feeling that tackling one will fix the other. I'm definitely planning on looking into it.

goeiecool9999:
@Veemyu try #514

Exzap:
It's probably a mistake to trust the implementation to immediately signal the fence for image availability in `vkAcquireNextImageKHR` even if one is actually available. The spec doesn't guarantee a maximum delay. I found this older [issue](https://gitlab.freedesktop.org/mesa/mesa/-/issues/3039) which proves the point that there are implementations out there that have/had a delay.

goeiecool9999:
Is roughly like you meant? @Exzap

Exzap:
Yes thats pretty much what I meant.

goeiecool9999:
I noticed corrupted graphics when resizing the window. Before I started tweaking the swapchain code cemu would not update at all while resizing. With this change it'll flicker black if it the recreation happens in between the game writing its frame to the swapchain image and the presentation of the frame. It's not super noticeable and definitely better than flashing corrupted graphics. Will look for a way to get rid of the flicker as well.

SupervisedThinking:
Just wondering, is there any way to get rid of the glslang dependecy & reuse Vulkan/Mesa headers instead of e.g.: https://github.com/cemu-project/Cemu/blob/main/src/Cafe/HW/Latte/Renderer/Vulkan/VKRPipelineInfo.cpp#L6

goeiecool9999:
> I'm aware of that but usually glslang is AFAIK(!) a validator which checks the code before runtime e.g.

glslang is the library that takes GLSL and turns them into SPIR-V for use in Vulkan Shader modules. It's an essential component that Cemu doesn't work without.

amayra:
> > I'm aware of that but usually glslang is AFAIK(!) a validator which checks the code before runtime e.g.
> 
> glslang is the library that takes GLSL shaders and turns them into SPIR-V for use in Vulkan Shader modules. It's an essential component that Cemu doesn't work without.

why can't we go from Wii U shader binary back to SPIR-V is there are any obstacles for this to happen?

what is the list of Vulkan extensions cemu use to run anyway ?

Exzap:
> why can't we go from Wii U shader binary back to SPIR-V is there are any obstacles for this to happen?

It's possible in theory, but writing a new shader translator that supports SPIR-V output is everything but trivial. It might happen eventually.

> what is the list of Vulkan extensions cemu use to run anyway ?

See `kOptionalDeviceExtensions` and `kRequiredDeviceExtensions` in https://github.com/cemu-project/Cemu/blob/c0ebbef9bd10d7d67cc2482ce9ef238b0a53cbf1/src/Cafe/HW/Latte/Renderer/Vulkan/VulkanRenderer.cpp

SupervisedThinking:
> > I'm aware of that but usually glslang is AFAIK(!) a validator which checks the code before runtime e.g.
> 
> glslang is the library that takes GLSL shaders and turns them into SPIR-V for use in Vulkan Shader modules. It's an essential component that Cemu doesn't work without.

Indeed well I guess I wasn't precise what I meant. Usually glslang isn't a system dependency unless you want to / need to run `glslangValidator` - the usual suspects like Dolphin or PCS2 build a static `libglslang.a` & RPCS3 e.g. `libSPIRV.a` so non of them relies on system packages but all added glslang as submodule & build the needed libs at build time.

PR #510, opened 2022-11-18
Fix gamepad hotkey and game profile setting

goeiecool9999:
This requires GTK headers so I added GTK as a dependency (no actual new dependency, wxwidgets already linked libgtk). Which also allows replacing all explicitly dynamically loaded calls with normal calls.
This also changes the way KeyboardController stops sending input after loss of focus. I changed the keyboard code that was already in gui_wrapper/CemuApp so it works as intended. Using the KILL_FOCUS event didn't work because pressing tab would immediately trigger that event and unpress all keys. The ACTIVATE_APP event doesn't have that issue. Now that the keyboard state tracking code updates to make all keys unpresed when focus is lost KeyboardController doesn't need to check if the windows is active anymore.

### DIFF
--- a/src/gui/CemuApp.cpp
+++ b/src/gui/CemuApp.cpp
@@ -218,10 +218,6 @@ int CemuApp::FilterEvent(wxEvent& event)
 		const auto& key_event = (wxKeyEvent&)event;
 		g_window_info.set_keystate(fix_raw_keycode(key_event.GetRawKeyCode(), key_event.GetRawKeyFlags()), false);
 	}
-	else if(event.GetEventType() == wxEVT_KILL_FOCUS)
-	{
-		g_window_info.set_keystatesdown();
-	}
 
 	return wxApp::FilterEvent(event);
 }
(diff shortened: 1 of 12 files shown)


src/gui/CemuApp.cpp
```diff
@@ -218,10 +218,6 @@ int CemuApp::FilterEvent(wxEvent& event)
 		const auto& key_event = (wxKeyEvent&)event;
 		g_window_info.set_keystate(fix_raw_keycode(key_event.GetRawKeyCode(), key_event.GetRawKeyFlags()), false);
 	}
-	else if(event.GetEventType() == wxEVT_KILL_FOCUS)
-	{
-		g_window_info.set_keystatesdown();
-	}
```
goeiecool9999, line 224: On linux pressing tab triggers this event, immediately resetting the tab key state. Makes the button unpressable. Perhaps some workaround could work where ignores the kill focus events when tab is held.

goeiecool9999: Fixed by checking if window is active in LatterRenderTarget. Perhaps it would be nicer to just set all keys to up when app is marked inactive in CemuApp. That way the Keyboard input manager also doesn't need to look that up.

PR #514, opened 2022-11-20
Fix render resolution at different UI scales

goeiecool9999:
I do not have a HiDPI/retina display at my disposal so I'd appreciate if people who do can test this.


goeiecool9999:
Disregard this for now. I wasn't paying close enough attention. Cemu already uses currentExtent when it is defined.

goeiecool9999:
Vulkan on wayland apparently returns 0xFFFFFFFF for the currentExtent. Not sure if that's also the case with XWayland. If it does though it would explain this behaviour. It's unnoticeable when using 1x scaling because cemu has a fallback to wxWidgets window coordinates, but when using scaling the size is no longer in physical pixels so the swapchain extent doesn't match the window size and you get graphical issues like this: 
image
using the physical pixel size should fix this. MacOS apparently also doesn't render at native resolution but this could be a different issue.

greybaron:
I have created a pull request #516 that fixes high DPI scaling on macOS which works quite well. (I wasn't aware of your changes)

I suggest you compare the changes and assess whether your macOS change is still necessary or if my implementation can be improved.

greybaron:
I have tested these changes on a Retina Mac, and they work well. However, there is also the same issue, that after a DPI change only the Metal backend resolution changes. So if the DPI changes from high DPI to normal, the view will only show the cut off top left area until the window is resized. I believe this is because `MainWindow::OnSizeEvent` is not triggered on DPI change, so `phys_width`/`phys_height` stay unchanged even though they should change.

goeiecool9999:
@greybaron I'm afraid that isn't the problem. According to [this wxWidgets forum post](https://forums.wxwidgets.org/viewtopic.php?p=194433&sid=0b5280bd7f37bcdce88b8b791845b33e#p194433) DPI changes across different monitor is not supported as of 2019. So I think to get DPI updates across monitors working we need to wait for wxWidgets to support it.

goeiecool9999:
> he view will only show the cut off top left area until the window is resized

Sorry. I missed this part. You're probably correct then!

goeiecool9999:
@greybaron I've added a DPIChanged event handler. Give the latest commit a try!

goeiecool9999:
> Cemu crashes on macOS if compiled with the merged commits from main.

I can't reproduce it myself so it might be DPI related. Crash log would be much appreciated.

greybaron:
I have tried some more, this is really weird:
- it happens regardsless of high DPI being used or not
- after trying for around 30 times, it did actually **not crash** twice
- most of the time, it crashes with `Error: signal 11:`
- sometimes, it crashes with signal 6.

[signal 6.log](https://github.com/cemu-project/Cemu/files/10076103/signal.6.log)
[signal 11.log](https://github.com/cemu-project/Cemu/files/10076104/signal.11.log)

goeiecool9999:
And it doesn't crash on the latest commit on the main branch?

greybaron:
correct.

greybaron:
@goeiecool9999 Just to make sure:
 
[THIS](https://github.com/cemu-project/Cemu/files/10078184/crashing_Cemu_release.zip) is my binary which is (almost always) crashing. Can you confirm that this does not crash for you?

goeiecool9999:
Okay strangely enough your binary runs fine for me. This is a bit puzzling.

goeiecool9999:
I have a branch [here](https://github.com/goeiecool9999/Cemu/tree/scalingfixcrashfixtest) (scalingfixcrashfixtest) that merges #525 with this PR. I hope it relieves the crashing issue. The changes made in #502 make cemu interact with vulkan in a non-standard way and it may be causing issues with MoltenVK. It's a long shot but it might work.

Tillsunset:
I've done some testing on my intel mac. 
The [main](https://github.com/cemu-project/Cemu/tree/c0ebbef9bd10d7d67cc2482ce9ef238b0a53cbf1) branch works fine for me. 

The "[scalingfix](https://github.com/goeiecool9999/Cemu/tree/608021dd94a735bd8066124c989b0c4c8f25b704)" branch and "[scalingfixcrashfixtest](https://github.com/goeiecool9999/Cemu/tree/9448e56b8cef3941e90161dfea296390dd487e9e)" branch crashes with
"libc++abi: terminating with uncaught exception of type std::length_error: allocator<T>::allocate(size_t n) 'n' exceeds maximum supported size
Error: signal 6:". 
[log.txt](https://github.com/cemu-project/Cemu/files/10081364/log.txt)

"[fixchainimagecount](https://github.com/goeiecool9999/Cemu/tree/460ddc40d92a6bab8beceabde9b059f1048f61e8)" works fine.

goeiecool9999:
Up till now I experienced the crash once. I have tried everything I could think of. I'm not sure how to continue looking from here. I know this may be a big ask but it could be helpful if someone rebased this branch onto main and then bisected the commit that causes the crash.

greybaron:
I'm pretty inexperienced with git, so I'll tell you step by step what I did...

- first I cloned this branch, **went back to the commit before the merge**, then checked out to a new branch
- I then rebased that branch to upstream/main. Same as you, I had a conflict in `VulkanCanvas.cpp` (I accepted incoming like you)
- (THIS)[https://user-images.githubusercontent.com/20914657/203842254-430f2e8e-82ec-4c81-9cc7-0701cd5d6a39.png] is the resulting commit history
- git bisect result: _the crashing starts after the commit with the conflict (316a38c)._

greybaron:
The reason i'm confused is that I still don't know which commit **to upstream** broke your commit ab0791d. However I guess that is why there was a merge conflict. Let me know if I can try anything else

goeiecool9999:
Thank you very much

> (I accepted incoming like you)

I don't believe I did accept incoming. I applied both the parent->GetClientSize() change and the wxNO_FULL_REPAINT_ON_RESIZE change in the merge commit.
However that doesn't matter. You're right this is confusing.
The obvious suspect is wxNO_FULL_REPAINT_ON_RESIZE. It appears that that flag combined with the changes in scalingfix is somehow causing the crash but that seems highly unlikely and we just learned that not using it also doesn't fix the problem either. So what's going on here?

One last thing we could try is to rebase main onto scalingfix with something like
```
git switch -c mainontoscaling origin/main
git rebase --onto=84361403 8f1ee9e5
```
and then bisect again.
Then you're bisecting upstream changes applied _after_ scaling fix instead of before. I really hope this will reveal something otherwise I'm tempted to chalk this up to undefined behaviour or something.

greybaron:
Now when rebasing, should I go with HEAD or incoming?

image

goeiecool9999:
> Now when rebasing, should I go with HEAD or incoming?

the commit should have both parent->GetClientSize() and wxNO_FULL_REPAINT_ON_RESIZE. I'm not sure what the vscode "accept both changes" does but that sounds most like what you want. It probably lets you manually edit the file if it can't do it automatically.

greybaron:
Finally found it 🎉

This is the issue:
image

Reverting the change fixes the crashing.

goeiecool9999:
Great! I'm still not sure why that change is causing a crash but at least that's something I can work with!

goeiecool9999:
Let me know if the latest version crashes or not. I think it should be fixed!

Tillsunset:
Still crashes, same logs

goeiecool9999:
> Still crashes, same logs

Very surprised by that. I thought for sure that was it.

goeiecool9999:
I'm just throwing stuff against the wall at this point but does this change anything?

greybaron:
still crashing

goeiecool9999:
I guess one thing we should try is to run with
``MTL_DEBUG_LAYER=1``
and see if it outputs anything extra before the crash. And if it doesn't maybe we should also try to trigger the crash with vulkan validation layers. Although I haven't gotten them to work on macOS so it might be a bit tricky.

goeiecool9999:
I have a feeling that imgui doesn't like rendering to the 40x40 image, and I think that this change should prevent an image like that from being created. Please let me know if it works.

greybaron:
I can confirm the crash is fixed. Thank you for your hard work!

Out of curiosity, did you find out why the crash never occurred on your machine?

goeiecool9999:
>I can confirm the crash is fixed.

Awesome!

> Out of curiosity, did you find out why the crash never occurred on your machine?

I have no idea. I did notice that my log never lists the tiny resolutions, even while using hi-DPI resolutions. That's probably why it doesn't crash for me. As to _why_ it doesn't use the tiny resolutions is anyone's guess. I'm not on official apple hardware and use an older macOS version so it could be a lot of different things.

goeiecool9999:
UI seems to change scale from hiDPI to regular DPI when moved between screens in macOS but I don't think that's a result of this patch.
IIRC per-monitor scaling is not supported on linux so I'd say displaying the right resolution and filling the screen at a different scale is good enough for now.
Then there's windows which from what I read [here](https://docs.wxwidgets.org/trunk/overview_high_dpi.html) would require a lot of work to get scaling working. Though users who are setting the DPI override in the executable properties might still benefit from this change. UI will probably be blurry but the game might be the right resolution.
I'm renaming the PR once again because calling it DPI awareness seems like an overstatement.

natopotato390:
I've just updated to 2.0-19 recently and I'm still having this issue, Moved from a 1080p monitor to a 4k TV and the UI and game is rendered at half res.

PR #522, opened 2022-11-21
Vulkan: Fix flicker on resize

goeiecool9999:
When a swapchain becomes suboptimal after an acquire for rendering, SwapBuffer would recreate the swapchain and acquire a new image from it which would get filled in with black because the content is undefined, resulting in black flickering when resizing.
By changing the order in AcquireNextSwapchainImage it'll always try to present the suboptimal image it already acquired and drew content to and drop the image if it is out of date and cannot be used. Then when no image is reserved anymore the next acquire will recreate the swapchain.


Exzap:
FYI on Windows there is a white flicker when resizing. It's been there pre-2.0 too. Interestingly, this PR doesn't affect it.

PR #524, opened 2022-11-21
UI addons

SSimco:
Add UI options to open the UserDataPath folder and remove shader caches.

### DIFF
--- a/src/gui/MainWindow.cpp
+++ b/src/gui/MainWindow.cpp
@@ -71,6 +71,7 @@ enum
 	// file
 	MAINFRAME_MENU_ID_FILE_LOAD = 20100,
 	MAINFRAME_MENU_ID_FILE_INSTALL_UPDATE,
+	MAINFRAME_MENU_ID_FILE_OPEN_CEMU_FOLDER,
 	MAINFRAME_MENU_ID_FILE_EXIT,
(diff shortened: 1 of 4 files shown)


qurious-pixel:
image

Crementif:
The Open Cemu folder option was long overdue indeed so it's nice to see it being PRed.

I do wonder if three shader cache options are really necessary. If there's not a use-case for being able to delete the precompiled caches specifically besides a MoltenVK issue (where a proper fix would be preferred to cluttering a menu with an additional option) it should be reduced to a singular option that just deletes all caches.

I also think the confirmation prompt is unnecessary. Graying out the deletion option makes more sense to me then having a prompt that shows whether deletion failed due to no cache existing.

SSimco:
> The Open Cemu folder option was long overdue indeed so it's nice to see it being PRed.
> 
> I do wonder if three shader cache options are really necessary. If there's not a use-case for being able to delete the precompiled caches specifically besides a MoltenVK issue (where a proper fix would be preferred to cluttering a menu with an additional option) it should be reduced to a singular option that just deletes all caches.
> 
> I also think the confirmation prompt is unnecessary. Graying out the deletion option makes more sense to me then having a "no shader cache was found" prompt inside the code.

I merged the three options into one and it should disable it if no caches are found.

Crementif:
Looks good, thanks!

PR #525, opened 2022-11-22
Vulkan: return to more conventional swapchain sync method, encapsulate more code

goeiecool9999:
#502 adopted a swapchain sync method that is correct in theory but is not commonly used. 
This changes the sync method back to what was used before.

The [validation layer](https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/12) appears to not have made up their mind about whether using only a fence is valid and one of the authors of the vulkan specification says it is unintended and basically advises against it [here](https://github.com/KhronosGroup/Vulkan-Docs/issues/117#issuecomment-200357476)

This PR also ups the minimum swapchain image count to at least two images (not min+1 which might be > 2 causing unnecessary input lag and wasting resources). Using only one swapchain image caused a freeze on AMD windows drivers when switching to fullscreen.
One of the issues that popped up after removing the acquire semaphore is that on macOS the fonts in the imgui overlay disappeared.
I do not have a concrete answer as to why moltenVK has issues while other platforms do not. It may be that moltenVK assumes swapchain image access is guarded with a semaphore.

Additionally some more swapchain code is moved from VulkanRenderer into the SwapchainInfoVk class. There is likely more code that could be moved there before the encapsulation is optimal but that would make the scope of the PR a bit too large.

### DIFF
--- a/src/Cafe/HW/Latte/Renderer/Vulkan/SwapchainInfoVk.cpp
+++ b/src/Cafe/HW/Latte/Renderer/Vulkan/SwapchainInfoVk.cpp
@@ -15,10 +15,10 @@ void SwapchainInfoVk::Create(VkPhysicalDevice physicalDevice, VkDevice logicalDe
 
 	// Use at least two swapchain images. Fewer than that causes problems on some drivers.
 	uint32_t image_count = std::max(2u, details.capabilities.minImageCount);
-    if(details.capabilities.maxImageCount > 0)
-        image_count = std::min(image_count, details.capabilities.maxImageCount);
-    if(image_count < 2)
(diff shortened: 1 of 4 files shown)


Exzap:
Since you already introduced `SwapchainInfoVk` you should use methods and not expose `m_imageAvailableFence` to VulkanRenderer directly.

I'd also prefer if we could avoid the usage of a timeout in `vkWaitForFences`. Whether the fence is going to be signaled or not should always be a well defined state?

goeiecool9999:
I think this moves the encapsulation of the swapchain in the right direction. I think it could go further but that would require more effort

PR #528, opened 2022-11-24
Vulkan: Correct swapchain error handling

goeiecool9999:
In #514 there were some mysterious crashes.
[greybaron](https://github.com/greybaron) was helpful in bisecting which revealed that #522 introduces a bug.
Because #522 checks to see if an image is already acquired _before_ checking if the swapchain is out of date the swapchain does not get recreated on a VK_ERROR_OUT_OF_DATE_KHR error. This meant that the swapchainImageIndex never got reset and cemu would try to present images it never acquired. This wasn't an issue on main and I think that is because only DPI switches trigger VK_ERROR_OUT_OF_DATE_KHR and only on some macOS systems (like the M1 macbook pro).


goeiecool9999:
I jumped to conclusions. This doesn't fix the bug

PR #543, opened 2022-11-30
Cross-platform screenshots

tomlally:
Uses wxWidgets API to write screenshots to clipboard and disk (instead of win32). Should work on all platforms. Screenshot notifications may be too verbose.


Exzap:
This ended up being quite tricky to figure out. Calling `CoInitialize` didn't help but `OleInitialize()` did resolve the error message.
Since I was already working on the code I decided to tweak it a bit:
- Screenshot filenames now contain the date instead of just a non-descriptive counter. GamePad screenshots will have `_GamePad` in the name
- Protecting clipboard access with a mutex because it could still fail when spamming the screenshot key due to attempting to open multiple instances simultanously
- Made notifications less verbose and only display one message per screenshot

PR #545, opened 2022-11-30
Automatically scale imgui text based on display pixel density

goeiecool9999:


emiyl:
Before
 
image

After

image

PR #555, opened 2022-12-05
Fix sign error of distance calculate

iamqk:
will cause distance == 0, when distance is a minus one.


iamqk:
_testAsm(0x4182FFB4, "beq .-0x4C");
fix this test failed

Exzap:
thanks!

PR #561, opened 2022-12-07
Minor AppImage fixes

jn64:
- Fix incorrect recursive copy. Previously, if you didn't delete AppDir between builds, `bin/` would be copied *into* an existing `AppDir/usr/share/Cemu`

- Add metainfo file. [According to appstream docs](https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html#qsr-app-introduction), `metainfo.xml` is correct, but [the warning from linuxdeploy](https://github.com/cemu-project/Cemu/actions/runs/3640469545/jobs/6145427858#step:5:1494) says `appdata.xml` so that's where I put it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -100,7 +100,7 @@ jobs:
     - name: "Install system dependencies"
       run: |
         sudo apt update -qq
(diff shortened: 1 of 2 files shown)


qurious-pixel:
install `appstream` and appstreamcli will validate the appdata.xml

PR #563, opened 2022-12-08
fix: use an alt implementation for tick_cached()

ElijahPepe:
std::chrono::steady_clock::now() may be slower than clock_gettime() (a direct call and vDSO). For an even faster implementation, cycle counting (through rdtsc) is likely needed, but this level of depth is unnecessary.

### DIFF
--- a/src/Common/precompiled.h
+++ b/src/Common/precompiled.h
@@ -69,6 +69,7 @@
 #include <filesystem>
 #include <memory>
 #include <chrono>
+#include <time.h>
 #include <regex>
 #include <type_traits>
 #include <optional>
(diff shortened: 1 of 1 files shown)


Tillsunset:
I did some testing and clock_gettime() got a slight speedup, 5-3% speedup. I also tested gettimeofday() and got a 40% speedup compared to now(). [File](https://gist.github.com/Tillsunset/1646b880a2847eda6fde9fde86d2e791) I used to test. I was using a mac and -o3 so this might not translate to other system, but it might be good idea to do some testing and post the result

ElijahPepe:
`gettimeofday()` can cause some issues (i.e. NTP synchronization, leap seconds, skewing of the system clock). It's best to stick with a function that may be slower but returns a the real monotonic clock.

Exzap:
Looks good to me. Thanks!

> For an even faster implementation, cycle counting (through rdtsc) is likely needed

We do exactly this for emulating the PowerPC timer since performance there is critical. The frequency is approximated so it can have a tiny drift which makes it unsuitable as a global solution. It would also be very overkill. Even `now_cached()` is probably unnecessary nowadays. It used to be required because the first implementations of std::chrono were quite unoptimized.

> [File](https://gist.github.com/Tillsunset/1646b880a2847eda6fde9fde86d2e791) I used to test. I was using a mac and -o3 so this might not translate to other system, but it might be good idea to do some testing and post the result

On godbolt for Linux clang x86-64 with -O3 I get:
```
Time taken by clock_gettime function: 239 microseconds
Time taken by gettimeofday function: 680 microseconds
Time taken by now function: 252 microseconds
```
So it seems gettimeofday is much slower there.

On a side note, something you have to be careful about when benchmarking short code is dead-code elimination, especially with -O2 and -O3. Since you don't store/sum up the results of the conversion to `std::chrono::steady_clock::time_point` the compiler is free to optimize it away. Under rare circumstances the compiler might also optimize away entire calls to external functions but you'd be able to tell if that happens from the result.

PR #565, opened 2022-12-09
Manpage

jn64:
See commit messages for details

Todo:

- AUTHORS — I took it from the *About* dialog of Cemu. Some manpages include "and contributors" or similar wording, or simply omit AUTHORS.
- Copyright statement ­— I didn't find a definitive copyright statement in README.md or source files. Also optional.
- I don't know what "URL prefix for account server" is (protocol? hostname or part of hostname? fully formed URL?)


Tachi107 (changes requested):
Instead of shipping a pre-generated artefact, could you please add a `man` target to the CMake build system instead?

Tachi107:
First of all thank you for bringing this up! `man` pages are underrated :)

>   * ~AUTHORS — I took it from the _About_ dialog of Cemu. Some manpages include "and contributors" or similar wording, or simply omit AUTHORS.~
>
>     * Actually reading man-pages(7) says this section is **strongly discouraged**. Will remove unless its wanted

I personally don't fully agree with the opinion expressed in the [`man-pages(7)`](https://man7.org/linux/man-pages/man7/man-pages.7.html) manual, and in fact other manuals such as [`mdoc(7)`](https://man.openbsd.org/mdoc.7) or [`scdoc(5)`](https://git.sr.ht/~sircmpwn/scdoc/tree/master/item/scdoc.5.scd) do not discourage this at all.

>   * Copyright statement ­— I didn't find a definitive copyright statement in README.md or source files. Also optional.

[Cemu is licensed under the MPL-2.0](https://github.com/cemu-project/Cemu#license)

>   * FILES — should list locations of config files (what about portable vs non-portable?)

I'd only describe the files used in "non portable" mode, and then mention that in portable mode everything is more or less dumped in the binary directory.

jn64:
> Instead of shipping a pre-generated artefact, could you please add a `man` target to the CMake build system instead?

I'm not including the actual manpage. The 2nd commit is just for preview here.

But I can't do the CMake thing, because

1. I don't know how, and don't intend to learn, and
2. As mentioned in the first commit, this would add a build dependency on pandoc (or whatever intermediate format chosen), which should be decided on first.


> Cemu is licensed under the MPL-2.0

I know, I was asking for an existing statement (if any) like `Copyright (c) XXXX-YYYY Cemu McCemuface`

Tachi107:
Il giorno ven 9 dic 2022 alle 08:57:06 -08:00:00, jn64 
***@***.***> ha scritto:
> I'm not including the actual manpage. The 2nd commit is just for 
> preview here.
> 

Oh ok, I wasn't sure about that, thanks for the clarification :)

> But I can't do the CMake thing, because
> 
> I don't know how, and don't intend to learn, and As mentioned in the 
> first commit, this would add a build dependency on pandoc (or 
> whatever intermediate format chosen), which should be decided on 
> first.

No worries, I can do it instead (either in a follow-up patch or by 
posting a diff here). Also, the intermediate format processor chosen 
doesn't have to be a hard build dependency; I'd first check if pandoc 
(or whatever) is available, and add the man target only if it is.

-- 

jonbarrow:
> I don't know what the --act-url argument "URL prefix for account server" is (protocol? hostname or part of hostname? fully formed URL?)

It's supposed to be a fully formed URL pointing to the NNID API server, such as `https://account.nintendo.net`

`--ecs-url` also exists for a similar purpose, for pointing to the eCommerce SOAP server + path, such as `https://ecs.wup.shop.nintendo.net/ecs/services/ECommerceSOAP`,  which used for getting title tickets and other metadata used during title installation

https://github.com/cemu-project/Cemu/blob/cd3fe0685faba2e631dcc7a1428f085c9fc29802/src/config/LaunchSettings.cpp#L179-L187

Though last time I checked these arguments haven't been working for a long time (since before Cemu 2.0 even came out), and seem to have been replaced with other config options

https://github.com/cemu-project/Cemu/blob/cd3fe0685faba2e631dcc7a1428f085c9fc29802/src/config/LaunchSettings.cpp#L279-L282

PR #572, opened 2022-12-12
Linux: Add CMake find module for wayland + make wayland optional

### LINKED ISSUE #571: gdk/gdkwayland.h
As of v2.0.19 it is no longer possible to build cemu on gentoo. previous builds were succesfull.

Buid stage fails:

/Cemu/src/gui/../gui/helpers/wxWayland.h:6:10: fatal error: 'gdk/gdkwayland.h' file not found
#include <gdk/gdkwayland.h>

Can this be fixed ?

SSimco:
This PR should fix #571 and also find `wayland-client.h` if it's not at the default location.


cmake/FindWayland.cmake
```diff
@@ -0,0 +1,14 @@
...
+if(PKG_CONFIG_FOUND)
+    pkg_check_modules(WAYLAND_CLIENT REQUIRED IMPORTED_TARGET wayland-client)
+
+    if(WAYLAND_CLIENT_FOUND)
+        find_path(WAYLAND_CLIENT_INCLUDE_DIR NAMES wayland-client.h HINTS ${WAYLAND_CLIENT_INCLUDE_DIRS})
+        add_library(Wayland::client ALIAS PkgConfig::WAYLAND_CLIENT)
+    endif()
+
+    set(WAYLAND_INCLUDE_DIR ${WAYLAND_CLIENT_INCLUDE_DIR})
+    set(WAYLAND_LIBRARIES ${WAYLAND_CLIENT_LIBRARIES})
```
Tachi107, line 12: You could add `WAYLAND_CLIENT_INCLUDE_DIR` to the list of `INTERFACE` include directories of the `Wayland::client` target, so that you don't then need to call both `target_link_libraries()` and `include_directories()` in code using the module

SSimco: Would that still be needed with the current changes?
Also, I think the include directory for `wayland-client` is already set after pkg_check_modules finds the module.

cmake/FindWayland.cmake
```diff
@@ -0,0 +1,12 @@
...
+if(PKG_CONFIG_FOUND)
+    pkg_check_modules(WAYLAND_CLIENT REQUIRED IMPORTED_TARGET wayland-client)
+
+    find_path(WAYLAND_CLIENT_INCLUDE_DIR NAMES wayland-client.h HINTS ${WAYLAND_CLIENT_INCLUDE_DIRS})
+    add_library(Wayland::client ALIAS PkgConfig::WAYLAND_CLIENT)
+
+    set(Wayland_INCLUDE_DIR ${WAYLAND_CLIENT_INCLUDE_DIR})
+    set(Wayland_LIBRARIES ${WAYLAND_CLIENT_LIBRARIES})
+    find_package_handle_standard_args(Wayland DEFAULT_MSG Wayland_LIBRARIES Wayland_INCLUDE_DIR)
+endif()
```
Tachi107, line 12: Using `REQUIRED` in the find module itself is considered bad practice; it should be specified when calling `find_package()` (and it is). Also, considering the changes made in the patch https://github.com/yuzu-emu/yuzu/pull/9395, I'd suggest changing the module as following:

```suggestion
if(PKG_CONFIG_FOUND)
    pkg_search_module(WAYLAND_CLIENT IMPORTED_TARGET wayland-client)
endif()

include(FindPackageHandleStandardArgs)
find_package_handle_standard_args(Wayland
    REQUIRED_VARS
        WAYLAND_CLIENT_LIBRARIES
        WAYLAND_CLIENT_INCLUDE_DIR
    VERSION_VAR WAYLAND_CLIENT_VERSION
)

if (Wayland_FOUND)
    add_library(Wayland::client ALIAS PkgConfig::WAYLAND_CLIENT)
endif()
```

cmake/FindWayland.cmake
```diff
@@ -0,0 +1,17 @@
+find_package(PkgConfig)
+
+if(PKG_CONFIG_FOUND)
+    pkg_search_module(WAYLAND_CLIENT IMPORTED_TARGET wayland-client)
+endif()
```
Tachi107, line 5: Last suggestion: as I just mentioned in https://github.com/yuzu-emu/yuzu/pull/9395#discussion_r1047456172, you could avoid explicitly checking for `PKG_CONFIG_FOUND`; it is done by `pkg_search_module()` automatically. Sorry for being so pedantic 😅️

cmake/FindWayland.cmake
```diff
@@ -0,0 +1,17 @@
...
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(Wayland
+    REQUIRED_VARS
+        WAYLAND_CLIENT_LIBRARIES
+        WAYLAND_CLIENT_INCLUDE_DIRS
```
Tachi107, line 11: The Ubuntu job is likely failing because `wayland-client.h` is installed in the default include dir, and no `-I` flag is thus returned by pkg-config/pkgconf; as no flags are needed, `WAYLAND_CLIENT_INCLUDE_DIRS` is empty, and `find_package_handle_standard_args()` returns an error. You could simply remove `WAYLAND_CLIENT_INCLUDE_DIRS`, everything is contained in the imported target anyway.

Exzap:
Looks good to me

Issue #530, opened 2022-11-25
Mystery 'c' account in Title Manager


jn64:
Cemu version: bc10485 (latest)

## Description

In the Title Manager, when selecting a save folder, the Account dropdown is populated with 2 entries, _80000001 (default)_ and one simply named _c_.

Selecting the _c_ account and clicking *Open directory* does nothing (it works for the default account)

Both *Download Manager* and *General Settings > Accounts* only show the default account.

I have not set up any accounts or changed any account settings. I also deleted the `<Accounts>` node from `settings.xml` to make sure everything related to accounts is default.

### Title Manager with the _c_ account
![screenshot_2022-11-25_19-22-05](https://user-images.githubusercontent.com/23169302/203976103-8a1049d6-ace5-43bf-9fb7-b25cc96c32a7.png)

### General Settings > Accounts does not have the _c_ account
![screenshot_2022-11-25_19-21-15](https://user-images.githubusercontent.com/23169302/203976236-534531e8-8ced-406c-b642-ec1427a8e2c2.png)

Exzap:
Fixed by #590

PR #590, opened 2022-12-24
fix: Mystery 'c' account in Title Manager #530

lijunyu-cn:


Exzap:
Thanks for looking into this.
Using the persistent ids of the installed accounts has the downside of not showing saves with unknown ids. It's a save manager so imho it is pretty important that it displays everything it can find regardless of whether the account is known or not.

This PR also does not fix the actual bug that causes the mystery 'c'. `PopulateSavePersistentIds` does not check if the parsed directory is actually a valid 8-digit hexadecimal id. When you have a save folder layout like this:
image
The `common` folder then incorrectly gets parsed as hexadecimal which results in 0xc (displayed as just `c`). A quick fix for this would be to use a regex check if the directory name is a valid 8 digit hex id first.

Exzap:
thanks!

Issue #569, opened 2022-12-10
Cemu2.0-19 Twilight Princess HD map bug


Meggiej:
For MacBook Pro
Macos 12.4
M1 core
Twilight Princess HD 
little map missing
<img width="1280" alt="截屏2022-12-10 18 14 29" src="https://user-images.githubusercontent.com/42362117/206849639-edb0fe99-30f2-4bdf-b7cc-58e24bc1794d.png">

Exzap:
We're currently waiting for MoltenVk to add geometry shaders. In the meantime I'm closing this since there is no point in having an open issue for every single game specific problem that is caused by lack of GS support.

PR #599, opened 2022-12-29
Add macOS to README.md

UltraHDR:


Exzap:
There should be a disclaimer of sorts that informs users that the macOS build is still pretty experimental and has known issues. Maybe with a link to [https://cemu.emiyl.com/](https://cemu.emiyl.com/)

UltraHDR:
Thanks @Exzap . Do you think you or @Tillsunset can create a "State of macOS" issue similar to https://github.com/cemu-project/Cemu/issues/107 and track all the known issues such as:

* Copied from https://cemu.emiyl.com/installation
```
Flickering or invisible textures: In some games, such as Mario Kart 8 and Bayonetta 2, some textures appear to flicker or not appear. This can be fixed by setting the environment variable MVK_CONFIG_FAST_MATH_ENABLED=0 in exchange for a 5-10% performance hit.

Pipeline failure: Cemu can sometimes fail to create or compile a pipeine cache when using MoltenVK. The best solution to this currently is to delete your existing shader cache, by right-clicking on the title in Cemu and clicking "Remove shader caches".

No sound: By default, Cemu on macOS sets audio output to 0%, with no audio device set. Go into the audio settings, select an audio device, and turn up the volume.
```
* https://github.com/cemu-project/Cemu/issues/569

Thanks

Exzap:
I don't use macOS personally so I don't feel comfortable with making a "State of macOS" issue but I'll ask around.

emiyl:
I'd be happy to write some stuff up when I have some free time

PR #604, opened 2023-01-06
UI: add option to hide the icon column

SSimco:
Remove the icon column and display the icon in the name column.
Before:
Screenshot_20230106_111429
After:
Screenshot_20230106_111346

### DIFF
--- a/src/config/CemuConfig.cpp
+++ b/src/config/CemuConfig.cpp
@@ -91,6 +91,8 @@ void CemuConfig::Load(XMLConfigParser& parser)
 	game_list_style = gamelist.get("style", 0);
 	game_list_column_order = gamelist.get("order", "");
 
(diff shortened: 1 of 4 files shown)


SSimco:
> 



> Hi, since you are taking the trouble to modify this aspect of the emulator, if I may, would you accept a suggestion?
> 
> How about instead, we keep the columns unlinked and also add the "Icon" column along with the rest?
> 
> This way you could hide/show each column at will as other emulators do (eg: Dolphin and RPCS3). Thank you very much. icn

Sure, might be a better idea than the original one to merge them.
With the current changes you should be able to hide the icon column when right clicking the header, just like the rest of the columns.

Exzap:
Somehow I didn't see that you changed it from the original behavior and then this PR entered my blind spot. Anyway, it's been fine to merge for ages so doing that now.

PR #612, opened 2023-01-12
Always show "Open Cemu folder" menu item

jn64:
Don't hide it when game is launched. Easier access to logs, screenshots etc


Exzap:
thanks!

PR #622, opened 2023-01-19
Update BUILD.md

ghost:


qurious-pixel:
Put these into one PR. No need to separate.

PR #631, opened 2023-01-27
Various fixes + readback performance improvement

Exzap:
Fixes:
- Potential random crash when reading controller button state
- Many Unity based games are working again (restored missing coreinit.OSMapMemory)
- General improvements to DS VC emulation. Yoshi's Island DS is now playable, other games still freeze during menus.
- Fixed z-sorting in DS VC, so backgrounds don't get drawn above everything else now.

I also restored and tweaked an optimization that avoids expensive texture readbacks if they are redundant. I'm not sure when this broke but either way it's working again now. In DS titles this boosts performance by a factor of 5x and in BotW the FPS shouldn't drop as much anymore when triggering dynamic fire (e.g. when burning grass). Other games don't really use readback frequently enough for it to make a difference.

Demonstration of Yoshi's Island DS:

https://user-images.githubusercontent.com/13877693/215043248-817f8306-b9a9-4ee7-9447-3c09800bb2a7.mp4

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,7 @@ bin/title_list_cache.xml
 bin/debugger/*
 bin/sdcard/*
 bin/screenshots/*
+bin/dump/*
 
 !bin/shaderCache/info.txt
 bin/shaderCache/*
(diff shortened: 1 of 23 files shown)


Squall-Leonhart:
>  I'm not sure when this broke but either way it's working again now

1.25

PR #638, opened 2023-01-29
Shared libraries crash MH3U on load so disable them by default.

Squall-Leonhart:
Surprised nobody reported this yet.


Squall-Leonhart:
Ok, this isn't the right idea, this game needs the user snd related rpl's, and the issue itself is happening in swkbd.rpl.

PR #641, opened 2023-01-30
Large-scale PowerPC recompiler rework

Exzap:
**Disclaimer:** This is work-in-progress. I'm opening this draft PR for visibility, so others can track progress and know not to alter recompiler code. Work started on this in November and the ETA for completion is somewhere in the span of the next few months, depending on my motivation.

### Goals

I originally started work on the recompiler in 2014 and since then I have learned a lot more about state-of-the-art compiler and IR design. While I'm generally happy with the quality of our code translation, some of the design choices I made along the way make it hard to introduce further optimizations or fixes. A lot of the complexity is at the burden of the x86-64 backend, which means that all of that would have to be reimplemented when targeting another architecture.

Overall, the idea is to make both the front-end (PPC to IR) and the back-end (IR to x86-64) as "dumb" as possible so that all the complex logic can be shifted to operate on platform-independent IR, lowering the burden on platform-specific code.

### State

- [x] Reorganized file and folder structure to be more modular
- [x] Modernize C-style code to use C++ features where it makes sense
- [x] Fundamentally rework PPC basic block handler to be more flexible. Support non-continous functions and potentially allow for complex inlining
- [x] Support for bool-based jumps and bool registers instead of having PowerPC CR logic embedded into the IR
- [x] Allow PowerPC CR bits, SPRs and XER carry bit to reside in registers and participate in register allocation
- [x] Avoid complex instructions in the IR when they could be implemented using basic operations only. The motivation for this is that it actually simplifies optimizations and allows for emitting more efficient code than having a ton of highly-specialized instructions. Ideally 1 IR instruction = 1 host instruction
  - [x] LSWI / STWSI
  - [x] SRAW / SRAWI
  - [x] BDNZ
  - [x] LWARX / STWCX
  - [x] ADDC and other arithmetic instructions with carry
  - [x] DCBZ
  - [x] MFCR / MTCRF
  - [x] RLWIMI
- [x] Support typed registers. For now everything is either a 32bit integer or a 2x64bit paired single register
- [x] Switch floating-point logic over to the newer register allocator that is currently only used for integer registers
- [x] Improve register allocator to support target instructions with hardcoded registers (e.g. x86's `SHL reg/mem, CL`). This is currently done suboptimal by the final emitter moving registers around whenever such an instruction is encountered
- [x] Support for calls to native code in arbitrary locations of the IR program. Currently calling external code is done hackily via macro instructions which need per-backend implementation
- [ ] Rework floating-point register handling. This is a big chapter on it's own and I'll expand on this once I get to it
- [x] Optimize! This includes bringing back optimizations lost with the restructuring as well as adding some new ones
  - [x] Added a new dead code elimination pass
  - [x] x86 specific: Conditional jumps will use eflags instead of emulating PPC CRx bits where possible
  - [x] Fix loop detection and move register loads/stores out of loops where possible

I know a lot of these are pretty abstract, so in the future I might add a few before-vs-after code examples to this text.

### Will this PR add ARM support?
No. But it will make adding a new target architecture a lot easier and if I am motivated enough I'll look into adding an aarch64 backend after this is done.

### Will this make Cemu faster?
Maybe? After everything is done the recompiler should output faster code, but CPU execution speed generally isn't a bottleneck in Cemu so it's hard to predict whether there will be an actual difference.

### What about the proposed plan to use LLVM?
I did quite a bit of research on that. The biggest downside is that LLVM is still quite JIT-unfriendly and comes with significant bloat. Not saying that it wouldn't work, but the cons outweigh the pros in my opinion. Plus we already got a pretty sophisticated recompiler and it would be a waste to throw it away.
On a personal note, I enjoy working on custom solutions more than plugging in libraries so it's easier for me to stay motivated and make progress. In regards to total effort both solutions are about the same.


Wunkolo:
What would be the scope of changing the x64 emitter over to something like [xbyak](https://github.com/herumi/xbyak/)?

[With the current x64 emitter](https://github.com/cemu-project/Cemu/blob/main/src/Cafe/HW/Espresso/Recompiler/PPCRecompilerX64.h), adding a new instruction or class of instructions would involve implementing the encoding for those instructions (REX, VEX, EVEX, ModR/M, SIB, etc) from scratch and then implementing the new instruction in particular AND detecting it when this redundant work can probably just be pushed onto a proven library.

Exzap:
Thanks for pointing out Xbyak, I wasn't aware of it. The assemblers I looked at were always a bit overkill for our purposes, usually focusing on human-friendly API and less towards a simple interface for machine generated code. We only need a very thin emitter, but Xbyak seems to be exactly that.

Related to this, as part of this rework I also started a new "cleaner" x86-64 high-performance emitter which I auto-generate from encoding tables. The effort for this is relatively minimal, but using a premade emitter would certainly cut down the effort even further. I'll think about it.

amayra:
did you drop this project ?

Exzap:
Nah just busy with other stuff. I'll get back to this eventually

iMonZ:
> Nah just busy with other stuff. I'll get back to this eventually

Thanks! ARM64 Support would make the CEMU emulator finally done and future proof!

Gabezin64:
This will finally fix the lens flare issue in The Wind Waker HD and Twilight Princess HD?

Exzap:
> This will finally fix the lens flare issue in The Wind Waker HD and Twilight Princess HD?

That's a graphical issue. It's unaffected by this CPU rework.

Ntemis:
+1 for motivation practices :1st_place_medal: 
We need Aarch64 support so gbtw

Exzap:
I consider this PR complete. There is more work that can be done but it's at a good point to merge so let's do that.
But first it would be nice to get some feedback. Anyone interested in testing this please [grab the executable from github actions](https://github.com/cemu-project/Cemu/actions/runs/11563460932?pr=641) and let me know about any issues.

Here is a benchmark. Previous PPC JIT in Cemu 2.2:
image

The reworked PPC JIT from this PR:
image
(lower numbers are better) While in the benchmark some tasks are much faster, real world performance will probably remain largely the same since CPU emulation never really was a bottleneck for us. There have also been some general accuracy improvements and the top post has all the under-the-hood changes that were made.

Exzap:
@boggydigital Can you post the log for other games as well

goeiecool9999:
It's the same story on Linux. The point where the access violation happens varies but most of the time it happens at the jump instruction in this screenshot (same behaviour in different games too). Hopefully this gives you a decent clue.
image

Exzap:
I was able to get it to crash by turning off BMI2 extension. Unsure if it's directly related to your crashes but we will see. Working on a fix

Exzap:
Can you grab the latest build and check again @goeiecool9999 @boggydigital

boggydigital:
Tried [the latest build](https://github.com/cemu-project/Cemu/actions/runs/11573071586/job/32214381030). It crashes for me as well. Attaching few more logs (filename = title id):
[00050000101c9400.txt](https://github.com/user-attachments/files/17561778/00050000101c9400.txt)
[000500001010ec00.txt](https://github.com/user-attachments/files/17561779/000500001010ec00.txt)
[000500001012bd00.txt](https://github.com/user-attachments/files/17561780/000500001012bd00.txt)
[000500001012dc00.txt](https://github.com/user-attachments/files/17561781/000500001012dc00.txt)
[0005000010180600.txt](https://github.com/user-attachments/files/17561782/0005000010180600.txt)

goeiecool9999:
That fixes it :partying_face:

boggydigital:
Likewise, I can't repro the crash in any of the ~10 titles I've tried. Thank you @Exzap!

Ntemis:
utb

Ammar-Sadaoui:
what is a real bottleneck here if CPU emulation was not the problem ?

Exzap:
> what is a real bottleneck here if CPU emulation was not the problem ?

It differs by game, but for the more graphically complex games it's usually the GPU command processor.
Discussing this at full length goes outside the scope of this PR but if you are curious about Cemu's architecture the best place to learn more is [our discord](https://discord.gg/5psYsup) where we have discussions about these things and anyone can ask questions.

squidbus:
Tested on macOS with most first party titles and didn't encounter any issues compared to main.

mkrcos:
Tested on linux and most games run great. The only issue that I've found is in Mario Kart 8, it gets stuck on the loading screen after finishing the first race.

niko325c:
Runs great on most titles i've tried. Only issue there is with it that i've noticed, is the loading gets stuck in Mario Kart 8 when finishing the first race in a cup, and crashes in Wii Party U when attemting to go into any minigame (No stack trace in the log, but the crashing issue also happends on the stable 2.X builds)

useranon350:
This makes a real difference in FPS for me in BOTW, by about 10FPS. It seems to introduce some bugs in sound emulation though, adding a periodic "popping" sound in some areas.

Mauro078:
Is this PR complete or it needs more work?

Issue #651, opened 2023-02-05
Error overlays do not respond to clicks or hover while the gamepad window is open.


Squall-Leonhart:
**Description:**
ErrEula overlays become unclickable while the Gamepad screen is open, STR will  be using Monster Hunter 3 Ultimate

**STR:**
1. Start MH3U
2. Open the Gamepad View
3. Create a new Character
4. Enter more than 10 characters for the name and confirm it to provoke erreula 199-0105

**Expected**: 

Error can be clicked off without closing the gamepad.

**Actual:**
The Error appears on both windows and is unclickable.

Version: 2.0.26: https://github.com/cemu-project/Cemu/commit/9a4f945f33e57a4b45af9892647a72362123f642
Graphics API: Vulkan and OpenGL
Graphics Driver: 528.24
Graphics Pack: with and without Resolution patch
Cafelibs used: Yes
MLC01 Dump: Partial

PR #664, opened 2023-02-17
Separate imgui contexts for TV and Pad windows.

goeiecool9999:
Does what the title says and also removes some duplicate code from the cache loading screen.
Feedback very welcome.
Perhaps the context initialization code should go in the Renderer Constructor/destructor instead of Initialize and Shutdown?
Fixes: #651

### DIFF
--- a/src/Cafe/HW/Latte/Core/LatteShaderCache.cpp
+++ b/src/Cafe/HW/Latte/Core/LatteShaderCache.cpp
@@ -268,7 +268,7 @@ void LatteShaderCache_load()
 	g_shaderCacheLoaderState.loadedShaderFiles = 0;
 
 	// get game background loading image
-	auto loadBackroundTexture = [](bool isTV, ImTextureID& out)
+	auto loadBackgroundTexture = [](bool isTV, ImTextureID& out)
 	{
 		TGAFILE file{};
 		out = nullptr;
@@ -295,8 +295,8 @@ void LatteShaderCache_load()
 		}
 	};
 
-	loadBackroundTexture(true, g_shaderCacheLoaderState.textureTVId);
-	loadBackroundTexture(false, g_shaderCacheLoaderState.textureDRCId);
+	loadBackgroundTexture(true, g_shaderCacheLoaderState.textureTVId);
+	loadBackgroundTexture(false, g_shaderCacheLoaderState.textureDRCId);
 
 	sint32 numLoadedShaders = 0;
 	uint32 loadIndex = 0;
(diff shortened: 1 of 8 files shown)


src/Cafe/HW/Latte/Core/LatteShaderCache.cpp
```diff
@@ -231,48 +268,36 @@ void LatteShaderCache_load()
...
 
 	// get game background loading image
-	TGAFILE TVfile{};
-	g_shaderCacheLoaderState.textureTVId = nullptr;
-
-	std::string tvTexPath = fmt::format("{}/meta/bootTvTex.tga", CafeSystem::GetMlcStoragePath(CafeSystem::GetForegroundTitleId()));
-	sint32 statusTV;
-	auto fscfile = fsc_open(tvTexPath.c_str(), FSC_ACCESS_FLAG::OPEN_FILE | FSC_ACCESS_FLAG::READ_PERMISSION, &statusTV);
-	if (fscfile)
+	auto loadBackroundTexture = [](bool isTV, ImTextureID& out)
```
Exzap, line 271: Small typo in lambda name

src/Cafe/HW/Latte/Renderer/Renderer.cpp
```diff
@@ -37,6 +37,32 @@ bool Renderer::GetVRAMInfo(int& usageInMB, int& totalInMB) const
...
+
+	auto setupContext = [](ImGuiIO& io){
+		io.WantSaveIniSettings = false;
+		io.IniFilename = nullptr;
+	};
+
+	imguiTVContext = ImGui::CreateContext(imguiFontAtlas);
+	setupContext(ImGui::GetIO());
+	imguiPadContext = ImGui::CreateContext(imguiFontAtlas);
+	setupContext(ImGui::GetIO());
```
Exzap, line 55: `ImGui::CreateContext` only changes the current context if there is none set yet. So there should be a `SetCurrentContext` after the second `CreateContext`. I would add it to both because it's not obvious that the context is being set as a side effect of creating it.

Exzap (changes requested):

Exzap:
Can confirm that this fixes the problem where swkbd/erreula wouldn't react to input with GamePad window open.
Good job 👍

> Perhaps the context initialization code should go in the Renderer Constructor/destructor instead of Initialize and Shutdown?

The issue with that is that the base constructor runs before the actual Vulkan/OpenGL constructor, so the renderer wouldn't actually be ready yet. We don't really get around to having an Initialize method for that reason and I think it's just fine the way you ended up implementing it.

PR #674, opened 2023-02-24
Fix building with cubeb disabled

nngeorgiy:
This PR fixes Cemu builds with `ENABLE_CUBEB=OFF` by adding a couple of missing ifdefs. I've had to disable cubeb as it caused the build to fail on my Ubuntu 22.04/amd64 PC.

### DIFF
--- a/src/audio/IAudioAPI.cpp
+++ b/src/audio/IAudioAPI.cpp
@@ -6,7 +6,9 @@
 #include "DirectSoundAPI.h"
 #endif
 #include "config/CemuConfig.h"
+#if HAS_CUBEB
 #include "CubebAPI.h"
+#endif
 
 std::shared_mutex g_audioMutex;
 AudioAPIPtr g_tvAudio;
@@ -80,7 +82,9 @@ void IAudioAPI::InitializeStatic()
 	if(!s_availableApis[XAudio2]) // don't try to initialize the older lib if the newer version is available
 		s_availableApis[XAudio27] = XAudio27API::InitializeStatic();
 #endif
+#if HAS_CUBEB
 	s_availableApis[Cubeb] = CubebAPI::InitializeStatic();
+#endif
 }
 
 bool IAudioAPI::IsAudioAPIAvailable(AudioAPI api)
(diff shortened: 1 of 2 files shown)


Exzap:
Thanks!
Out of curiosity, what build errors did you encounter?

nngeorgiy:
I do a `RelWithDebInfo` build and CMake fails while spamming dozens of identical errors:
```
CMake Error in CMakeLists.txt:
  IMPORTED_LOCATION not set for imported target "cubeb::cubeb" configuration
  "RelWithDebInfo".
```

Tachi107:
@nngeorgiy did you install `libcubeb-dev` from `apt`? I maintain the cubeb Debian/Ubuntu package, and if there's something wrong there I can try to fix it.

nngeorgiy:
@Exzap @Tachi107
I've actually had `libcubeb-dev` installed, and once I removed it, [I got a warning](https://gist.github.com/nngeorgiy/1ccfd6d32e7a325e119015191b9f50f3) instead. Note the reference to `vcpkg` there. It's possible this is a vcpkg bug? Other than the warning, it built just fine with the local cubeb copy, using the same `RelWithDebInfo` mode. I'm sorry, but this is it for me: Cemu builds take over a hour on my laptop, so I can't exactly tinker with the build system.

Tachi107:
It's hard to tell what the issue could be with little context, but yeah it may be possible that vcpkg somehow conflicted with the cubeb package you had installed. I don't use vcpkg myself though, so I can't tell for sure.

Exzap:
We don't use vcpkg for cubeb. It's included as a [git submodule](https://github.com/cemu-project/Cemu/tree/main/dependencies).
iirc Cemu will always use the system cubeb over the one that is bundled with Cemu so maybe the distro package just has some tweaks that clash with our build config.

Tachi107:
I think I know what's going on. vcpkg always wraps `find_package()` calls, so even if cubeb is installed on the system it gets discovered by the vcpkg wrapper (unless you fully disable it with `-DENABLE_VCPKG=false`).

Now, the issue is that, as mentioned in https://github.com/mesonbuild/meson/issues/7632, vcpkg expects to find a `cubebTargets-RelWithDebInfo.cmake` file if you set `CMAKE_BUILD_TYPE` to `RelWithDebInfo`, even though CMake itself doesn't make guarantees about how target exports should be named (in fact, the only requirement that `find_package()` specifies is that the CMake Config file must be named either `cubebConfig.cmake` or `cubeb-config.cmake` - all the various `cubebTargets*.cmake` files are not required to be present).

As in Debian we only build packages with `CMAKE_BUILD_TYPE=None` (by default), `libcubeb-dev` only contains a `cubebTargets-none.cmake` file, which defines a `cubeb::cubeb` with the `IMPORTED_CONFIGURATIONS` target property set to `NONE` and `IMPORTED_LOCATION_NONE` set to the library file path.

So yeah, this is a bug in vcpkg in my opinion. Instead of aborting the build if `IMPORTED_LOCATION_MyBuildType` isn't defined for the discovered target, it should just fall back to using whatever variant it finds. The build type used to build an installed package isn't relevant 90% of the times, as it doesn't mess with ABI, API, etc (at least on Linux).

PR #682, opened 2023-03-02
Utilize VK_KHR_SHADER_FLOAT_CONTROLS if available

Exzap:
`VK_KHR_SHADER_FLOAT_CONTROLS` lets us set rounding behavior for floating point math in shaders to `round-to-even` instead of truncate rounding, which more closely matches console behavior. If the extension is not supported then Cemu will fall back to default rounding.

This is a bit of a risky change since it subtly affects *all* shaders and may break things in unexpected places. For this reason I'll leave this up for a bit before merging. Testing is appreciated! But make sure to clear the shader cache first.

I'm still doing some testing myself but so far one known effect is that it fixes mobs in Minecraft having misplaced body parts:
image


jn64:
Those using my Fedora Copr package [jn64/Cemu](https://copr.fedorainfracloud.org/coprs/jn64/Cemu/) can test this PR by doing:

```sh
# dnf swap Cemu Cemu-PR682
```

And to swap back to normal package:

```sh
# dnf swap Cemu-PR682 Cemu
```

PR #696, opened 2023-03-12
Check compiler for AES support

ColinKinloch:
ARM builds fail as `-maes` is not recognised.


Exzap:
I dont think we need to set this flag at all anymore. AES support is determined and handled at runtime. It should be fine to drop it completely.

Exzap:
Commit e803c6be47e8351782843aa264ef9d4da0f46d7f removes the `-maes` option

Issue #678, opened 2023-02-26
Feature Request: Show location of title in Title Manager


Crementif:
**Who**
This'd be a good issue to be fixed by someone who wants to get involved, but doesn't have much C++ or wxWidgets knowledge. 

**Why**
It's currently hard to know without opening the title entry directory whether it's pulling games/updates/DLCs from the MLC01 path or from the games path. Hence why it'd be a nice QoL improvement.

**How**
The process of fixing this would be adding a new column to the Title Manager which'd show whether the game is located from the MLC01 folder (/"installed") or whether it's found from one of the game paths.

**Extra**
Another task could be seeing how duplicate entries are being treated by the game list, but that can happen later.
  
---
To any (new or existing) developer that might look into this, thanks for wanting to help! Let me know if you need any help or have any questions if I forgot to mention something.

RivanParmar:
Hi, I would like to work on this issue! However, I have a few questions regarding this issue:
- What should be the name of the column?
- What exactly should the column show? (Like only "MLC01" and "Game Path" or something else)

Crementif:
Awesome! For a lot of the design decisions I would just experiment a bit to see what looks best. If you want some guidance/opinions: depending on the column you could go with Installed (Yes/No) or Location (Installed/Game Folders, MLC/Game Path etc.), at least those are some things I was thinking off.

RivanParmar:
I've setup the project as described in the `BUILD.md` file. I've got Visual Studio 2019 with Windows 10 SDK. However, upon building the project I'm getting the following errors:
```
Cemu\build\vcpkg_installed\x64-windows-static\include\fmt\format.h(3548,11): error C2993: 'fmt::v9::detail_exported::fixed_string<Char,N>': is not a valid type for non-type template parameter 'Str'
Cemu\build\vcpkg_installed\x64-windows-static\include\fmt\format.h(4163,61): error C3686: 'operator ""_a': literal operator template must have exactly one template parameter that is a parameter pack
Cemu\build\vcpkg_installed\x64-windows-static\include\fmt\format.h(4163,61): error C3687: 'operator ""_a': literal operator template must have non-type template parameter of type 'char'
```
In total I'm getting 66 errors that are of the same type as above but for different modules. So, could you please tell me on how I can fix these errors?

lijunyu-cn:
[...]

PR #697, opened 2023-03-12
Added location column in Title Manager

RivanParmar:
Fixes issue #678

### DIFF
--- a/src/gui/components/wxTitleManagerList.cpp
+++ b/src/gui/components/wxTitleManagerList.cpp
@@ -943,9 +943,11 @@ wxString wxTitleManagerList::GetTitleEntryText(const TitleEntry& entry, ItemColu
 	}
 	case ColumnLocation:
 	{
-		if (_pathToUtf8(entry.path).find("usr/") != std::string::npos ||
-			_pathToUtf8(entry.path).find("sys/") != std::string::npos)
-			return _("MLC01");
+		const auto relative_mlc_path = 
+			entry.path.lexically_relative(ActiveSettings::GetMlcPath()).string();
+
+		if (relative_mlc_path.starts_with("usr") || relative_mlc_path.starts_with("sys"))
+			return _("MLC");
 		else
 			return _("Game Paths");
 	}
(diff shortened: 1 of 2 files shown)


src/gui/components/wxTitleManagerList.cpp
```diff
@@ -935,6 +941,14 @@ wxString wxTitleManagerList::GetTitleEntryText(const TitleEntry& entry, ItemColu
 		return "";
 		//return wxStringFormat2("{}", entry.format);
 	}
+	case ColumnLocation:
+	{
+		if (_pathToUtf8(entry.path).find("usr/") != std::string::npos ||
+			_pathToUtf8(entry.path).find("sys/") != std::string::npos)
```
Crementif, line 947: There's technically an edge case where a user could store their games in a path that also contains "usr/" or "sys/". The following code would fix that by creating a relative path from the user's MLC folder first.
```suggestion
		const auto relative_mlc_path = entry.path.lexically_relative(ActiveSettings::GetMlcPath()).string();
		if (relative_mlc_path.starts_with("usr") || relative_mlc_path.starts_with("sys"))
```
There's also other ways you could address this issue.

src/gui/components/wxTitleManagerList.cpp
```diff
@@ -935,6 +941,14 @@ wxString wxTitleManagerList::GetTitleEntryText(const TitleEntry& entry, ItemColu
...
+			return _("MLC01");
```
Crementif, line 948: In hindsight, I think using MLC instead of MLC01 is slightly better since Cemu's general settings also call it the MLC path and not MLC01 path.
```suggestion
			return _("MLC");
```

Crementif (changes requested):
Looks good! Since it already adheres to Cemu's coding style, I have only some minor changes that I thought of after giving you my suggestions and a solution for a potential edge case. Sorry that it's kinda late again!

Crementif (approved):
Looks good! Will merge if the CI succeeds.

PR #706, opened 2023-03-22
Latte optimizations and tweaks

### LINKED ISSUE #380: BotW: high GPU load in buy/sell menus
Cemu version: f0938e1, ..., 6f985c3
OS: Fedora 36
GPU: NVIDIA GeForce GTX 1050 Ti
Driver version: 515.76, 520.56.06
Game: Breath of the Wild (EU, v208 / 1.5.0)

[log.txt](https://github.com/cemu-project/Cemu/files/9794488/log.txt)

## Description

In BotW, there is abnormally high GPU load (+100% or more over average) in buy/sell menus.

- Only affects the standard NPC buy/sell menus. Doesn't affect any similar menu I tested: inventory, dyeing shop, Great Fairy, Hateno statue
- No input lag  ~~or frame drops~~ (I've seen other issues about those in various menus, so this doesn't seem to be the same)
- FPS++ graphics pack:
   - With FPS++ (30FPS limit), the only effect is the increased GPU load.
   - Without FPS++, there is increased GPU load, *and* FPS drops from 30 to 20 in buy/sell menus (don't know if this is normal for the game).
- At higher internal res, the effect is more noticeable
   - 1280×720: 35% normal load to 75% in buy/sell menus
   - 1600×900: 40% normal load to 95% in buy/sell menus
- It's especially noticeable after ada8bbb which brought down the average load (:+1:)
- Possibly related issue on Windows (not mine): https://www.reddit.com/r/cemu/comments/xl3goc/shops_and_other_npc_interactions_drop_framerate/

## Screenshots

I'm using gamescope to upscale so the reported resolution doesn't match the image size. Has no effect on the issue.

(1) Normal GPU load (~35%):
1-normal

(2) High load in shop menus (~75%):
2-shop

(3) For comparison, lower load in inventory (~15%):
3-inventory

### Higher effect at 1600×900

(4) Normal load (~40%):
4-900p-normal

(5) Shop menus (~95%):
5-900p-shop

Exzap:
Changes:
- Reworked "fast drawing" logic. Fast draws are a special mode where the renderer can skip some of the more expensive per-drawcall checks if only a subset of states was changed. With this PR more drawcalls now qualify for the fast mode, reducing overall CPU load on the GPU emulation thread.
- Lowered amount of drawcalls that need to accumulate before LatteVk flushes the command buffer. Should decrease latency for texture readbacks.
- Use heuristic to determine the real resolution of colorbuffers. In cases where the aligned colorbuffer resolution mismatches the texture resolution (e.g. 1920x1088 vs 1920x1080) it is necessary to allocate and synchronize two textures. If we can guess the real colorbuffer resolution correctly we can avoid all the overhead of double textures, lowering VRAM requirements and texture synchronization overhead. Fixes #380

### DIFF
--- a/src/Cafe/HW/Latte/Core/LatteRenderTarget.cpp
+++ b/src/Cafe/HW/Latte/Core/LatteRenderTarget.cpp
@@ -295,6 +295,15 @@ LatteTextureView* LatteMRT::GetColorAttachmentTexture(uint32 index, bool createN
 	uint32 colorBufferHeight = pitchHeight / colorBufferPitch;
 	uint32 colorBufferWidth = colorBufferPitch;
 
+	// colorbuffer width/height has to be padded to 8/32 but the actual resolution might be smaller
+	// use the scissor box as a clue to figure out the original resolution if possible
+	uint32 scissorBoxWidth = LatteGPUState.contextNew.PA_SC_GENERIC_SCISSOR_BR.get_BR_X();
+	uint32 scissorBoxHeight = LatteGPUState.contextNew.PA_SC_GENERIC_SCISSOR_BR.get_BR_Y();
+	if (((scissorBoxWidth + 7) & ~7) == colorBufferWidth)
+		colorBufferWidth = scissorBoxWidth;
+	if (((colorBufferHeight + 31) & ~31) == colorBufferHeight)
+		colorBufferHeight = scissorBoxHeight;
+
 	bool colorBufferWasFound = false;
 	sint32 viewFirstMip = 0; // todo
 
--- a/src/Cafe/HW/Latte/Core/LatteRenderTarget.cpp
+++ b/src/Cafe/HW/Latte/Core/LatteRenderTarget.cpp
@@ -295,14 +295,33 @@ LatteTextureView* LatteMRT::GetColorAttachmentTexture(uint32 index, bool createN
 	uint32 colorBufferHeight = pitchHeight / colorBufferPitch;
 	uint32 colorBufferWidth = colorBufferPitch;
 
-	// colorbuffer width/height has to be padded to 8/32 but the actual resolution might be smaller
+	// colorbuffer width/height has to be padded to 8/32 alignment but the actual resolution might be smaller
 	// use the scissor box as a clue to figure out the original resolution if possible
+#if 0
 	uint32 scissorBoxWidth = LatteGPUState.contextNew.PA_SC_GENERIC_SCISSOR_BR.get_BR_X();
 	uint32 scissorBoxHeight = LatteGPUState.contextNew.PA_SC_GENERIC_SCISSOR_BR.get_BR_Y();
 	if (((scissorBoxWidth + 7) & ~7) == colorBufferWidth)
 		colorBufferWidth = scissorBoxWidth;
 	if (((colorBufferHeight + 31) & ~31) == colorBufferHeight)
 		colorBufferHeight = scissorBoxHeight;
+#endif
+
+	// log resolution changes if the above heuristic takes effect
+	// this is useful to find resolutions which need to be updated in gfx pack texture rules
+#if 0
+	uint32 colorBufferHeight2 = pitchHeight / colorBufferPitch;
+	static std::unordered_set<uint64> s_foundColorBufferResMappings;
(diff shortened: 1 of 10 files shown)


lijunyu-cn:
Perfect most of the time, except for looking into the distance with a telescope or camera. 
https://user-images.githubusercontent.com/60218343/227747349-5bfefe50-a14c-418f-85db-72bf813a5c21.mp4

Squall-Leonhart:
> Perfect most of the time, except for looking into the distance with a telescope or camera. https://user-images.githubusercontent.com/60218343/227747349-5bfefe50-a14c-418f-85db-72bf813a5c21.mp4

this is likely due to incompatibility with existing graphics packs

LuismaSP89:
I don't pretend to be off-topic, but since here I can see some optimizations on the graphics/vulkan's side, maybe this can be used for Cemu:

https://www.khronos.org/blog/you-can-use-vulkan-without-pipelines-today

Squall-Leonhart:
> I don't pretend to be off-topic, but since here I can see some optimizations on the graphics/vulkan's side, maybe this can be used for Cemu:

Since the effort has already been put in to have the pipeline cache, there isn't any benefit to be had here.

src/Cafe/HW/Latte/Core/LatteRenderTarget.cpp
```diff
@@ -295,6 +295,15 @@ LatteTextureView* LatteMRT::GetColorAttachmentTexture(uint32 index, bool createN
 	uint32 colorBufferHeight = pitchHeight / colorBufferPitch;
 	uint32 colorBufferWidth = colorBufferPitch;
 
+	// colorbuffer width/height has to be padded to 8/32 but the actual resolution might be smaller
+	// use the scissor box as a clue to figure out the original resolution if possible
+	uint32 scissorBoxWidth = LatteGPUState.contextNew.PA_SC_GENERIC_SCISSOR_BR.get_BR_X();
+	uint32 scissorBoxHeight = LatteGPUState.contextNew.PA_SC_GENERIC_SCISSOR_BR.get_BR_Y();
+	if (((scissorBoxWidth + 7) & ~7) == colorBufferWidth)
+		colorBufferWidth = scissorBoxWidth;
```
czak, line 303: I wonder if this is a typo and this was meant instead:

```suggestion
	if (((colorBufferWidth + 7) & ~7) == colorBufferWidth)
		colorBufferWidth = scissorBoxWidth;
```

czak (commented):
Thank you for taking the time to look into it! I was mostly interested in the [buy/sell menu bug](https://github.com/cemu-project/Cemu/issues/380) so I only tested the [last commit](8e36e719ee4df9c5563b28bdf065541f2ce5434c) on top of current `main` (i.e. b8dec03cb5841f66ba).

I definitely noticed improved performance with this fix. In the buy/sell screens I can now reach my target 48FPS no problem:

image
 
However, this fix seems to introduce another issue:

image

Unfortunately, I don't have enough experience to debug this further.
I thought it was the typo in `LatteRenderTarget.cpp`, but changing that actually made it worse :(
Happy to test any alternatives if anyone can figure this out.

Squall-Leonhart:
The issue you're seeing is because some aspects of graphics packs are not functional on this PR.

czak:
> The issue you're seeing is because some aspects of graphics packs are not functional on this PR.

Ah you're right thank you! I had the graphics pack enabled for 1080p. Turning it off resolves the issue :+1:

src/Cafe/HW/Latte/Core/LatteRenderTarget.cpp
Exzap: Not a typo

SunshineFluffy:
@Squall-Leonhart do you have a pointer on which graphics pack is causing this and how to fix it?

Squall-Leonhart:
> @Squall-Leonhart do you have a pointer on which graphics pack is causing this and how to fix it?

The graphics pack will be updated eventually, Exzap doesn't intend to merge this until 2.1, but its the base resolution pack.

Exzap:
Since the color buffer heuristic breaks graphic packs I have disabled it for now. I started updating gfx packs but it will take a while to cover them all. Once those updates are pushed we can enable the heuristic again.

PR #709, opened 2023-03-24
Disable the "Disable screen saver" option in MacOS

Tillsunset:
This option, when enabled, causes a crash in MacOS. SDL_DisableScreenSaver() requires the SDL_Video subsystem to be initialized to function properly. This causes the SDL_Events subsystem to also initialize and cause a crash elsewhere.

This PR simply disables the option box and causes the SetInhibit() function to return early.


Exzap:
We'll try updating SDL first (commit coming soon), if that doesn't work I'll merge this PR as a workaround.

Exzap:
Well it still crashes. So I guess we'll go with the workaround for now

Issue #703, opened 2023-03-19
Migrate legacy logging to new logging system


Crementif:
### Why
Currently Cemu implements two logging systems, the old one that uses the printf formatting syntax and the new one that uses the much better [fmt library](https://fmt.dev/latest/index.html). The old methods being [these](https://github.com/cemu-project/Cemu/blob/main/src/Cemu/Logging/CemuLogging.h#L120-L122) while the new system is found [here](https://github.com/cemu-project/Cemu/blob/main/src/Cemu/Logging/CemuLogging.h#L78-L106C2).

### How

This basically is just removing all the usages of the `cafeLog_log` logging methods to use `cemuLog_log` instead. The [fmt string formatting syntax](https://hackingcpp.com/cpp/libs/fmt.html#cheat-sheets) these new methods use is pretty different, but it should be straightforward otherwise. There's a lot of it however... hence why it hasn't been done yet! But it's appreciated if someone is willing to help with this. And you are free to do it in parts and create multiple pull requests. Preferred even since it'll be easier to review.

---
If you have any issues with updating some of the more complex logging usages, do let me know and I can help you convert them over. Or if you have any other issues/questions for that matter.

why-keith:
Happy to take a look at this. Any advice on how to test the updated logs? Trying to figure out how to trigger each one might be quite time consuming

Crementif:
Oh, nice to hear.

I don't think there's a great method to reproduce all of them, but I think most of them should be fairly trivial (and I'll also review them so any mistakes I'll probably catch). It's just a lot of busy work. Plus the fmt library does a lot of type checking at compile time.

If you are unsure about a particular line you could place them closer to the initialization point with some fake variables as parameters, just to make sure it's correct. As long as the type matches it should be good.

Crementif:
Btw, while changing them, if you see any "\n" at the end of the strings in one of the old methods, can you also remove them. These currently just add new lines which makes the logging look really messy too.

why-keith:
Presumably any uses of the macros defined [here](https://github.com/cemu-project/Cemu/blob/18a266155f95de398627fbab25851994198fa313/src/Cemu/Logging/CemuLogging.h#L164-L180) need to be replaced as well?

Crementif:
Yeah, those need removing. The forceLogRemoveMe_printf one doesn't have an equivalent though, but I would just leave that one specifically for now or just remove them entirely.

why-keith:
[...]

PR #710, opened 2023-03-25
switched cafeLog_logLine() with cemuLog_log()

why-keith:
Replaces direct uses of cafeLog_logLine()

Part of #703


Exzap:
Looks good to me. Thanks!

PR #553, opened 2022-12-04
Linux: Add Vulkan support for wayland

SSimco:

PR #715, opened 2023-03-26
Fix GamePad window when using vulkan with wayland

ColinKinloch:
Use the same wayland subsurface mechanism for the GamePad window as the main window. 
Fixes this bug:

| before | after |
| - | - |
| Screencast from 2023-03-22 00-54-04.webm cemu_gamepad_before_b cemu_gamepad_before_a | cemu_gamepad_after |

The window decoration is shown for a single frame whenever the FPS counter is updated


ColinKinloch:
@Exzap any thoughts on this? It's a pretty simple fix to an annoying issue.

ColinKinloch:
https://github.com/cemu-project/Cemu/pull/553#issuecomment-1340250142

> Since the pad view creates a new frame that doesn't have a menu bar, it doesn't matter on what surface it draws on.

It just sounds like the subsurface was deemed unnessisary. Maybe a different compositor or wx version.

PR #718, opened 2023-03-28
Wayland: Set app_id for icon in kde

ColinKinloch:
KDE requires [set_app_id](https://wayland.app/protocols/xdg-shell#xdg_toplevel:request:set_app_id) to be called to display the app icon.
I don't know if there's a way to set this value in wxWidgets, in Gtk it's for each window by the [GtkApplication](https://docs.gtk.org/gio/property.Application.application-id.html)
This hack set's it directly for a Gtk window.

cemu_kde
Screenshot from 2023-03-28 16-28-27

Currently it's only applied to the main window and the GamePad window but ideally it would run after `Show` for each toplevel window (`wxFrame`).

The `gdk_wayland_window_set_application_id` function was added in gtk `3.24.22`. Ubuntu 20.04 LTS supports `3.24.18` and with updates `3.24.20`. Therefore the function call will be dynamic.

### DIFF
--- a/src/gui/CemuApp.cpp
+++ b/src/gui/CemuApp.cpp
@@ -12,6 +12,10 @@
 #include "gui/helpers/wxHelpers.h"
 #include "Cemu/ncrypto/ncrypto.h"
 
+#if BOOST_OS_LINUX && HAS_WAYLAND
+#include "gui/helpers/wxWayland.h"
+#endif
+
 #include <wx/image.h>
 #include <wx/filename.h>
 #include <wx/stdpaths.h>
@@ -175,6 +179,11 @@ bool CemuApp::OnInit()
 
 	SetTopWindow(m_mainFrame);
 	m_mainFrame->Show();
+
+#if BOOST_OS_LINUX && HAS_WAYLAND
+	if (wxWlIsWaylandWindow(m_mainFrame))
+		wxWlSetAppId(m_mainFrame, "info.cemu.Cemu");
+#endif
 
 	// show warning on macOS about state of builds
 #if BOOST_OS_MACOS
--- a/src/gui/helpers/wxWayland.cpp
+++ b/src/gui/helpers/wxWayland.cpp
@@ -1,0 +1,24 @@
+#include "gui/helpers/wxWayland.h"
+
+#if BOOST_OS_LINUX && HAS_WAYLAND
+
+#include <dlfcn.h>
+
+bool wxWlIsWaylandWindow(wxWindow* window)
+{
+	GtkWidget* gtkWindow = static_cast<GtkWidget*>(window->GetHandle());
+	GdkWindow* gdkWindow = gtk_widget_get_window(gtkWindow);
+	return GDK_IS_WAYLAND_WINDOW(gdkWindow);
+}
+
+void wxWlSetAppId(wxFrame* frame, const char* applicationId)
+{
+	GtkWidget* gtkWindow = static_cast<GtkWidget*>(frame->GetHandle());
+	gtk_widget_realize(gtkWindow);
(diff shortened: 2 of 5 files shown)


src/gui/CemuApp.cpp
```diff
@@ -176,6 +180,10 @@ bool CemuApp::OnInit()
 	SetTopWindow(m_mainFrame);
 	m_mainFrame->Show();
 
+#if BOOST_OS_LINUX && HAS_WAYLAND
+	wxWlSetAppId(m_mainFrame, "info.cemu.Cemu");
```
Exzap, line 184: HAS_WAYLAND only indicates compile time support for Wayland, the actual window still might end up using X11. I recommend adding an extra check like this:
```cpp
if (wxWlIsWaylandWindow(m_mainFrame))
    wxWlSetAppId(m_mainFrame, "info.cemu.Cemu");
```
And then implement wxWlIsWaylandWindow like this:
```cpp
bool wxWlIsWaylandWindow(wxFrame* frame)
{
    GtkWidget* gtkWindow = static_cast<GtkWidget*>(frame->GetHandle());
    GdkWindow* gdkWindow = gtk_widget_get_window(gtkWindow);

    return GDK_IS_WAYLAND_WINDOW(gdkWindow);
}
```
Note that this code is untested and I wrote it off the top of my head. You could also avoid the extra function and check inside `wxWlSetAppId` directly, but personally I think conditionally calling wxWlSetAppId makes it more obvious that the window isn't always wayland baked.

ColinKinloch:
The build is on Ubuntu 20.04 LTS which is might be GTK 3.24.18 ([20.04 libs](https://packages.ubuntu.com/focal/libs/)).
Seems the function was [merged](https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/2205) 2 years ago.
I'll make the function call dynamic, but would there be any downside to bumping the build image to 22.04 LTS?

Exzap:
> The build is on Ubuntu 20.04 LTS which is might be GTK 3.24.18 ([20.04 libs](https://packages.ubuntu.com/focal/libs/)).

Yeah not sure whats going on. In the log I saw 3.0 somewhere. But the function is definitively not available. So maybe the dev package is outdated but GTK itself isn't. Granted, I only did a surface level check.

> I'll make the function call dynamic, but would there be any downside to bumping the build image to 22.04 LTS?

It would lock out anyone trying to compile on 20.04. Might also affect the base requirements of the appimage but I'm not sure whether it comes with its own GTK or not.

src/gui/helpers/wxWayland.cpp
```diff
@@ -0,0 +1,24 @@
...
+}
+
+void wxWlSetAppId(wxFrame* frame, const char* applicationId)
+{
+	GtkWidget* gtkWindow = static_cast<GtkWidget*>(frame->GetHandle());
+	gtk_widget_realize(gtkWindow);
+	GdkWindow* gdkWindow = gtk_widget_get_window(gtkWindow);
+	static auto gdk_wl_set_app_id = reinterpret_cast<void (*) (GdkWindow*, const char*)>(dlsym(nullptr, "gdk_wayland_window_set_application_id"));
+	if (gdk_wl_set_app_id)
+		gdk_wl_set_app_id(gdkWindow, application_id);
```
Exzap, line 21:
```suggestion
		gdk_wl_set_app_id(gdkWindow, applicationId);
```
instead of `application_Id`

ColinKinloch: Hah, I have that on my working tree. I guess I forgot to `git add`.

ColinKinloch: I've got nothing else to add, merge it :)

Issue #716, opened 2023-03-26
Assignable hotkey for toggling fullscreen would be really nice


Leafdomain:
When going fullscreen by pressing alt+enter, the "file not found" sound effect of Windows will be heard. This shouldn't be happening and is clearly an issue. It would be really good if this could be fixed.

It would also be really nice if the fullscreen hotkey could be assigned to something else, it's usually F11 in most emulators, and that's much more user-friendly, as it is a single button instead and thus much easier to press. Regardless of anyone's personal preference, surely a significant amount of people would prefer F11, or being able to assign it to any other key that they desire, so it would be really awesome if the fullscreen hotkey was made to be assignable.

Squall-Leonhart:
you mean the exclamation sound?

i don't experience this, something on your machine is generating an exception when going fullscreen.

Leafdomain:
Yes, the exclamation sound. I'm sure that the Cemu window is focused. And it happens on no other emulator. It seems to have to do something with Cemu. What else could possibly cause such thing anyways? I checked and it doesn't even happen while staring in a folder. It happens only while using Cemu and pressing alt+enter to go fullscreen.

cabrerahector:
Is this happening on a specific Cemu version, @Leafdomain? I tried this out (Windows 10, Cemu 2.0-29) and wasn't able to reproduce this either. Going fullscreen using alt+tab doesn't trigger any sounds at all so it might be something specific to your setup.

Leafdomain:
@Squall-Leonhart @cabrerahector
Sorry everyone, turns out I was in an older build where this was still an issue. I updated to the latest and now it's gone. Thanks for commenting.

I do still want to go ahead and be suggesting the assignable hotkey though, so I will rename the title and leave the issue up. It would be awesome if we had an option to be able to switch the hotkey to F11, or to be able to assign it to anything else that the user desires.

Silent-Remux:
Yes I would also like F11 to be fullscreen like many other applications.

Leafdomain:
@rluxv has gone ahead and implemented the feature.
Works wonderfully. Thank you so much for taking my suggestion, really appreciate it!

Closing issue.

PR #720, opened 2023-03-29
Allow user to enter fullscreen with F11

rluxv:
In regard to [this issue](https://github.com/cemu-project/Cemu/issues/716), I added the ability to also enter fullscreen mode with F11.


shinra-electric:
For macOS the convention is fn+F.

Leafdomain:
@rluxv
Works wonderfully. Thank you so much for taking my suggestion, really appreciate it!

Issue #726, opened 2023-03-29
[bug] disable screensaver in game not working


mx00010:
CEMU Version : [v2.0-31](https://github.com/cemu-project/Cemu/tree/v2.0-3)  https://github.com/cemu-project/Cemu/commit/b3180bc4a6558ce43b41f75eeeaaa62c8f93e837
OS : Windows 11 | 10.0.22621.1413

With the new release 2.0-31, a new option "Disable screen saver" has appeared in the general setting window  This option is checked by default and not editable.

So, surprisingly, with this latest release, the screen saver turn on in game. Quickly tested and this is not the case with older version (tested with 1.2.7.1)

Maybe this is related with the following pull request : https://github.com/cemu-project/Cemu/pull/497

PR #729, opened 2023-03-30
use "if" macro instead of "ifdef"

Tillsunset:
Fixes https://github.com/cemu-project/Cemu/issues/726

BOOST_OS_MACOS is defined on other platforms, but is only true when on building MacOS.

The workaround ended up affecting all platforms


Tillsunset:
Closed as duplicate of https://github.com/cemu-project/Cemu/pull/728

Issue #556, opened 2022-12-05
Default Nintendo Switch Pro Controller Button Assignment Wrong


hstriepe:
When configuring the Nintendo Switch Pro Controller and hitting add, the a <-> b and x <-> y button assignments are switched.

It is easily fixed, but may lead to confusion.

markmaci:
Also had this same issue when using a Nintendo Switch Power A Gamecube controller.

mx00010:
This seem is not an issue but the way Xinput protocol is designed.

The Xinput protocol doing a button automapping, but this one follow the layout of an Xbox controller. This explain why a <-> b and x <-> y button assignments are switched when you're using an other kind of controller.

davidhedlund:
> When configuring the Nintendo Switch Pro Controller and hitting add, the a <-> b and x <-> y button assignments are switched.
> 
I can confirm that this is true for SDL2.

davidhedlund:
> > When configuring the Nintendo Switch Pro Controller and hitting add, the a <-> b and x <-> y button assignments are switched.
> 
> I can confirm that this is true for SDL2.

By default, the **Nintendo Switch Pro Controller** layout mimics that of the **Sony DualSense**, mapping the buttons as follows:
* A: b
* B: a
* X: y
* Y: x

It has to be reconfigured to actually work for Nintendo Switch Pro Controller:
* A: a
* B: b
* X: x
* Y: y

PR #741, opened 2023-04-02
fix: wrong key mapping for NS Pro controller

### LINKED ISSUE #327: Buttons Are Wrong Mapped & Unchangeable - 8Bitdo Pro 2 in S Mode with SDLController
I'm using Release 2.0 of Cemu, I have an 8Bitdo Pro 2. _Before anything I need to teach you what exactly an 8Bitdo Pro 2 controller is and what it's functions are._ This is also useful because I think you should add this controller to [Motion Controls Wiki Page](https://wiki.cemu.info/wiki/Motion_controls) under Full-motion solutions.  The 8Bitdo Pro 2 has a 4-option mode switch on the backside of the controller. These modes known as S-A-D-X work as follows:

S is for the controller to function as a Switch Pro Controller
A is for the controller to function as a Mac OSX Controller
D is for the controller to function with Direct Input
X is for the controller to function with X Input

This controller has built-in rumble motors, gyroscope, and accelerometer. I believe the gyroscope and accelerometer only function while the controller is in S (Switch) mode because X Input and Direct Input wouldn't know what to do with that information. The controller can sync wirelessly to a PC bluetooth in A-D-X modes, but it **can not** sync wirelessly to a PC in S mode. So in order for me to use this controller in S mode on my PC, I had to plug it in directly through USB.

The 4 face button layout on an 8Bitdo Pro 2 matches exactly what you see on any 4-button Nintendo controller. The top button is X, left button is Y, right button is A, bottom button is B. When I play PC games, which is typically wireless through X Input mode, I always have in mind that the Xbox controller button prompts swap A-B and X-Y. Finally, I can talk about the issue.

---

When using Cemu 2.0 and in order to get motion controls to work for me, I needed to set my 8Bitdo Pro 2 to Switch mode and plug it in to my PC through USB. Then in Cemu, I open up _Options->Input Settings_, switch over _Emulated controller_ from **Wii U Pro Controller** to **Wii U Gamepad**, push the minus button after the controller selection drop down box. push the plus button to bring up the API and Controller selection tooltip, select API to be **SDLController** and then select Controller to be **Nintendo Switch Pro Controller**. I pushed the _Add_ button, then I pushed the _Settings_ button. In the _Controller settings_ sub window, I checkmarked **Use motion**. Having done all of that I could now use motion controls in Cemu.

Now this is where the bug is happening. Once I am using **Nintendo Switch Pro Controller (SDLController)**, I can no longer edit the controller mapping. All the button mapping boxes are greyed out, The default controller mapping is wrong. It has A set to b, B set to a, X set to y, Y set to x. Please change the default mapping to be 1-1. Assuming that someone did plug in an actual Switch Pro Controller and not an 8Bitdo Pro 2, those default unchangeable button mapping would still be wrong because the bottom button has always been B on Nintendo controllers

### LINKED ISSUE #739: [bug] wrong input mapping when use Nintendo Switch Pro in macOS
image

you can see, `X Y` and `A B` has wrong mapping.

meritozh:
problem described in https://github.com/cemu-project/Cemu/issues/739

Fix #739 
Fix #327

guess #556 is same, but I don't have Linux device, so cannot test XInput protocol.

### DIFF
--- a/src/input/emulated/ClassicController.cpp
+++ b/src/input/emulated/ClassicController.cpp
@@ -174,6 +174,39 @@ bool ClassicController::set_default_mapping(const std::shared_ptr<ControllerBase
 				{kButtonId_StickR_Right, kRotationXP},
 			};
 		}
+		else if (sdl_controller->get_guid() == SDLController::kSwitchProController)
+		{
+			mapping =
+			{
+				{kButtonId_A, kButton0},
+				{kButtonId_B, kButton1},
+				{kButtonId_X, kButton2},
+				{kButtonId_Y, kButton3},
+
+				{kButtonId_L, kButton9},
+				{kButtonId_R, kButton10},
+				{kButtonId_ZL, kTriggerXP},
+				{kButtonId_ZR, kTriggerYP},
+
+				{kButtonId_Plus, kButton6},
+				{kButtonId_Minus, kButton4},
+
+				{kButtonId_Up, kButton11},
+				{kButtonId_Down, kButton12},
+				{kButtonId_Left, kButton13},
+				{kButtonId_Right, kButton14},
+
+				{kButtonId_StickL_Up, kAxisYN},
+				{kButtonId_StickL_Down, kAxisYP},
+				{kButtonId_StickL_Left, kAxisXN},
+				{kButtonId_StickL_Right, kAxisXP},
+
+				{kButtonId_StickR_Up, kRotationYN},
+				{kButtonId_StickR_Down, kRotationYP},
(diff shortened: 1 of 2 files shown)


SuperSamus:
For SDL, instead of manually switching the values of the Switch Pro Controller, the SDL hint `SDL_HINT_GAMECONTROLLER_USE_BUTTON_LABELS` could be used.
Documentation here: https://github.com/libsdl-org/SDL/blob/5b9fdc9dd12564b54612ef7bd45d4b119cb18f03/include/SDL3/SDL_hints.h#L563

PR #760, opened 2023-04-12
[CI] Use System GLM

qurious-pixel:
Uses system GTK3 and GLM for linux.
Fixes crash when opening wxwidget file pickers.

Requires removing cached vcpkg GTK3 module

GTK3 override used to enable wayland build. `-DENABLE_WAYLAND=OFF` does not need override

Possible to upgrade SDL to 2.26.5 for third-party controller hotfixes.


Exzap:
If I understand the changes correctly, this PR adds a dummy GTK3 overlay port which doesn't actually install anything. So vcpkg should then fall back to system GTK? Did you run into issues after #747 because that should have already made it so that we use system GTK3, or did you just want a more robust solution?

On a side note, I cannot replicate the file picker crashes on my Ubuntu 20.04 machine. Could you post a stack trace? I'm curious about where exactly it crashes

qurious-pixel:
This uses the vcpkg GTK3, but with the wayland flag set.
There was a conflict with the libpng 1.6.39 that is being pulled from some library (likely fontconfig).

The override is not necessary when building with wayland off.

The crash is from libpng. 
Failure to launch is somehow related to GLM.

Also added an Environmental Variable since fontconfig changed the behavior of finding the default font path.

qurious-pixel:
Cemu_wx_filepicker_crash
[Cemu-2.0-32_appimage_crash.log](https://github.com/cemu-project/Cemu/files/11214279/Cemu-2.0-32_appimage_crash.log)
[Cemu-2.0-32_binary_crash.log](https://github.com/cemu-project/Cemu/files/11214281/Cemu_2.0-32_binary.log)
[Cemu-2.0-32_binary_gdb.log](https://github.com/cemu-project/Cemu/files/11214280/Cemu-2.0-32_binary_gdb.log)

[vcpkg_GLM.log](https://github.com/cemu-project/Cemu/files/11214282/vcpkg_GLM.log)

vcpkg_GLM

Tachi107:
This should also probably update the build instructions, as libglm-dev (or equivalent) is now required

qurious-pixel:
Ubuntu 20.04/22.04 and Fedora 38 Beta work with the change, but Manjaro 22 gnome and Mint 20 were having gtk/gdk pixbuf errors.
Still trying to identify those crashes.

Build Instructions will be updated for `libglm` requirement

Exzap:
Ended up cannibalizing your changes into #772 since I wanted to push a new release asap.
Thanks for looking into this!

PR #767, opened 2023-04-14
Logging migration (inputLog_printf)

why-keith:
Converts all instances of the inputLog_printf() logging method with the newer cemuLog_log().

Part of #703

### DIFF
--- a/src/Cafe/OS/libs/vpad/vpad.cpp
+++ b/src/Cafe/OS/libs/vpad/vpad.cpp
@@ -205,7 +205,7 @@ namespace vpad
 
 	void VPADSetAccParam(sint32 channel, float radius, float sensitivity)
 	{
-		inputLog_printf("VPADSetAccParam(%d, %f, %f)", channel, radius, sensitivity);
+		cemuLog_log(LogType::InputAPI, "VPADSetAccParam({}, {}, {})", channel, radius, sensitivity);
 		vpadbreak();
 		g_vpad.controller_data[channel].acc_param.radius = radius;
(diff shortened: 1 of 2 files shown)


src/Cafe/OS/libs/vpad/vpad.cpp
```diff
@@ -1017,7 +1017,8 @@ void vpadExport_VPADSetGyroDirection(PPCInterpreter_t* hCPU)
 {
 	ppcDefineParamU32(channel, 0);
 	ppcDefineParamStructPtr(dir, VPADDir, 1);
-	inputLog_printf("VPADSetGyroDirection(%d, <<%f, %f, %f>, <%f, %f, %f>, <%f, %f, %f>>)", channel, (float)dir->x.x, (float)dir->x.y, (float)dir->x.z
+	cemuLog_log(LogType::InputAPI, "VPADSetGyroDirection({}, <<{:f}, {:f}, {:f}>, <{:f}, {:f}, {:f}>, <{:f}, {:f}, {:f}>>)", channel
```
Crementif, line 1020: Is there a reason you used `{:f}` here instead of `{}`. It doesn't matter according to the fmt docs so you don't have to really change it. Either way works

Crementif (approved):
Looks good to me!

src/Cafe/OS/libs/vpad/vpad.cpp
why-keith: None reason in particular. Probably just how the first example I came across did it

Issue #781, opened 2023-04-19
clang 16 build (Fedora 38) crashes when loading any game with multi-core recompiler


jn64:
I know non-vcpkg builds aren't supported, but this is way beyond my knowledge so I'm hoping to get more eyeballs on it.

Cemu version: https://github.com/cemu-project/Cemu/commit/e3e167b8ba31a0d4be79fbe226d416a7154c20c9, https://github.com/cemu-project/Cemu/commit/f48ad6a1ca13d1abebd2c3b1f789bbe10f6fff1a (latest)

## Problem

Cemu built with clang 16 on Fedora 38 crashes on loading any game when the CPU mode is set to _Multi-core recompiler_ (but works with both _Single-core recompiler_ and _Single-core interpreter_).

compiler | built on | result |
--|--|--
clang 15 | F37 | Works (on both F37 and F38)
clang 16 | F38 | Crashes with multi-core recompiler
gcc 13 | F38 | Works

Games tested were BotW, Bayonetta 2, and Xenoblade Chronicles X. All graphics packs off. Vulkan, x11.

## Logs

Cemu logs don't show much:

- [clang 16 build, BotW, crash with multi-core recompiler](https://github.com/jn64/Cemu-rpm/files/11269492/Cemu-botw-multi.log)
  > ```
  > [14:35:00.258] Recompiler initialized
  > Error: signal 11:
  > Cemu(+0x179d2a) [0x55c2a1211d2a]
  > /lib64/libc.so.6(+0x3dba0) [0x7f5f9c25eba0]
  > Cemu(+0x179d2a) [0x55c2a1211d2a]
  > Cemu(+0x24b4cd) [0x55c2a12e34cd]
  > Cemu(+0x2c8f68) [0x55c2a1360f68]
  > /lib64/libc.so.6(+0x54e10) [0x7f5f9c275e10]
  > ```
- [clang 16 build, BotW, working with single-core recompiler](https://github.com/jn64/Cemu-rpm/files/11269494/Cemu-botw-single.log)

## Workaround

Build with gcc on F38. [My Fedora Copr package](https://copr.fedorainfracloud.org/coprs/jn64/Cemu/) will use gcc on F38 for now, starting with `2.0^20230417gite3e167b-2`. I'll continue to test every build with clang 16.

jn64:
Building with gcc on F38 works. So it is something about clang 16 (or Fedora's implementation) that causes this issue, compared to clang 15 on F37.

[...]

Issue #779, opened 2023-04-18
Cemu 2.0-34 fails to run on Fedora 38


MateusRodCosta:
OS: Fedora Silverblue 38
Cemu version: 2.0-34

Steps to reproduce:

1. Download both 2.0-33 and 2.0-34
2. Try to run 2.0-33
3. Notice everything works fine
4. Try to run 2.0-34
5. Notice how it never opens

If you run 2.0-34 via Terminal you will have the following output:

```
$ ./Cemu-2.0-34-x86_64.AppImage 
Gtk-Message: 11:32:51.304: Failed to load module "canberra-gtk-module"
Gtk-Message: 11:32:51.306: Failed to load module "canberra-gtk-module"
**
Gdk:ERROR:../../../../gdk/gdkcairo.c:197:gdk_cairo_surface_paint_pixbuf: assertion failed: (cairo_image_surface_get_width (surface) == gdk_pixbuf_get_width (pixbuf))
Bail out! Gdk:ERROR:../../../../gdk/gdkcairo.c:197:gdk_cairo_surface_paint_pixbuf: assertion failed: (cairo_image_surface_get_width (surface) == gdk_pixbuf_get_width (pixbuf))
Aborted!
Error: signal 6:
/lib64/libc.so.6(+0x8eb94) [0x7f398c363b94]
/lib64/libc.so.6(+0x3dba0) [0x7f398c312ba0]
/lib64/libc.so.6(+0x8eb94) [0x7f398c363b94]
/lib64/libc.so.6(gsignal+0x1e) [0x7f398c312aee]+0x1e) [0x7f398c312aee]
/lib64/libc.so.6(abort+0xdf) [0x7f398c2fb87f]+0xdf) [0x7f398c2fb87f]
/tmp/.mount_Cemu-2McLnOK/usr/bin/../lib/libglib-2.0.so.0(+0x1db43) [0x7f398b47db43]
/tmp/.mount_Cemu-2McLnOK/usr/bin/../lib/libglib-2.0.so.0(g_assertion_message_expr+0x4f) [0x7f398b4dacef]+0x4f) [0x7f398b4dacef]
/tmp/.mount_Cemu-2McLnOK/usr/bin/../lib/libgdk-3.so.0(+0x31c11) [0x7f398b920c11]
[...]

Issue #763, opened 2023-04-12
Cemu Fedora crashes when accessing menu


leifliddy:
I compiled Cemu in Fedora37 Container:
https://github.com/leifliddy/podman-cemu-builder

The issue is that the Cemu application will launch, but if I left click on any of the top menu items (File, Options, Tools...etc) or if I right click on any game title -- it crashes with the following. 

```
Fontconfig error: Cannot load default config file: No such file: (null)

(Cemu_release:3811271): Gtk-CRITICAL **: 23:20:19.274: gtk_box_gadget_distribute: assertion 'size >= 0' failed in GtkScrollbar
Segmentation fault!
Error: signal 11:
/lib64/libc.so.6(+0x16dfe3) [0x7fddd01bafe3]
/lib64/libc.so.6(+0x3cb20) [0x7fddd0089b20]
/lib64/libc.so.6(+0x16dfe3) [0x7fddd01bafe3]
./Cemu_release() [0x11b5dd4]
./Cemu_release(png_read_update_info+0x1f) [0x11a3fdf]+0x1f) [0x11a3fdf]
/lib64/libgdk_pixbuf-2.0.so.0(+0x1db07) [0x7fddd0698b07]
/lib64/libgdk_pixbuf-2.0.so.0(+0x20a62) [0x7fddd069ba62]
/lib64/libpng16.so.16(png_process_data+0x836) [0x7fddcfd8d546]+0x836) [0x7fddcfd8d546]
/lib64/libgdk_pixbuf-2.0.so.0(+0x1d109) [0x7fddd0698109]
/lib64/libgdk_pixbuf-2.0.so.0(+0xdc2d) [0x7fddd0688c2d]
/lib64/libgdk_pixbuf-2.0.so.0(gdk_pixbuf_loader_close+0x175) [0x7fddd068dd15]+0x175) [0x7fddd068dd15]
/lib64/libgdk_pixbuf-2.0.so.0(+0x136d3) [0x7fddd068e6d3]
/lib64/libgdk_pixbuf-2.0.so.0(gdk_pixbuf_new_from_stream+0x30) [0x7fddd068e860]+0x30) [0x7fddd068e860]
/lib64/libgtk-3.so.0(+0x13aec4) [0x7fddd093aec4]
/lib64/libgtk-3.so.0(+0x13b145) [0x7fddd093b145]
/lib64/libgtk-3.so.0(+0x12eadd) [0x7fddd092eadd]
[...]

Issue #722, opened 2023-03-29
Can't run latest Linux binaries 


lodriguez:
I'm on the latest Arch Linux using KDE plasma, Wayland, amdgpu

If I try to run the 2.0-30 AppImage, I get (last two lines keep repeating): 
```
Gtk-Message: 16:21:34.616: Failed to load module "colorreload-gtk-module"
Gtk-Message: 16:21:34.616: Failed to load module "window-decorations-gtk-module"

(AppRun.wrapped:228320): GLib-GObject-WARNING **: 16:21:34.638: specified instance size for type 'PangoCairoFcFontMap' is smaller than the parent type's 'PangoFcFontMap' instance size

(AppRun.wrapped:228320): GLib-GObject-CRITICAL **: 16:21:34.638: g_type_add_interface_static: assertion 'G_TYPE_IS_INSTANTIATABLE (instance_type)' failed

(AppRun.wrapped:228320): GLib-CRITICAL **: 16:21:34.638: g_once_init_leave: assertion 'result != 0' failed

(AppRun.wrapped:228320): GLib-GObject-CRITICAL **: 16:21:34.638: g_object_new_with_properties: assertion 'G_TYPE_IS_OBJECT (object_type)' failed

(AppRun.wrapped:228320): Pango-CRITICAL **: 16:21:34.638: pango_font_map_create_context: assertion 'fontmap != NULL' failed

(AppRun.wrapped:228320): Pango-CRITICAL **: 16:21:34.638: pango_cairo_context_set_font_options: assertion 'PANGO_IS_CONTEXT (context)' failed

(AppRun.wrapped:228320): GLib-GObject-CRITICAL **: 16:21:34.638: g_object_get_qdata: assertion 'G_IS_OBJECT (object)' failed

(AppRun.wrapped:228320): GLib-GObject-CRITICAL **: 16:21:34.638: g_object_replace_qdata: assertion 'G_IS_OBJECT (object)' failed
```

If I'm using the 2.0-30 Ubuntu build, I get:
```
(cemu:250725): GLib-GObject-CRITICAL **: 16:44:29.805: cannot register existing type 'PangoFontMap'

(cemu:250725): GLib-GObject-CRITICAL **: 16:44:29.805: cannot add private field to invalid (non-instantiatable) type '<invalid>'

[...]

PR #782, opened 2023-04-19
[LINUX] Build Source Libs

qurious-pixel:
Builds Linux libs from source rather than VCPKG.
Caches the builds to speed-up subsequent builds.

Fixes [779](https://github.com/cemu-project/Cemu/issues/779), [763](https://github.com/cemu-project/Cemu/issues/763), [722](https://github.com/cemu-project/Cemu/issues/722)

**EDIT:** Added easier VCPKG build with Conditional switch

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,6 +17,11 @@ env:
 jobs:
   build-ubuntu:
     runs-on: ubuntu-20.04
+    env:
+      CACHEDIR: "${{github.workspace}}/dependencies/linux_submodules"
+      CC: clang-12
+      CXX: clang++-12
+      VCPKG: OFF
     steps:
     - name: "Checkout repo"
       uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


jn64:
I'm not sure how this fixes #781. My Fedora Copr builds would not be affected by changes to the github builds.

qurious-pixel:
> I'm not sure how this fixes #781. My Fedora Copr builds would not be affected by changes to the github builds.

I may have jumped to quickly on the source of the error, but those were the error messages printed on console for Fedora 38 that I had prior to this PR build.

Tachi107:
I'm not sure this is an appropriate fix for Cemu. Not only this works around the vcpkg package manager used by the CMake build system, but this avoids the build system completely by running a single script. Talking about the script, installing things into `/usr` without involving the system package manager might break the system.

I appreciate your efforts in fixing things for Linux users, but what I'm trying to say is that maybe this isn't the right approach; if there are issues in vcpkg's packages, they should be fixed there.

qurious-pixel:
Superceded by https://github.com/cemu-project/Cemu/pull/784

Issue #790, opened 2023-04-22
[Feature Request] Add integrated Feral Interactive GameMode support for Linux


deadmeu:
This is a feature request for an integrated performance toggle to enable the use of [Feral Interactive GameMode](https://github.com/FeralInteractive/gamemode) on Linux devices. Currently, to use gamemode users have to manually launch cemu with `gamemoderun cemu` or edit their desktop entries to have it automatically run. Having an integrated toggle would allow users to more easily enable the feature and would allow gamemode to be disabled when a game isn't currently running.

[Prism Launcher](https://github.com/PrismLauncher/PrismLauncher) is a good example of how I envision the integration to work, with a simple setting to enable/disable the feature. The Cemu Flatpak would need updating (maybe the same way the [Steam Flatpak was updated](https://github.com/flathub/com.valvesoftware.Steam/pull/384) for gamemode support).

On my system gamemode doesn't bring a ton of performance improvement (maybe an extra 1-2 fps) but could help produce smoother frame latency and may have greater effect on lower power devices such as laptops or handhelds (such as the Steam Deck) with its optimisations.

Exzap:
Implemented with #796

PR #796, opened 2023-04-24
Add GameMode support for Linux build

rluxv:
This PR would integrate [FeralInteractive's GameMode](https://github.com/FeralInteractive/gamemode) support into Cemu, adding a setting that would allow gamemode to start automatically when Cemu runs.

Mentioned in [#790](https://github.com/cemu-project/Cemu/issues/790), this could give a minor boost in performance to those with lower-spec computers.

### DIFF
--- a/src/gui/CemuApp.cpp
+++ b/src/gui/CemuApp.cpp
@@ -16,11 +16,6 @@
 #include "gui/helpers/wxWayland.h"
 #endif
 
-//GameMode support
-#if BOOST_OS_LINUX && defined(ENABLE_FERAL_GAMEMODE)
-#include "gamemode_client.h"
-#endif
-
 #include <wx/image.h>
 #include <wx/filename.h>
 #include <wx/stdpaths.h>
@@ -190,18 +185,6 @@ bool CemuApp::OnInit()
 		wxWlSetAppId(m_mainFrame, "info.cemu.Cemu");
 #endif
 
-    //GameMode support
-#if BOOST_OS_LINUX && defined(ENABLE_FERAL_GAMEMODE)
-    if(GetConfig().feral_gamemode)
-    {
-        // attempt to start gamemode
-        if(gamemode_request_start() < 0)
-        {
-            // GameMode failed to start
-            cemuLog_log(LogType::Force, "Could not start GameMode");
-        }
-    }
-#endif
 	// show warning on macOS about state of builds
 #if BOOST_OS_MACOS
 	if (!GetConfig().did_show_macos_disclaimer)
--- a/src/gui/CemuApp.cpp
+++ b/src/gui/CemuApp.cpp
@@ -17,7 +17,7 @@
 #endif
 
 //GameMode support
-#if BOOST_OS_LINUX
+#if BOOST_OS_LINUX && ENABLE_FERAL_GAMEMODE
 #include "gamemode_client.h"
 #endif
 
@@ -191,7 +191,7 @@ bool CemuApp::OnInit()
 #endif
 
     //GameMode support
-#if BOOST_OS_LINUX
+#if BOOST_OS_LINUX && ENABLE_FERAL_GAMEMODE
     if(GetConfig().feral_gamemode)
     {
         // attempt to start gamemode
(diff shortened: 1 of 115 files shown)


qurious-pixel:
image






`src/gui/CMakeLists.txt`
```
if(ENABLE_FERAL_GAMEMODE)
	target_link_libraries(CemuGui PRIVATE gamemode)
endif()
```

rluxv:
> `src/gui/CMakeLists.txt`
> 
> ```
> if(ENABLE_FERAL_GAMEMODE)
> 	target_link_libraries(CemuGui PRIVATE gamemode)
> endif()
> ```

Thanks!

src/gui/CemuApp.cpp
```diff
@@ -16,6 +16,11 @@
 #include "gui/helpers/wxWayland.h"
 #endif
 
+//GameMode support
+#if BOOST_OS_LINUX && ENABLE_FERAL_GAMEMODE
```
Exzap, line 20:
```suggestion
#if BOOST_OS_LINUX && defined(ENABLE_FERAL_GAMEMODE)
```
Imho it's more robust to check if `ENABLE_FERAL_GAMEMODE` is defined instead of checking if the value is non-zero since it's not assigned an explicit value.

src/gui/CemuApp.cpp
```diff
@@ -185,6 +190,18 @@ bool CemuApp::OnInit()
 		wxWlSetAppId(m_mainFrame, "info.cemu.Cemu");
 #endif
 
+    //GameMode support
+#if BOOST_OS_LINUX && ENABLE_FERAL_GAMEMODE
```
Exzap, line 194:
```suggestion
#if BOOST_OS_LINUX && defined(ENABLE_FERAL_GAMEMODE)
```

src/gui/GeneralSettings2.cpp
```diff
@@ -179,6 +179,13 @@ wxPanel* GeneralSettings2::AddGeneralPage(wxNotebook* notebook)
 			m_disable_screensaver->SetToolTip(_("Prevents the system from activating the screen saver or going to sleep while running a game."));
 			second_row->Add(m_disable_screensaver, 0, botflag, 5);
 
+            // Enable/disable feral interactive gamemode
+#if BOOST_OS_LINUX && ENABLE_FERAL_GAMEMODE
```
Exzap, line 183:
```suggestion
#if BOOST_OS_LINUX && defined(ENABLE_FERAL_GAMEMODE)
```

rluxv:
> Thanks for implementing this @rluxv! I haven't tested this yet but I do have some thoughts:
> 
>     * If the user doesn't have GameMode installed on their machine does the setting get disabled?
> 
>     * I would recommend changing the config name to "Enable Feral GameMode" (this is how it is commonly referred to in other programs). Given that it's Linux only, it might be a good idea to put "(Linux)" afterwards as well, or maybe in the tool tip?
> 
>     * It looks like you have some inconsistent indentation specifically in files `CemuConfig.cpp/h` and `GeneralSettings2.cpp/h`

Hey!
1. According to FeralInteractive, ["Developers can integrate the request directly into an app. Note that none of these client methods force your users to have the daemon installed or running - they will safely no-op if the host is missing." ](https://github.com/FeralInteractive/gamemode/tree/master/lib)The setting will still show up however.
2. I can change the name, and the checkbox only shows up if you are on Linux, it checks that when you open the settings window.
3. I'll fix that as well.

Thanks!

jn64:
Is `gamemode` (the daemon package on distros) actually a build dep?

Does this start gamemode when Cemu starts? It should ideally only start gamemode when a game is loaded (and in the future when Cemu can stop emulation, it should stop gamemode as well).

rluxv:
> > Yes, it starts gamemode when Cemu starts. I can update it so it starts only when you start a game but I'm not sure if that's necessary until Cemu can also stop emulation.
> 
> If you're able to update it I think it would be a good idea to make that change sooner rather than later to ensure it doesn't get accidentally forgotten about.

Sure. Working on that now.

jn64:
I tested 8b17d3e, works as expected. With this change, I guess the tooltip shouldn't say a restart is required?

rluxv:
That would be correct, I'll update that.

Exzap:
Looks good to me! Thanks for implementing this

Issue #537, opened 2022-11-28
Chinese module does not display


zzb16888:
Hello, the mac version of ecmu does not support Chinese? I added a module with a Chinese name, but it doesn't show up in ecmu. but works
![WX20221128-221009@2x](https://user-images.githubusercontent.com/51110833/204302590-4792f491-d79c-4bd5-9ff6-d79fc2b5dbb5.png)

JH8888ph:
It's probably related to the text encoding issue, or probably related to missing fonts. I also got this graphic pack, and cannot properly show Chinese. Below is one of the "rule.txt" files inside the graphic pack. You can notice that all the Chinese characters are shown in nonsense characters or symbols. 
<img width="794" alt="Screenshot 2022-11-28 at 9 11 10 AM" src="https://user-images.githubusercontent.com/117418120/204339422-87933fb1-eecc-49a4-8cb1-cc60cd8b48ce.png">

The file cannot properly open in Windows as well.
<img width="1222" alt="Screenshot 2022-11-28 at 9 12 21 AM" src="https://user-images.githubusercontent.com/117418120/204339643-653c7abe-70db-4721-861d-1c889f05c148.png">

So this is probably not the issue of CEMU or MAC OS.

zzb16888:
Mine can be read normally. It is simply that cemu cannot display

zzb16888:
<img width="616" alt="image" src="https://user-images.githubusercontent.com/51110833/204360091-760f8772-22a7-469e-a41f-c433167ac254.png">
This mod can be used normally under win

Jerry-XU1010:
Have the same problem...

Airven131:
又有解决办法没？
is here any way to solve it？

Jerry-XU1010:
还没有啊

wsw0108:
> 又有解决办法没？ is here any way to solve it？

```
cd graphicPacks
fd-find -t f rules.txt -exec some-tool-to-convert-gbk-to-utf8 {}
```

PR #802, opened 2023-04-25
GraphicsPackWindow2: Use UTF8

### LINKED ISSUE #801: [Bug Report] Graphic Packs don't accept Certain Symbols.
Hello, I would like to report a couple of problems with the Graphics Pack System:

Visual Issue:
If your Graphic Pack rules.txt file (name, path or description) contains problematic symbols, Cemu's Graphic Pack Manager will display garbled characters instead.

Usability Issue:
If your Graphic Pack path contains problematic symbols, Cemu will not let you enable that Graphic Pack.
(When you open Cemu again, you will see that your Graphic Pack has been automatically deactivated).

 Example of problematic symbols:
  · Accute Accent: ´
  · N with Tilde: ñ 
  · Upside Down Question/Exclamation Mark: ¡ ¿

 On the other hand, there are other symbols that (unlike the above), will not cause any problems in Cemu:

 Example of not problematic symbols:
  · Parentheses: ( )
  · Square Brackets: [ ]
  · Downwards Question Mark: ?
  · Downwards Exclamation Mark: !
  · Grave Accent: `

These two problems have occurred in every version I have tested, including the most recent to date: Cemu 2.0-36 (Experimental). (Commit hash: f48ad6a)

Because of this, I would like to request the correction of these problems, so that such symbols can be entered without causing issues when they are used.
Thank you very much.

Issue

goeiecool9999:
Attempt to fix #801
may also affect #537

### DIFF
--- a/src/gui/GraphicPacksWindow2.cpp
+++ b/src/gui/GraphicPacksWindow2.cpp
@@ -390,7 +390,7 @@ void GraphicPacksWindow2::LoadPresetSelections(const GraphicPackPtr& gp)
 		}
 
 		wxString categoryWxStr = wxHelper::FromUtf8(category);
-		wxString label(category.empty() ? _("Active preset") : wxString(categoryWxStr));
+		wxString label(category.empty() ? _("Active preset") : categoryWxStr);
 		auto* box = new wxStaticBox(m_preset_sizer->GetContainingWindow(), wxID_ANY, label);
 		auto* box_sizer = new wxStaticBoxSizer(box, wxVERTICAL);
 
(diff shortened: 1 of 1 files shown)


src/gui/GraphicPacksWindow2.cpp
```diff
@@ -386,28 +388,29 @@ void GraphicPacksWindow2::LoadPresetSelections(const GraphicPackPtr& gp)
 		{
 			continue;
 		}
-					
-		wxString label(category.empty() ? _("Active preset") : wxString(category));
+
+		wxString categoryWxStr = wxHelper::FromUtf8(category);
+		wxString label(category.empty() ? _("Active preset") : wxString(categoryWxStr));
```
Exzap, line 393: Small nitpick but since `categoryWxStr` is already a wxString it shouldn't be necessary to wrap it in another `wxString(...)` constructor

goeiecool9999: Good catch

PR #807, opened 2023-05-05
Resolve domains to IPs in Friends NEX

jonbarrow:
The official Nintendo servers only ever return IP addresses for hosts in NEX, but real hardware is capable of resolving domains back to IP addresses. Custom servers like the ones at Pretendo Network make use of domains rather than IP addresses, which prevents Cemu from going online

These changes allow Cemu to resolve IPs from domains when connecting to the Friends NEX server

Tested on Ubuntu 22.04

### DIFF
--- a/src/Cafe/IOSU/legacy/iosu_fpd.cpp
+++ b/src/Cafe/IOSU/legacy/iosu_fpd.cpp
@@ -181,8 +181,8 @@ namespace iosu
 				cemuLog_log(LogType::Force, "IOSU_FPD: Resolved IP for hostname {}, {}", nexTokenResult.nexToken.host, addrstr);
 
 				// start session
-				uint32 hostIp;
-				inet_pton(AF_INET, addrstr, &hostIp);
+				const uint32_t hostIp = ((struct sockaddr_in*)addrs->ai_addr)->sin_addr.s_addr;
+				freeaddrinfo(addrs);
 				g_fpd.nexFriendSession = new NexFriends(hostIp, nexTokenResult.nexToken.port, "ridfebb9", myPid, nexTokenResult.nexToken.nexPassword, nexTokenResult.nexToken.token, accountId, (uint8*)&miiData, (wchar_t*)screenName, (uint8)countryCode, g_fpd.myPresence);
 				g_fpd.nexFriendSession->setNotificationHandler(notificationHandler);
 				cemuLog_log(LogType::Force, "IOSU_FPD: Created friend server session");
--- a/src/Cafe/IOSU/legacy/iosu_fpd.cpp
+++ b/src/Cafe/IOSU/legacy/iosu_fpd.cpp
@@ -172,7 +172,11 @@ namespace iosu
 
 				const int status = getaddrinfo(nexTokenResult.nexToken.host, NULL, &hints, &addrs);
 				if (status != 0) {
+#ifdef _WIN32
+					cemuLog_log(LogType::Force, "IOSU_FPD: Failed to resolve hostname {}, {}", nexTokenResult.nexToken.host, gai_strerrorA(status));
+#else
 					cemuLog_log(LogType::Force, "IOSU_FPD: Failed to resolve hostname {}, {}", nexTokenResult.nexToken.host, gai_strerror(status));
+#endif
 					return;
 				}
 
--- a/src/Cafe/IOSU/legacy/iosu_fpd.cpp
+++ b/src/Cafe/IOSU/legacy/iosu_fpd.cpp
@@ -168,9 +168,7 @@ namespace iosu
 
 				// Resolve potential domain to IP address
 				struct addrinfo hints = {0}, *addrs;
-				hints.ai_family = AF_UNSPEC;
-				hints.ai_socktype = SOCK_DGRAM;
-				hints.ai_protocol = IPPROTO_UDP;
+				hints.ai_family = AF_INET;
 
 				const int status = getaddrinfo(nexTokenResult.nexToken.host, NULL, &hints, &addrs);
 				if (status != 0) {
@@ -178,15 +176,9 @@ namespace iosu
 					return;
 				}
 
-				if (addrs->ai_family != AF_INET) {
-					cemuLog_log(LogType::Force, "IOSU_FPD: Resolved IP for hostname {} not IPv4", nexTokenResult.nexToken.host);
-					return;
-				}
-
-				char addrstr[15];
-				void *ptr = &((struct sockaddr_in*)addrs->ai_addr)->sin_addr;
-				inet_ntop(addrs->ai_family, ptr, addrstr, 15);
-				cemuLog_log(LogType::Force, "IOSU_FPD: Resolved IP for hostname {}, ", nexTokenResult.nexToken.host, addrstr);
+				char addrstr[NI_MAXHOST];
+				getnameinfo(addrs->ai_addr, addrs->ai_addrlen, addrstr, sizeof addrstr, NULL, 0, NI_NUMERICHOST);
+				cemuLog_log(LogType::Force, "IOSU_FPD: Resolved IP for hostname {}, {}", nexTokenResult.nexToken.host, addrstr);
 
 				// start session
 				uint32 hostIp;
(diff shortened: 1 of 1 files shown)


src/Cafe/IOSU/legacy/iosu_fpd.cpp
```diff
@@ -165,9 +165,32 @@ namespace iosu
 				g_fpd.myPresence.isOnline = 1;
 				g_fpd.myPresence.gameKey.titleId = CafeSystem::GetForegroundTitleId();
 				g_fpd.myPresence.gameKey.ukn = CafeSystem::GetForegroundTitleVersion();
+
+				// Resolve potential domain to IP address
+				struct addrinfo hints = {0}, *addrs;
+				hints.ai_family = AF_UNSPEC;
+				hints.ai_socktype = SOCK_DGRAM;
+				hints.ai_protocol = IPPROTO_UDP;
```
Tachi107, line 173: Are these hints necessary? Why can't you simply pass a null pointer to `getaddrinfo()`?

src/Cafe/IOSU/legacy/iosu_fpd.cpp
```diff
@@ -165,9 +165,32 @@ namespace iosu
...
+				const int status = getaddrinfo(nexTokenResult.nexToken.host, NULL, &hints, &addrs);
+				if (status != 0) {
+					cemuLog_log(LogType::Force, "IOSU_FPD: Failed to resolve hostname {}, {}", nexTokenResult.nexToken.host, gai_strerror(status));
+					return;
+				}
+
+				if (addrs->ai_family != AF_INET) {
+					cemuLog_log(LogType::Force, "IOSU_FPD: Resolved IP for hostname {} not IPv4", nexTokenResult.nexToken.host);
+					return;
+				}
```
Tachi107, line 184: If only IPv4 addresses are acceptable you should probably set it in the hints field. Are IPv6 address unsupported?

src/Cafe/IOSU/legacy/iosu_fpd.cpp
```diff
@@ -165,9 +165,32 @@ namespace iosu
...
+
+				char addrstr[15];
+				void *ptr = &((struct sockaddr_in*)addrs->ai_addr)->sin_addr;
+				inet_ntop(addrs->ai_family, ptr, addrstr, 15);
```
Tachi107, line 188: You should use [`getnameinfo()`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/getnameinfo.html) to get the string representation of an address obtained by `getaddrinfo()`

```suggestion
				char addrstr[NI_MAXHOST];
				getnameinfo(addrs->ai_addr, addrs->ai_addrlen, addrstr, sizeof addrstr, NULL, 0, NI_NUMERICHOST);
```

Tachi107 (changes requested):
A few suggestions around the usage of `getaddrinfo()`.

Also, while the function is standardized and present in pretty much every OS, on Windows you need to explicitly link to the `Ws2_32` library and setup the WinSock2 stack with something like:

```cpp
WSADATA wsaData;
// Manually setting the version to 2.2 instead of using MAKEWORD
WSAStartup(0b00000010'00000010, &wsaData);
```

See the Microsoft documentation for further details: <https://learn.microsoft.com/en-us/windows/win32/api/ws2tcpip/nf-ws2tcpip-getnameinfo>

Exzap:
Using socket API directly is generally a pain in regards to cross-platform compatibility. Windows has a slightly different interface for a lot of the socket API and needs special handling. If you don't have a Windows machine for testing I can look into this.
There is also the option of using `boost::asio::ip::udp::resolver::query` instead. Although I'm not really familiar with it so I'm not sure if it's more convenient in this case.


@Tachi107 
I'm not sure why you are bringing up `WSAStartup`. It's already called in other places, including by all the network-related dependencies like libcurl. For the same reasons the winsock library is already linked.

Tachi107:
> Using socket API directly is generally a pain in regards to cross-platform compatibility. Windows has a slightly different interface for a lot of the socket API and needs special handling. If you don't have a Windows machine for testing I can look into this.

Apart from the `WSAStartup` thing and the header to include `getaddrinfo()` has been pretty much identical to what you'd find in any POSIX system, possibly because they're also standardized by [RFC3493](https://tools.ietf.org/html/rfc3493).

> @Tachi107 I'm not sure why you are bringing up `WSAStartup`. It's already called in other places, including by all the network-related dependencies like libcurl. For the same reasons the winsock library is already linked.

I'm not that familiar with Cemu's code base, and I simply wasn't aware of that, so I've mentioned it just in case :)

jonbarrow:
> Using socket API directly is generally a pain in regards to cross-platform compatibility. Windows has a slightly different interface for a lot of the socket API and needs special handling. If you don't have a Windows machine for testing I can look into this.
> 
> There is also the option of using `boost::asio::ip::udp::resolver::query` instead. Although I'm not really familiar with it so I'm not sure if it's more convenient in this case.

Ah I wasn't aware that Windows had a slightly different interface, I suppose that's why the Windows build failed? I haven't looked at that log yet as I've been in the hospital, my apologies for not checking sooner

No I do not have a Windows machine to test with, so any assistance there would be wonderful thank you

src/Cafe/IOSU/legacy/iosu_fpd.cpp
jonbarrow: > Are IPv6 address unsupported?

Real hardware only accepts IPv4 addresses, from my testing. I have not delved into Cemu's source enough to know if it accepts IPv6, but I know that the real servers only ever returned IPv4 addresses so I assume Cemu is built in a similar way

src/Cafe/IOSU/legacy/iosu_fpd.cpp
```diff
@@ -165,9 +165,24 @@ namespace iosu
...
+
+				char addrstr[NI_MAXHOST];
+				getnameinfo(addrs->ai_addr, addrs->ai_addrlen, addrstr, sizeof addrstr, NULL, 0, NI_NUMERICHOST);
+				cemuLog_log(LogType::Force, "IOSU_FPD: Resolved IP for hostname {}, {}", nexTokenResult.nexToken.host, addrstr);
+
 				// start session
 				uint32 hostIp;
-				inet_pton(AF_INET, nexTokenResult.nexToken.host, &hostIp);
+				inet_pton(AF_INET, addrstr, &hostIp);
 				g_fpd.nexFriendSession = new NexFriends(hostIp, nexTokenResult.nexToken.port, "ridfebb9", myPid, nexTokenResult.nexToken.nexPassword, nexTokenResult.nexToken.token, accountId, (uint8*)&miiData, (wchar_t*)screenName, (uint8)countryCode, g_fpd.myPresence);
```
Tachi107, line 186: Since you already have the address in a `struct addrinfo`, you could directly use that instead re-converting `addrstr` with `inet_pton()`.

Also, the `struct addrinfo` returned by `getaddrinfo()` should be freed with `freeaddrinfo()` once you don't need it anymore.

Hence, you could do something like this:

```cpp
const uint32_t hostIp = ((struct sockaddr_in*)addrs->ai_addr)->sin_addr.s_addr;
freeaddrinfo(addrs);
```

Thanks for your patience 😅

jonbarrow: No no, thank *you* for your patience

src/Cafe/IOSU/legacy/iosu_fpd.cpp
```diff
@@ -165,9 +165,28 @@ namespace iosu
...
 				g_fpd.myPresence.gameKey.titleId = CafeSystem::GetForegroundTitleId();
 				g_fpd.myPresence.gameKey.ukn = CafeSystem::GetForegroundTitleVersion();
+
+				// Resolve potential domain to IP address
+				struct addrinfo hints = {0}, *addrs;
+				hints.ai_family = AF_INET;
+
+				const int status = getaddrinfo(nexTokenResult.nexToken.host, NULL, &hints, &addrs);
+				if (status != 0) {
+#ifdef _WIN32
```
Exzap, line 175:
```suggestion
#if BOOST_OS_WINDOWS
```
Checking for `_WIN32` is fine, but we use the boost OS constants everywhere else so it'd be nice if this also follows that.

Exzap:
LGTM. Thanks!

PR #819, opened 2023-05-15
Add missing template arguments for std::forward

Smertig:
Hi!

This PR fixes missing template arguments for `std::forward` call inside `PPCCoreCallback`. 

Without this trivial fix you'll get compile-time error on `PPCCoreCallback` instantiation with enum `T` and non-empty `args...` pack.  I guess there's no such usages in current code base, that's why it didn't fire yet.


Exzap:
Template argument deduction should cover all the relevant cases but regardless it doesn't hurt to be more explicit. If you ran into issues while using the template I'd be curious to know what the exact use case was.

Thanks!

Smertig:
Thank you!

> Template argument deduction should cover all the relevant cases but regardless it doesn't hurt to be more explicit.

Template argument deduction wouldn't work in this case, because both overloads of `std::forward` have non-deducible template parameters: https://godbolt.org/z/3bPKGazE7

PR #825, opened 2023-05-19
Expose logging systems and remove deprecated logging system

Crementif:
Here's some final changes to remove the old logging system implementation now that @why-keith finished migrating everything to the new logging system.

I also removed all the calls to `cemuLog_force` since it's considered a deprecated shorthand version of `cemuLog_log(LogType::Force, ...)`.

Since all calls are done through the new logging system, we can also expose systems that only worked with the new logging system like the advanced PPC context logging and a logging window.


Exzap:
Looks good to me!

PR #836, opened 2023-05-27
Wayland: Use viewporter

ColinKinloch:
This allows the compositor to scale the vulkan surface to match its parent window during resize.

| before | after |
| --- | --- |
| Screenshot from 2023-05-27 20-53-03 | Screenshot from 2023-05-27 20-58-49 |

It also adds cmake modules from the kde project to generate wayland protocol headers (https://github.com/KDE/extra-cmake-modules.git)

Out of interest is there any plan to move away from wxWidgets?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -125,7 +125,16 @@ endif()
 if (UNIX AND NOT APPLE)
 	find_package(X11 REQUIRED)
 	if (ENABLE_WAYLAND)
-		find_package(Wayland REQUIRED)
+		find_package(Wayland REQUIRED Client)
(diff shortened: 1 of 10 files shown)


Exzap:
> Out of interest is there any plan to move away from wxWidgets?

Redoing all the UI would be quite the massive undertaking. If anyone is up for it they are free to get it started, but personally I'm quite content with wxWidgets despite it's quirks.

jn64:
This adds a new build dependency on `wayland-protocols-devel` on Fedora.

amayra:
i reallly wish if cemu go with QT instead of wxWidgets

PR #854, opened 2023-06-08
Update debug menu text for frame profiler

ErikSwan:
Updates the debug menu text to note that the "Show frame profiler" option is only functional for the OpenGL backend, to avoid user confusion when attempting to enable the frame profiler when using the Vulkan backend.

This was something I personally ran into, and when asking about it in the Discord I was told by @Exzap that the Frame Profiler is only functional with the OpenGL backend, and is not functional for Vulkan.


ErikSwan:
After further discussion with @Exzap on Discord, it appears like the frame profiler is also not functional with the OpenGL backend on the latest 2.0-39 release, and it was suggested that it just be removed entirely. That's beyond the scope of contributions I feel qualified to make, but I wanted to make a note of the discussion.

I'll leave this PR as-is and let maintainers decide what they want to do with it.

PR #878, opened 2023-06-26
Fix Hermite bicubic mistake

Filoppi:
It seems like there's a mistake in the Hermite bicubic filtering implementation. My change is based on the reference sources, and it also seems to make more sense, I don't see how squaring the inverse texture size would make sense. I don't think this change was on purpose, as the variable was called `doubleSize`, not `squaredSize`.

Sources:
https://www.shadertoy.com/view/MllSzX
https://www.shadertoy.com/view/XdGXWt


Filoppi:
Replaced by https://github.com/cemu-project/Cemu/pull/879

PR #879, opened 2023-06-26
Fix multiple issues with resampling

Filoppi:
1 ) It seems like there's a mistake in the Hermite bicubic filtering implementation. My change is based on the reference sources, and it also seems to make more sense, I don't see how squaring the inverse texture size would make sense. I don't think this change was on purpose, as the variable was called doubleSize, not squaredSize.

Sources:
https://www.shadertoy.com/view/MllSzX
https://www.shadertoy.com/view/XdGXWt

2 ) `BicubicHermiteTexture` was based on the output resolution instead of the rendering resolution.

3 ) The bilinear filtering was not offet by 0.5 and it would either have wrong results, or shift the output by a pixel.

### DIFF
--- a/src/Cafe/HW/Latte/Renderer/RendererOuputShader.cpp
+++ b/src/Cafe/HW/Latte/Renderer/RendererOuputShader.cpp
@@ -48,24 +48,23 @@ vec4 cubic(float x)
 	return w / 6.0;
 }
 
-vec4 bcFilter(vec2 texcoord, vec2 texscale)
+vec4 bcFilter(vec2 uv, vec4 texelSize)
 {
-	float fx = fract(texcoord.x);
-	float fy = fract(texcoord.y);
-	texcoord.x -= fx;
-	texcoord.y -= fy;
+	vec2 pixel = uv*texelSize.zw - 0.5;
+	vec2 pixelFrac = fract(pixel);
+	vec2 pixelInt = pixel - frac;
 
-	vec4 xcubic = cubic(fx);
-	vec4 ycubic = cubic(fy);
+	vec4 xcubic = cubic(pixelFrac.x);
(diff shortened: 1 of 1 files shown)


Exzap:
Any comparison screenshots?
On a side note, I think bicubic filter doesn't work on the Vulkan renderer at all because it never got proper support for arbitrary upscaling shaders. But I may misremember

Filoppi:
> Any comparison screenshots?

Sorry, I don't, to be honest I haven't even tried this.
I just implemented all this stuff in Dolphin and spent quite a while making sure all custom resampling algorithms looked right, and that's how I noticed the ones from Cemu (which I used as an example) had some issues with them.

https://github.com/dolphin-emu/dolphin/pull/11999

I'm pretty sure about fix 1 and 3 though, you could run some screenshots yourself if you want.
I'll just leave it here, I don't really use Cemu, I just wanted to share my findings.
Thanks.

Exzap:
Ah ok, it's appreciated regardless. I will test it myself when I get a chance.

Filoppi:
I believe that working with the output/target resolution for the Hermite resampling makes it more downsampling friendly, because it picks 16 samples for each output pixel, but these 16 samples aren't aligned with the source image (the emulation output), thus you could be skipping source samples.
This was the first time I dealt with resampling so I'm still figuring things out.
Ideally, it would pick the number of samples to average based on the ratio between the source and target resolution?

Filoppi:
@Exzap my PR is now ready and tested:
https://github.com/dolphin-emu/dolphin/pull/11999
I won't continue work here but it's a good source of information if you are willing on improving resampling in cemu.

goeiecool9999:
Thank you very much for pointing these mistakes out. Many (if not all?) of your changes were used in #1392, so there's no longer any reason to keep this PR around.

PR #888, opened 2023-07-04
Fix NEX DataStore uploads (nlibcurl) and stub Miiverse PostApp

EpicUsername12:
- Fixes bugs and now allows MK8 to upload rankings!

### DIFF
--- a/src/Cafe/OS/libs/nn_olv/nn_olv.cpp
+++ b/src/Cafe/OS/libs/nn_olv/nn_olv.cpp
@@ -202,7 +202,7 @@ namespace nn
 		}
 
 		static SysAllocator<OSThread_t> s_OlvReleaseBgThread;
-		SysAllocator<uint8, 1024 * 128> s_OlvReleaseBgThreadStack;
+		SysAllocator<uint8, 1024> s_OlvReleaseBgThreadStack;
 		SysAllocator<char, 32> s_OlvReleaseBgThreadName;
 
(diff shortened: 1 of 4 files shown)


src/Cafe/OS/libs/nn_olv/nn_olv.cpp
```diff
@@ -5,6 +5,9 @@
 #include "nn_olv_DownloadCommunityTypes.h"
 #include "nn_olv_UploadFavoriteTypes.h"
 
+#include "Cafe/OS//libs/proc_ui/proc_ui.h"
+#include "Cafe/OS//libs/coreinit/coreinit_Time.h"
```
Exzap, line 9: Can you remove the double slashes

src/Cafe/OS/libs/nn_olv/nn_olv.cpp
```diff
@@ -198,10 +201,35 @@ namespace nn
 			osLib_returnFromFunction(hCPU, memory_getVirtualOffsetFromPointer(&portalAppParam->serviceToken));
 		}
 
-		uint32 UploadPostDataByPostApp(void *postParam)
+		static SysAllocator<OSThread_t> s_OlvReleaseBgThread;
+		SysAllocator<uint8, 1024 * 128> s_OlvReleaseBgThreadStack;
```
Exzap, line 205: Since the OlvRelease thread doesn't really need any stack, I recommend changing the size from ``1024 * 128` to just `1024`.

PR #890, opened 2023-07-05
Fix UTC offset calculation

EpicUsername12:
If you're ahead of UTC you should have a positive value, if you're behind you should have a negative value, so i inverted the difference calculation.

UTC+2 now returns a 7200 second offset, not -7200.

### DIFF
--- a/src/Cafe/OS/libs/nn_act/nn_act.cpp
+++ b/src/Cafe/OS/libs/nn_act/nn_act.cpp
@@ -125,6 +125,21 @@ namespace act
 			return getNNReturnCode(resultCode, actRequest);
 		}
 
+		sint64 GetUtcOffset()
+		{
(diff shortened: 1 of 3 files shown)


src/Cafe/OS/libs/nn_act/nn_act.cpp
```diff
@@ -125,6 +125,21 @@ namespace act
 			return getNNReturnCode(resultCode, actRequest);
 		}
 
+		sint64 GetUtcOffset()
+		{
+			return ((ppcCyclesSince2000 / ESPRESSO_CORE_CLOCK) - (ppcCyclesSince2000_UTC / ESPRESSO_CORE_CLOCK)) * 1'000'000;
+		}
+
+		sint32 GetUtcOffsetEx(uint64* pOutOffset, uint8 slotNo)
```
Exzap, line 133:
```suggestion
		sint32 GetUtcOffsetEx(sint64be* pOutOffset, uint8 slotNo)
```
This is probably a signed int since the offset can be negative, but more importantly it should be big endian?

EpicUsername12: That's right, will fix!

PR #894, opened 2023-07-09
Linux: Save icons to user data directory, use path to AppImage as executable path in desktop entries

### LINKED ISSUE #893: Trying to create desktop shortcut quits Cemu immediately
Using the new feature from #889, when I click Save in the file dialog for the location of the desktop file, Cemu quits immediately with no error. No desktop file or icon is created.

Location is the default given, `~/.local/share/applications`

Filename is the default given, e.g. `Breath of the Wild.desktop`

Cemu version: f1c200a01675714e51548137f2365ab2a1a9977b ([build](https://copr.fedorainfracloud.org/coprs/jn64/Cemu/build/6153874/)), Cemu-2.0-44-x86_64.AppImage
Fedora 38

capitalistspz:
Fixes #893


Squall-Leonhart:
are you handling the case where the directory doesn't exist (for some reason), Wx will not automatically create the folders when creating the file.

Issue #902, opened 2023-07-15
Skylanders portal speaker doesn’t work


TheRealWario:
It’s known that the speaker audio in skylanders trap team doesn’t work in the game, as when the speaker is supposed to play audio, the portal doesn’t light up the hole for the trap items. It is mentioned in the comparability list, but I believe the cause for this problem not mentioned is with the drivers. I remember hearing in the discord that cemu has sub-par usb driver support, and the portal has to be installed as a usb input device. I know that the process is easier with dolphin emu, where they apparently have got the portal speaker to work, by using a different driver to interface with the game. It makes using the portal tricky due to having to change the driver for each emulator every time, but I also believe the driver support is why the speaker can’t play. Hopefully development on the usb drivers may help any other usb devices in some way, too.

Exzap:
I don't remember for sure anymore but I believe the issue was due to Windows HID API not exposing the second usb device which is used for sound playback. The libusb PR likely will fix this.

slice-the-idiot:
would like to see libusb implemented for windows in this case

deReeperJosh:
Libusb has now been implemented as of https://github.com/cemu-project/Cemu/pull/1471 so this should be able to be closed

PR #903, opened 2023-07-15
nsyshid: Experimental libusb build

deReeperJosh:
Just creating this for visibility - and for any interested parties who may know more than I do. I am trying to make the nsyshid library cross platform, so that it can be used on MacOS and Linux. This is step 1 in trying to port some code that I have worked on to bring Emulated USB devices such as the Skylanders Portal/Infinity Base/Dimensions Toypad. LibUSB requires some more information in order to actually be able to send such requests though, such as bmRequest/bmRequestType, endpoint and some other bits of data. If anyone has some ideas about where to obtain this info it would be much appreciated! 

I would also like to note that I haven't implemented libusb prior to attempting this - and I am currently in the process of moving overseas so haven't got access to my Wii U at the time of opening this PR, so this may take a while.


TheRealWario:
Does this bear any relevance to my post? https://github.com/cemu-project/Cemu/issues/902

deReeperJosh:
It doesn't, but other libusb implementations (RPCS3 and Dolphin) have working Trap Team audio transfers, so the speakers should work on Cemu after this :)

src/Cafe/CMakeLists.txt
```diff
@@ -485,7 +485,10 @@ endif()
 
 set_property(TARGET CemuCafe PROPERTY MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
 
-target_include_directories(CemuCafe PUBLIC "../")
+target_include_directories(CemuCafe PUBLIC 
+  "../" 
+  ${LIBUSB_INCLUDE_DIR}
+)
```
Tachi107, line 491: Please avoid manually adding a list of include dirs and libraries like these, and use imported targets instead. I'm not sure how to patch vcpkg dependencies to add imported targets when they are not already provided, but you could always submit a patch to <https://github.com/microsoft/vcpkg/tree/master/ports/libusb>.

deReeperJosh: I was just going off of the usage for the libusb vcpkg, not sure if I would actually know what needs to change to submit a package there either but I will look in to it

LuismaSP89:
Would this libusb eventually bring us the possibility to use real wiimote support on the emulator? Just like the bluetooth passthrough on dolphin, were you can use the original bluetooth controller of the Wii instead of emulating it.

deReeperJosh:
In theory yes, but I wouldn't be able to implement that functionality myself, not yet anyway

ssievert42:
Hey, I opened #946 yesterday (have been working on and off on this and had it kind of working since may 2023), but totally missed this PR :sweat_smile:
Completely emulated devices sound like a great idea!
In #946, I've got the real Lego Dimensions portal to work reliably on linux :)
An unsolved issue are still controls transfers, but at least the Lego Dimensions portal seems to not really need them.
In theory, my implementation should be cross platform, but I've got no way to test on Windows or macOS.

My suggestion would be, that I'd restructure #946 a bit to allow implementing virtual devices, and once that is merged, you could then use that as a base for fancy emulated devices.
Do you think that requiring a device to have the following functionality would be enough?
- get vendor and product id
- get max packet sizes for RX and TX
- try to write a block of data to the device and return the bytes written
- try to read a block of data into a buffer and return the bytes read
- get descriptors as implemented in `export_HIDGetDescriptor`
- try to do control transfers with some of the arguments of `_hidSetReportSync`

deReeperJosh:
Oh that's great! I actually haven't been able to touch this work since I opened the PR, so honestly just do whatever works, I think creating some sort of Device class, that can be extended to be either a USB device or an Emulated Device would be ideal, but I will leave that up to you! And I will close off this PR, and collaborate with you as needed on your PR

PR #909, opened 2023-07-22
Vulkan: Retry instance creation if validation layer is not present

ColinKinloch:
When running the flatpak if the vulkan validation layer is enabled Cemu crashes with an error box that says "Error when initializing Vulkan renderer: Unable to create a Vulkan instance: -6"

an error box that says "Error when initializing Vulkan renderer: Unable to create a Vulkan instance: -6"

For some reason this goes away when running with the `--devel` option.

This MR adds a second attempt at creating a instance disabling all vulkan layers.

This is what [Q2RTX](https://github.com/NVIDIA/Q2RTX/blob/4f5662a7bbe70becae36b2295d956f360826ca73/src/refresh/vkpt/main.c#L892) does.

I initially tried checking against the list returned by `vkEnumerateInstanceLayerProperties` but it reports that `VK_LAYER_KHRONOS_validation` is present.

### DIFF
--- a/src/Cafe/HW/Latte/Renderer/Vulkan/VulkanRenderer.cpp
+++ b/src/Cafe/HW/Latte/Renderer/Vulkan/VulkanRenderer.cpp
@@ -350,7 +350,15 @@ VulkanRenderer::VulkanRenderer()
 	create_info.ppEnabledLayerNames = m_layerNames.data();
 	create_info.enabledLayerCount = m_layerNames.size();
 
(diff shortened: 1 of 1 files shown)


Exzap:
It is by design that Cemu refuses to start with validation layer enabled but not present. Granted, we could display a better error message. But I believe that just silently ignoring the error will lead only to confusion on the users side, especially since the validation layer needs to be turned on explicitly.

ColinKinloch:
The fact the debug options are in the menu bar rather than the settings also complicated debugging this.

Exzap:
Will merge this as-is and add error logging

PR #917, opened 2023-07-27
Quote and escape desktop entry executable path 

### LINKED ISSUE #912: ".desktop"-files created by Cemu do not work, if the path to Cemu contains spaces or some other symbols
When creating a ".desktop"-file for a game with Cemu, and the path to Cemu contains spaces, or other symbols, that need a `\` in front of them, it will not show up in the app-menu in GNOME (and probably other app-menus too).

This can be manually fixed, by modifying the "Exec=" line in the ".desktop"-files by adding `\` in front of spaces, and some other  symbols that need them, like (double) quotes.

`Exec=/home/username/Games/"Nintendo Wii U"-Emulator/Cemu/Cemu-2.0-45-x86_64.AppImage …` does not work.
This is how Cemu currently does it.

`Exec=/home/username/Games/\"Nintendo\ Wii\ U\"-Emulator/Cemu/Cemu-2.0-45-x86_64.AppImage …` does work.
This is what should be there instead.

Cemu-version: Cemu-2.0-45-x86_64.AppImage
Operating-system: "Debian GNU/Linux 12"

capitalistspz:
Hopefully fixes #912

### DIFF
--- a/src/gui/components/wxGameList.cpp
+++ b/src/gui/components/wxGameList.cpp
@@ -21,8 +21,6 @@
 #include <boost/algorithm/string.hpp>
 #include <boost/tokenizer.hpp>
 
-#include <fmt/std.h>
-
 #include "config/ActiveSettings.h"
 #include "config/LaunchSettings.h"
 #include "Cafe/TitleList/GameInfo.h"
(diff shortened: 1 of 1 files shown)


src/gui/components/wxGameList.cpp
```diff
@@ -21,6 +21,8 @@
 #include <boost/algorithm/string.hpp>
 #include <boost/tokenizer.hpp>
 
+#include <fmt/std.h>
```
Exzap, line 24: This include shouldn't be necessary?

Issue #920, opened 2023-07-29
[Linux] Cemu does not work with fmt 10


alba4k:
[fmtlib](https://github.com/fmtlib/fmt) recently saw a major release (10.0.0) that Cemu seems to be incompatible with.

updating it causes Cemu to fail to open.

trying to open Cemu just errors `cemu: error while loading shared libraries: libfmt.so.9: cannot open shared object file: No such file or directory`

Exzap:
Which distro and which binary format of Cemu are you running? (flatpak, appimage, etc.)

hirak99:
It does not also compile with libfmt 10.

Previously compiled versions may not find libfmt.so.9 if the distro has it already upgraded, e.g. Arch has moved to libfmt 10.

alba4k:
I am indeed on arch using Arch. I have Cemu from the AUR (chaotic-aur, but doesn't really matter) and fmt from the extra repo

hirak99:
@alba4k as a workaround if you still have the cemu binary compiled, you can try this -

1. Get hold of the libfmt.so.9 binary. If you have 9.1.0, a symlink `ln -s libfmt.so.9.1.0 binary libfmt.so.9` works too.
2. Make a temporary directory until Cemu is updated, and put the libfmt.so.9 there.
3. Start cemu as `LD_LIBRARY_PATH ~/my_holdback_limfmt_dir cemu`

This will give cemu access to the library without downgrading it for the whole system.

This does not resolve the issue however, compilation of cemu is broken with libfmt 10.0.

alba4k:
Just in case anybody is reading this, you did a typo, `LD_LIBRARY_PATH=/path/to/link.so.9 cemu`

anyway thanks I will try that

hirak99:
Yeah - you need an actual `libfmt.so.9.1.0` or `libfmt.so.9` for this.

Linking the libfmt.so.10 does not work.

alba4k:
I see

I will take it out of an old package then ig

jn64:
Just to note, the next Fedora version (current Rawhide, will be 39, expected 2023-10-24) also has fmt 10. I turned off Rawhide builds for my Cemu package at the moment because no time to deal with it.

kescherCode:
This is being worked on at #924.

alba4k:
fmt 10 works now

libboost doesnt anymore lol (https://github.com/cemu-project/Cemu/issues/958)

jn64:
[...]

Issue #919, opened 2023-07-28
[Linux Compile Error] `IAudioAPI.cpp`


ThisNekoGuy:
For some reason, using the same compile flags as those in some of my other recent reports, [rebuilding with Cemu via latest git commit via the AUR](https://aur.archlinux.org/packages/cemu-git) throws a build error regarding `/usr/include/fmt/core.h:2561:10` and `/usr/include/fmt/core.h:1690:3` when trying to build `IAudioAPI.cpp`

Not sure why though

<details><summary>Compile Flags:</summary>

```bash
CC=clang
CXX=clang++
AR="/usr/bin/llvm-ar"
NM="/usr/bin/llvm-nm"
AS="/usr/bin/llvm-as"
RANLIB="/usr/bin/llvm-ranlib"
OBJCOPY="/usr/bin/llvm-objcopy"
CFLAGS="-g -O3 -pipe -fno-plt -fpic -fpie -Wl,-z,relro,-z,now -w \
        -fexceptions -Wp,-D_FORTIFY_SOURCE=2 \
        -Wformat -Werror=format-security \
        -fstack-clash-protection -fstack-protector-strong -fcf-protection \
        -march=znver2 -mtune=znver2"
CXXFLAGS="${CFLAGS} -Wp,-D_GLIBCXX_ASSERTIONS"
LDFLAGS="-lc++abi -fuse-ld=lld -unwind=libunwind -pie -Wl,-O3,--sort-common,--as-needed,-z,relro,-z,now"
```
</details>

<details><summary>Log:</summary>

```make
[ 24%] Building CXX object src/audio/CMakeFiles/CemuAudio.dir/CubebInputAPI.cpp.o
In file included from <built-in>:508:
In file included from /build/cemu-git/src/build/src/audio/CMakeFiles/CemuAudio.dir/cmake_pch.hxx:5:
In file included from /build/cemu-git/src/cemu/src/Common/precompiled.h:9:
/usr/include/fmt/core.h:2561:10: error: call to deleted constructor of 'formatter<mapped_type, char_type>' (aka 'formatter<IAudioInputAPI::AudioInputAPI, char>')
  return formatter<mapped_type, char_type>().parse(ctx);
         ^
/usr/include/fmt/core.h:2620:23: note: in instantiation of function template specialization 'fmt::detail::parse_format_specs<IAudioInputAPI::AudioInputAPI, fmt::detail::compile_parse_context<char>>' requested here
[...]

PR #921, opened 2023-07-30
cmake: exclude fmt10

abouvier:
Related to https://github.com/cemu-project/Cemu/issues/919 and https://github.com/cemu-project/Cemu/issues/920.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -114,7 +114,7 @@ find_package(ZLIB REQUIRED)
 find_package(zstd MODULE REQUIRED) # MODULE so that zstd::zstd is available
 find_package(OpenSSL COMPONENTS Crypto SSL REQUIRED)
 find_package(glm REQUIRED)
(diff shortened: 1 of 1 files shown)


Exzap:
Interesting that this works. Thanks for the contribution!

Exzap:
Does this actually fix the issue on arch? I was under the impression that only fmt 10 is available now

abouvier:
No, but I created a [fmt9](https://aur.archlinux.org/packages/fmt9) package on the AUR in the meantime. But if you have both `fmt` and `fmt9` installed, CMake will pick the fmt10 version without this fix because fmt actually [accept anything](https://github.com/fmtlib/fmt/blob/757564f5cd2fa78b634dd698c63dbf069818e6fd/CMakeLists.txt#L366).

alba4k:
This is still a workaround however, as forcing an old library version is not a solution (in the long term)

PR #924, opened 2023-07-31
fmt: enable compatibility with fmt>=10

abouvier:
I have no idea what I'm doing, but this seems to compile fine against both ~~fmt9~~ and fmt10.
fmt10 basically removed implicit conversions for enums.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -114,7 +114,7 @@ find_package(ZLIB REQUIRED)
 find_package(zstd MODULE REQUIRED) # MODULE so that zstd::zstd is available
 find_package(OpenSSL COMPONENTS Crypto SSL REQUIRED)
 find_package(glm REQUIRED)
-find_package(fmt 9.1.0...<10 REQUIRED)
+find_package(fmt 9.1.0 REQUIRED)
 find_package(PNG REQUIRED)
 
 # glslang versions older than 11.11.0 define targets without a namespace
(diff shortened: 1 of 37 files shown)


abouvier:
Well, it failed spectacularly but at least it works with fmt10 :p

SSimco:
Wouldn't it be easier to add a generic formatter for enums in `Common/precompiled.h` instead of adding `using fmt::enums::format_as` in every namespace?
Something like this:
```
template <typename Enum>
requires std::is_enum_v<Enum>
struct fmt::formatter<Enum> : fmt::formatter<underlying_t<Enum>>
{
	auto format(const Enum& e, format_context& ctx) const
	{
		return formatter<underlying_t<Enum>>::format(fmt::underlying(e), ctx);
	}
};
```
And might as well update vcpkg and use fmt 10.0.0 at this point.

abouvier:
[It's what `fmt::enums::format_as` do](https://github.com/fmtlib/fmt/blob/ee475d64095091aa683664a209a9bcdd3496c743/include/fmt/format.h#L4120-L4125). But having a generic enum formatter in global namespace cause some errors with fmt9.

abouvier:
https://github.com/cemu-project/Cemu/commit/757d458161180598e606ddddc6f2a9f157e956e0

https://github.com/cemu-project/Cemu/issues/919#issuecomment-1656372888

PR #927, opened 2023-08-02
debugger: add logging breakpoint and misc fixes

Crementif:
This PR introduces logging breakpoints, something which might be helpful if you want to analyze more complex interactions in the code.  You can set these by right-clicking empty space in the breakpoint window. Some noteworthy features:
 - If you give the comment a name in the breakpoint window it'll also log it so that you can more easily identify multiple logging breakpoints.
 - If you enable advanced PPC debugging in the Debug->Logging menu you can also have it generate stack traces for each hit, which allows you to more easily see how the code ultimately ends up executing your code. Do note that the stack trace feature works best when Cemu is set to use interpreter mode.

Another "addition" is that the debugger will now store any breakpoints and comments across sessions. I'm not sure how it got broken, but at least now it doesn't lose all the comments and breakpoints you make after you restart Cemu.

Minor features/fixes include things like:
 - You can now also delete breakpoints by right-clicking them in the breakpoint window.
 - The GoTo window now prefers hex addresses, which allows you to enter non "0x" prefixed hex addresses more easily, which is especially useful for the breakpoint and memory dump windows.
 - You can now also create execution breakpoints with a manually inputted address in the breakpoint window, alongside the new logging breakpoints.
 - And some other minor fixes.

### DIFF
--- a/src/Cafe/HW/Espresso/Debugger/Debugger.cpp
+++ b/src/Cafe/HW/Espresso/Debugger/Debugger.cpp
@@ -1,5 +1,6 @@
 #include "gui/guiWrapper.h"
 #include "Debugger.h"
+#include "Cafe/OS/RPL/rpl_structs.h"
 #include "Cemu/PPCAssembler/ppcAssembler.h"
 #include "Cafe/HW/Espresso/Recompiler/PPCRecompiler.h"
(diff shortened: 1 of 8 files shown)


Exzap:
Looks good to me!

PR #945, opened 2023-08-22
[Linux/MacOS] Further Wiimote changes for parity with Windows

capitalistspz:
(Edit) Changes:
- Change Wiimote comparison so that controllers cannot be mistaken as others, so that using multiple Wiimotes is actually possible
- Use non-blocking reads to reduce input lag for extra Wiimotes
- Enable nunchuk motion
- Enable the Wiimote settings menu

Original Comment: 
> Now there should be no way that one controller is mistaken as being another connected controller

### DIFF
--- a/src/gui/input/InputSettings2.cpp
+++ b/src/gui/input/InputSettings2.cpp
@@ -976,7 +976,7 @@ void InputSettings2::on_controller_settings(wxCommandEvent& event)
 
 	case InputAPI::Keyboard: break;
 
-	#if BOOST_OS_WINDOWS
+	#if SUPPORTS_WIIMOTE
 	case InputAPI::Wiimote: {
 		const auto wiimote = std::dynamic_pointer_cast<NativeWiimoteController>(controller);
 		wxASSERT(wiimote);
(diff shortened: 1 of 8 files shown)


gamrXerus:
This pr fixes my second wii remote not being able to connect to Cemu using my standard Bluetooth dongle. The only remaining issue for me using Wiimote Api on Linux is that my Nintendo RVL-CNT-01-TR pointer lags. There is a noticeable delay when pointing and my remotes seem to disconnect randomly. My standard wii remote plus RVL-CNT-01, the pointer seems to work fine in Wii Party U when tested.

gamrXerus:
I ran Cemu with Logging turned on for the input API and this is what I get: 
wiimotelog
Not sure what is causing the disconnects. All I see is WPADProbe(0) KPADReadEx(0, 0x10) Blinking while I wait for reconnects. Hope this helps.

gamrXerus:
I can confirm now that this pr fixes my issues. Thank you! :)

gamrXerus:
The settings menu works for me now. I sincerely apologize if this is the wrong place to post this, but I noticed in my testing that Wii Sports Club, motion and pointer controls do not work.
wiisportsclub
I also noticed that the value after the zero in vpad.VPADRead(0, is changing every half second. That does not happen in Wii Party U (a game where motion and pointer controls work). I dont know if that has anything to do with it.

PR #951, opened 2023-08-30
Using reserve() for optimization insert in containers

GermanAizek:
I think you will like PR changes, it affects Debug and Release configuration.


Exzap:
Appreciate the effort but most of this falls into the category of unnecessary micro-optimizations. None of your optimizations are on performance critical code paths. In those cases code readability matters more than saving a few microseconds of allocation time.

Issue #898, opened 2023-07-11
Revisiting Save States


Spegs21:
This has been asked about before [here](https://bugs.cemu.info/issues/161) and it was mentioned that it wasn't feasible due to the storage space required but with some modern games taking well over 150GB on the same systems people run this emulator, I don't think 2GB is asking a lot. There is already an option to dump the RAM in the debug menu, laying some of the ground work for this.

I'm willing to work on this personally but had a question about the implementation. 

Is it as easy as saving the memory contents, register values, and next instruction address to be loaded back in or is there more to the machine state that is needed?

cabrerahector:
> I don't think 2GB is asking a lot

That's if it's only 2GB though. Think having several save states for the same game and/or multiple games (having just one would be pretty limiting after all). In such cases you would run out of storage pretty quickly.

Spegs21:
Sure, multiple 2GB save states will add-up but the burden of managing storage is on the user.

My use case of save states is to work around poor save system implementations so I'd only need one per game, if it was problematic. I wouldn't be saving an additional state every few minutes, I'd use the same one occasionally. I'm sure other people have other uses cases. I just don't think this should be precluded because we don't trust the user to manage storage. Is that really our concern?

Exzap:
The reason we don't have save states yet isn't so much the size and loading times but just the sheer quantity of busy work it is to implement them. Also, any change to any system in Cemu that holds some kind of dynamic runtime information (and there are a lot, it's an HLE emulator after all) would invalidate previous save states if no backwards compatibility code is added. So save states would put a substantial maintainability burden on the whole codebase. That said, the code is pretty stable and doesn't change often anymore.

It's one of the most common feature requests after all. If you decided to spearhead it, it would be appreciated by many :)

Last year someone already started work on it but dropped it pretty quickly. Their repo is [here](https://github.com/ChrisNonyminus/Cemu/tree/savestates) if you want to check it out as a reference.

Off the top of my head what needs to be stored:
- Every PPC thread and it's current thread context (`OSThread_t` struct)
- Full RAM dump
- All of the [CafeOS modules and drivers](https://github.com/Exzap/Cemu/tree/main/src/Cafe/OS/libs) need their own load/store. Not very complicated per module, there are just a lot of them.
- Similarly, all of the [IOSU modules](https://github.com/Exzap/Cemu/tree/main/src/Cafe/IOSU)
- Dynamic textures which only resides in VRAM and cant be restored from a RAM dump alone
[...]

PR #953, opened 2023-08-31
Add save states

Spegs21:
On going save state implementation per @Exzap comments in #898 

Initial work based on [this fork](https://github.com/ChrisNonyminus/Cemu/tree/savestates).

- [X] Every PPC thread and it's current thread context (`OSThread_t` struct)
- [X] Full RAM dump
- All of the [CafeOS modules and drivers](https://github.com/Exzap/Cemu/tree/main/src/Cafe/OS/libs) need their own load/store
   - [X] coreinit 
- All of the [IOSU modules](https://github.com/Exzap/Cemu/tree/main/src/Cafe/IOSU)
   - [X] fsa
- [ ] Dynamic textures which only resides in VRAM and cant be restored from a RAM dump alone
- [ ] The vertex and uniform cache because it can hold information that isn't synced back to RAM
- Various settings:
   - [ ] current graphic packs configuration 
   - [ ] controller config
   - [ ] renderer settings
- [X] Pause & resume emulation feature (not sure about the implementation of this one)


amayra:
did this RP get dropped ?

PR #956, opened 2023-09-02
Localization improvements and fixes

Fs00:
This PR aims to fix several issues in UI localization that I've found while working to update the Italian translation to the latest version.
The issues I've fixed can be roughly divided in the following categories:
- the UI displayed some strings that couldn't be translated
- the UI displayed some strings in their untranslated form even if the translation was actually present, due to missing/wrong gettext lookups (eaf29a1, 63daf47, 63ca572, 2399808)
- the source code included some strings that could have been translated, but those strings belonged to unused functions or were never displayed to the user (fe1297f, af1881f)

Overall, there is a fair amount of different changes in this PR, but they should be fairly quick and easy to review commit by commit.

### DIFF
--- a/src/Cafe/Account/Account.h
+++ b/src/Cafe/Account/Account.h
@@ -16,22 +16,6 @@ enum class OnlineAccountError
 	kPasswordCacheEmpty,
 	kNoPrincipalId,
 };
-template <>
-struct fmt::formatter<OnlineAccountError> : formatter<string_view> {
-	template <typename FormatContext>
-	auto format(const OnlineAccountError v, FormatContext& ctx) {
-		switch (v)
-		{
-		case OnlineAccountError::kNoAccountId: return formatter<string_view>::format("AccountId missing (The account is not connected to a NNID)", ctx);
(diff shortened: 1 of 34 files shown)


Exzap:
Looks good!

Just one nitpick:
We have specific wrappers for string conversation to guaranteed that all `std::string` are utf8. For wxString <-> std::string there is [to_wxstring/from_wxstring](https://github.com/cemu-project/Cemu/blob/d7f0d679047e9fe87de6b36cdd7f2be8b9800477/src/gui/helpers/wxHelpers.h#L90). Always use this over e.g. `.ToStdString()`

Exzap:
FYI Using wxWidgets functions in the emulation core (or really anything outside of `src/gui` subdirectory) is something we ideally want to avoid. The goal is to fully decouple UI code from the core eventually to make it possible to introduce other UI libraries. Hence it is preferable to use `std::format` over `formatWxString` in the core. Of course there are other existing issues too like direct calls to `wxMessageBox` and various wx dialogs but let that be a problem for another day.

At any rate this PR looks good as-is. Thanks!

Fs00:
> Hence it is preferable to use std::format over formatWxString in the core

Yeah makes a lot of sense. But as you've already pointed out, the real problem here is calling `wxMessageBox` outside the GUI code... unless we move those calls outside the core, we have no choice but to deal with `wxString`s in one way or another :man_shrugging:

PR #963, opened 2023-09-06
Korean translation

Hackjjang:
Binary file not uploading


Exzap:
I merged your PR over at Cemu-Language and it will be part of the next experimental release. You don't need to submit it here too.

Issue #588, opened 2022-12-22
Disney Infinity/LEGO Dimensions built-In portal emulation


Fancy2209:
As the title says, it would be interesting if CEMU could have a way to emulate those toys to life Portals like how RPCS3 (and soon dolphin from what I have seen) can for the Skylanders Portal at least

Squall-Leonhart:
It already does.

Fancy2209:
> It already does.
> 
> https://wiki.cemu.info/wiki/Skylanders:_Swap_Force

I do not mean native portals 
I mean a Emulated Portal

Fancy2209:
E.g on RPCS3 Manage -> Skylanders Portal

spudpiggy:
> As the title says, it would be interesting if CEMU could have a way to emulate those toys to life Portals like how RPCS3 (and soon dolphin from what I have seen) can for the Skylanders Portal at least

They don't really need to be built in imo. Lego Dimensions' Wii U version is often considered its worst. I personally emulate the PS3 version [using this toypad emulator. inb4 yes it works on cemu](https://github.com/Berny23/LD-ToyPad-Emulator)

Fancy2209:
> 

I've used that before. To be honest I made this issue more for Skylanders/Disney Infinity than for LEGO Dimensions

minilandl:
I am trying to get the original toypad working in cemu on Linux 

Tried using the arch package.
The toypad showed up fine in lsusb 
When launching Lego dimensions the toypad was still not detected.

Could be windows specific as previously I couldn't get the toypad working under Linux but it worked in windows and was detected by cemu

Vicourld51:
I saw that the RPCS3 code for portal of power had been used on dolphin, is it possible to adapt it to cemu?

Squall-Leonhart:
No, as it would require the original contributor to dual license it.

Vicourld51:
So maybe adapt only the RPCS3 code, and name the license like dolphin did. I wasn't talking about taking dolphin's code directly!

XargonWan:
I'm looking for this feature as well, I have the physical one, but I would like an emulated one for when I'm not docked with my Steam Deck.

Fancy2209:
There's #971

Exzap:
Reopening this because only Skylander portal is emulated for now.

Fancy2209:
[...]

PR #971, opened 2023-09-20
nsyshid: Emulate Skylander Portal

deReeperJosh:
Carrying on from some of the work I have done on Dolphin and RPCS3, and thanks to https://github.com/cemu-project/Cemu/pull/950 being merged, I am opening this PR up with the ability to emulate the Skylander Portal on Cemu.

Currently have tested Trap Team and Swap Force as working on my Mac, but hoping to get feedback and some testers on this PR too. Hoping to add Disney Infinity in the near future, but I don't currently have access to my Wii U so can't dump the necessary erreula to test my changes have actually worked there.

Would partly close https://github.com/cemu-project/Cemu/issues/588

### DIFF
--- a/src/Cafe/OS/libs/nsyshid/Backend.h
+++ b/src/Cafe/OS/libs/nsyshid/Backend.h
@@ -23,6 +23,55 @@ namespace nsyshid
 		/* +0x12 */ uint16be maxPacketSizeTX;
 	} HID_t;
 
+	struct TransferCommand
+	{
+		uint8* data;
+		sint32 length;
+
+		TransferCommand(uint8* data, sint32 length)
+			: data(data), length(length)
+		{
+		}
+		virtual ~TransferCommand() = default;
+	};
+
+	struct ReadMessage final : TransferCommand
+	{
+		sint32 bytesRead;
+
+		ReadMessage(uint8* data, sint32 length, sint32 bytesRead)
+			: bytesRead(bytesRead), TransferCommand(data, length)
+		{
+		}
+		using TransferCommand::TransferCommand;
+	};
+
+	struct WriteMessage final : TransferCommand
+	{
+		sint32 bytesWritten;
+
+		WriteMessage(uint8* data, sint32 length, sint32 bytesWritten)
+			: bytesWritten(bytesWritten), TransferCommand(data, length)
+		{
+		}
+		using TransferCommand::TransferCommand;
+	};
+
+	struct ReportMessage final : TransferCommand
+	{
+		uint8* reportData;
+		sint32 length;
+		uint8* originalData;
+		sint32 originalLength;
+
+		ReportMessage(uint8* reportData, sint32 length, uint8* originalData, sint32 originalLength)
+			: reportData(reportData), length(length), originalData(originalData),
+			  originalLength(originalLength), TransferCommand(reportData, length)
+		{
+		}
+		using TransferCommand::TransferCommand;
+	};
+
 	static_assert(offsetof(HID_t, vendorId) == 0x8, "");
 	static_assert(offsetof(HID_t, productId) == 0xA, "");
 	static_assert(offsetof(HID_t, ifIndex) == 0xC, "");
@@ -69,7 +118,7 @@ namespace nsyshid
 			ErrorTimeout,
 		};
 
-		virtual ReadResult Read(uint8* data, sint32 length, sint32& bytesRead) = 0;
+		virtual ReadResult Read(ReadMessage* message) = 0;
 
 		enum class WriteResult
 		{
@@ -78,7 +127,7 @@ namespace nsyshid
 			ErrorTimeout,
 		};
 
-		virtual WriteResult Write(uint8* data, sint32 length, sint32& bytesWritten) = 0;
+		virtual WriteResult Write(WriteMessage* message) = 0;
 
 		virtual bool GetDescriptor(uint8 descType,
 								   uint8 descIndex,
@@ -88,7 +137,7 @@ namespace nsyshid
 
 		virtual bool SetProtocol(uint32 ifIndef, uint32 protocol) = 0;
 
-		virtual bool SetReport(uint8* reportData, sint32 length, uint8* originalData, sint32 originalLength) = 0;
+		virtual bool SetReport(ReportMessage* message) = 0;
 	};
 
 	class Backend {
@@ -120,6 +169,8 @@ namespace nsyshid
 		void DetachDevice(const std::shared_ptr<Device>& device);
 
 		std::shared_ptr<Device> FindDevice(std::function<bool(const std::shared_ptr<Device>&)> isWantedDevice);
+
+		bool FindDeviceById(uint16 vendorId, uint16 productId);
 
 		bool IsDeviceWhitelisted(uint16 vendorId, uint16 productId);
 
--- a/src/Cafe/OS/libs/nsyshid/Skylander.cpp
+++ b/src/Cafe/OS/libs/nsyshid/Skylander.cpp
@@ -1,0 +1,939 @@
+#include "Skylander.h"
+
+#include "nsyshid.h"
+#include "Backend.h"
+
+#include "Common/FileStream.h"
+
+namespace nsyshid
+{
+	SkylanderUSB g_skyportal;
+
+	const std::map<const std::pair<const uint16, const uint16>, const std::string>
+		listSkylanders = {
+			{{0, 0x0000}, "Whirlwind"},
+			{{0, 0x1801}, "Series 2 Whirlwind"},
+			{{0, 0x1C02}, "Polar Whirlwind"},
+			{{0, 0x2805}, "Horn Blast Whirlwind"},
+			{{0, 0x3810}, "Eon's Elite Whirlwind"},
+			{{1, 0x0000}, "Sonic Boom"},
+			{{1, 0x1801}, "Series 2 Sonic Boom"},
+			{{2, 0x0000}, "Warnado"},
+			{{2, 0x2206}, "LightCore Warnado"},
+			{{3, 0x0000}, "Lightning Rod"},
+			{{3, 0x1801}, "Series 2 Lightning Rod"},
+			{{4, 0x0000}, "Bash"},
+			{{4, 0x1801}, "Series 2 Bash"},
+			{{5, 0x0000}, "Terrafin"},
+			{{5, 0x1801}, "Series 2 Terrafin"},
+			{{5, 0x2805}, "Knockout Terrafin"},
+			{{5, 0x3810}, "Eon's Elite Terrafin"},
+			{{6, 0x0000}, "Dino Rang"},
+			{{6, 0x4810}, "Eon's Elite Dino Rang"},
+			{{7, 0x0000}, "Prism Break"},
+			{{7, 0x1801}, "Series 2 Prism Break"},
+			{{7, 0x2805}, "Hyper Beam Prism Break"},
+			{{7, 0x1206}, "LightCore Prism Break"},
+			{{8, 0x0000}, "Sunburn"},
+			{{9, 0x0000}, "Eruptor"},
+			{{9, 0x1801}, "Series 2 Eruptor"},
+			{{9, 0x2C02}, "Volcanic Eruptor"},
+			{{9, 0x2805}, "Lava Barf Eruptor"},
+			{{9, 0x1206}, "LightCore Eruptor"},
+			{{9, 0x3810}, "Eon's Elite Eruptor"},
+			{{10, 0x0000}, "Ignitor"},
+			{{10, 0x1801}, "Series 2 Ignitor"},
+			{{10, 0x1C03}, "Legendary Ignitor"},
+			{{11, 0x0000}, "Flameslinger"},
+			{{11, 0x1801}, "Series 2 Flameslinger"},
+			{{12, 0x0000}, "Zap"},
+			{{12, 0x1801}, "Series 2 Zap"},
+			{{13, 0x0000}, "Wham Shell"},
+			{{13, 0x2206}, "LightCore Wham Shell"},
+			{{14, 0x0000}, "Gill Grunt"},
+			{{14, 0x1801}, "Series 2 Gill Grunt"},
+			{{14, 0x2805}, "Anchors Away Gill Grunt"},
+			{{14, 0x3805}, "Tidal Wave Gill Grunt"},
+			{{14, 0x3810}, "Eon's Elite Gill Grunt"},
+			{{15, 0x0000}, "Slam Bam"},
+			{{15, 0x1801}, "Series 2 Slam Bam"},
+			{{15, 0x1C03}, "Legendary Slam Bam"},
+			{{15, 0x4810}, "Eon's Elite Slam Bam"},
+			{{16, 0x0000}, "Spyro"},
+			{{16, 0x1801}, "Series 2 Spyro"},
+			{{16, 0x2C02}, "Dark Mega Ram Spyro"},
+			{{16, 0x2805}, "Mega Ram Spyro"},
+			{{16, 0x3810}, "Eon's Elite Spyro"},
+			{{17, 0x0000}, "Voodood"},
+			{{17, 0x4810}, "Eon's Elite Voodood"},
+			{{18, 0x0000}, "Double Trouble"},
+			{{18, 0x1801}, "Series 2 Double Trouble"},
+			{{18, 0x1C02}, "Royal Double Trouble"},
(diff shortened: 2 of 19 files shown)


src/Cafe/OS/libs/nsyshid/Skylander.cpp
```diff
@@ -0,0 +1,400 @@
...
+	}
+
+	std::array<uint8, 64> SkylanderUSB::get_status()
+	{
+		std::array<uint8, 64> interrupt_response = {};
+
+		if (!m_queries.empty())
+		{
+			interrupt_response = m_queries.front();
+			m_queries.pop();
```
ssievert42, line 351: Excuse my paranoia, but I think interacting with `m_queries` like this should be protected by a mutex, and the `m_queries.push` in `SkylanderUSB::control_transfer` on line 197 as well.

ssievert42 (changes requested):

src/Cafe/OS/libs/nsyshid/nsyshid.cpp
```diff
@@ -256,6 +258,17 @@ namespace nsyshid
...
 
+	void AttachEmulatedDevices()
+	{
+		if (GetConfig().emulated_usb_devices.emulate_skylander_portal && !backendList.front()->m_foundSkylander)
+		{
+			cemuLog_logDebug(LogType::Force, "Attaching Emulated Portal");
+			// Add Skylander Portal
+			auto device = std::make_shared<SkylanderPortalDevice>();
+			AttachDevice(device);
+		}
```
ssievert42, line 269: How about moving this code to a new backend, attaching that backend in `AttachDefaultBackends`, and running the code in `AttachVisibleDevices`?
This would require a method in the `Backend` class to check for devices that were attached by other backends though; see my other comment [here](https://github.com/cemu-project/Cemu/pull/971#discussion_r1331513601).
That way, there wouldn't be any device specific code in nsyshid.cpp, like what is currently the case for libusb and Windows HID stuff.

ssievert42 (changes requested):

src/Cafe/OS/libs/nsyshid/Skylander.cpp
```diff
@@ -0,0 +1,400 @@
...
+		*(uint16be*)(currentWritePtr + 7) = 0x001D; // wDescriptorLength
+		currentWritePtr = currentWritePtr + 9;
+		// endpoint descriptor 1
+		*(uint8*)(currentWritePtr + 0) = 7;		  // bLength
+		*(uint8*)(currentWritePtr + 1) = 0x05;	  // bDescriptorType
+		*(uint8*)(currentWritePtr + 1) = 0x81;	  // bEndpointAddress
+		*(uint8*)(currentWritePtr + 2) = 0x03;	  // bmAttributes
+		*(uint16be*)(currentWritePtr + 3) = 0x40; // wMaxPacketSize
+		*(uint8*)(currentWritePtr + 5) = 0x01;	  // bInterval
+		currentWritePtr = currentWritePtr + 7;
```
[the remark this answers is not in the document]
deReeperJosh: I think I tested this with my portal and disney base previously and they both had the same or similar, besides max packet size on the disney base was 0x20

src/Cafe/OS/libs/nsyshid/Skylander.cpp
```diff
@@ -0,0 +1,400 @@
...
+		message->isFake = true;
+		message->expectedTimeMillis = 10;
+		message->bytesRead = message->length;
+		return Device::ReadResult::Success;
+	}
+
+	Device::WriteResult SkylanderPortalDevice::Write(WriteMessage* message)
+	{
+		return Device::WriteResult::Success;
+	}
```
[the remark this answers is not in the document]
deReeperJosh: Good point

src/Cafe/OS/libs/nsyshid/Skylander.cpp
deReeperJosh: Yeah a mutex makes sense here

Exzap:
Is this still being worked on or is it ready for review? Asking because some suggestions by @ssievert42 are still open or without comment.

deReeperJosh:
Yes sorry, just been busy with work, will fix those comments this week hopefully

deReeperJosh:
Just pushed up some changes, but still haven't implemented the find device across backends method, so will aim to do that before the end of the week.

deReeperJosh:
All those requested changes have now been made, let me know if that's what you had anticipated it looking like @ssievert42 

Otherwise, I am not convinced with the performance on the Vicarious Visions games (Swap Force and Superchargers) will work yet, I've had varying degrees of success because at some point the game decides to stop sending commands to the portal, presumably because the emulated portal has replied too fast or too slow. Will keep investigating the perfect timing.

deReeperJosh:
Had a playthrough of Swap Force and 10s might be the sweet spot for now, could be something that I can add as a custom option so that users can increase/decrease themselves to try to avoid the same issues that occur with a real portal, but for now the PR is good to go

src/Cafe/OS/libs/nsyshid/nsyshid.cpp
```diff
@@ -759,6 +762,19 @@ namespace nsyshid
...
+		std::lock_guard<std::recursive_mutex> lock(this->m_devicesMutex);
+		for (const auto& device : m_devices)
+		{
+			if (device->m_vendorId == vendorId && device->m_productId == productId)
+			{
+				return true;
+			}
+		}
+		return false;
+	}
```
ssievert42, line 776: This won't work to check for globally connected devices, as we just walk through the backend's own list of devices that should be detached when the backend is detached.
Although I generally don't like the word "global", maybe we could rename this method to something like `FindGlobalDeviceById` and make it iterate over the `deviceList` and lock the `hidMutex` that are part of `nsyshid`?
Suggestions welcome :sweat_smile:

src/Cafe/OS/libs/nsyshid/Skylander.cpp
```diff
@@ -0,0 +1,401 @@
...
+		*(uint8*)(currentWritePtr + 6) = 0x01;	  // bInterval
+		currentWritePtr = currentWritePtr + 7;
+		// endpoint descriptor 2
+		*(uint8*)(currentWritePtr + 0) = 7;		  // bLength
+		*(uint8*)(currentWritePtr + 1) = 0x05;	  // bDescriptorType
+		*(uint8*)(currentWritePtr + 1) = 0x02;	  // bEndpointAddress
+		*(uint8*)(currentWritePtr + 2) = 0x03;	  // bmAttributes
+		*(uint16be*)(currentWritePtr + 3) = 0x40; // wMaxPacketSize
+		*(uint8*)(currentWritePtr + 5) = 0x01;	  // bInterval
+		currentWritePtr = currentWritePtr + 7;
```
ssievert42, line 108: This still writes twice to `currentWritePtr + 1` and leaves `currentWritePtr + 6` uninitialised.
I guess I should have initially commented that on both of the endpoint descriptors.

ssievert42 (commented):
I like it!
Just some minor nitpicks :)
About the timing issues: These kinds of problems tend to make me really nervous :sweat_smile:. Do you think subtracting the time it takes to run the other stuff in `Read` and `SetReport` from the time that should be slept before we can return might help?

src/Cafe/OS/libs/nsyshid/Backend.h
```diff
@@ -23,6 +23,34 @@ namespace nsyshid
...
+		uint8* data;
+		sint32 length;
+	};
+
+	struct ReadMessage final : TransferCommand
+	{
+		uint8* data;
+		sint32 length;
+		sint32 bytesRead;
+	};
```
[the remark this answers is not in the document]
deReeperJosh: Had to be a bit creative with this, but think that this should be resolved now

src/Cafe/OS/libs/nsyshid/nsyshid.cpp
deReeperJosh: Have kept the same name, but just have a backend version and an nsyshid version, let me know if that looks right!

deReeperJosh:
@ssievert42 more review work added - unfortunately without a proper core timing module there isn't much we can do besides sleep the thread for a bit, I have the same work on Dolphin and that would be ideal (schedule the transfer to be completed at x amount of thread ticks rather than time based), but would need a major rework to Cemu's core. Even detaching/reattaching doesn't fix the issue, the game just decides to stop sending commands to the portal, go figure

ssievert42:
Welp, that's certainly a nice thing to do of the game.
Putting aside the fact that I can't actually test this, because I don't own any Skylanders games, I've run out of things to complain about :)

LGTM :+1:

ElBread3:
Hello, I just did some testing with Skylanders Giants. Cemu consistently crashes, either when the Activision logo is about to stop spinning, at various random points at the middle and end of the Vicarious Visions logo, or (my most successful attempt) a crash on "Press A". Just for kicks I tested with vanilla Cemu and it booted all the way through.
I wonder if that's the timing issue with the emulated portal, but from previous talk it sounded like it didn't crash Cemu, as by then you wouldn't be able to "detach and reattach the emulated portal". Hope my testing helps, looking forward to seeing this feature in Cemu

deReeperJosh:
@ElBread3 thanks for helping test, I don't have giants for the Wii U so can't confirm that but will see if I can get someone else to recreate the bug so I can try to fix it locally. I would have thought trap team would have the same issue but I guess not!

ElBread3:
Tested with a debug executable today and did not crash, though it takes so long to load. Planning on testing with Disney Infinity later once I buy it

deReeperJosh:
I ended up grabbing giants and I had no crashes - might be Windows only so I will try to build on a Windows machine to see if I can replicate. Have you tried any other Skylander games @ElBread3?

deReeperJosh:
Just made the layout of the window look a bit nicer and even

ElBread3:
OK I tested the print fix and Giants booted and is playable. To answer if I have other Skylanders games I also own Imaginators and Superchargers; The only other game that crashed when I did my first tests was Superchargers, crashing on the logo appearing, but it booted after a handful of attempts and still boots consistently. And Imaginators works fine.

ssievert42:
@deReeperJosh That call to `fmt::print` can go :+1:
(And shouldn't be there in the first place; If you look at the history of that `_debugPrintHex` function you'll notice that I managed to forget that call in there.)
But I don't understand what could cause that error message either. At least in the context of `_debugPrintHex` we shouldn't be overwriting parts of the stack ¯\\\_(ツ)_/¯
Looking at recent PRs, #977 fixes a crash on Windows caused by the same `fmt::print` call, so I guess it's just not a good idea to use `fmt` to print stuff in Cemu, when there is a perfectly good function called `cemuLog_logDebug` to log something instead.

Squall-Leonhart:
>  But I don't understand what could cause that error message either. At least in the context of _debugPrintHex we shouldn't be overwriting parts of the stack ¯\_(ツ)_/¯

We had some chat about it in https://discord.com/channels/286429969104764928/1017917335874572308/1155847973863964702 if removing it is ideal, then go ahead and i'll close the other pr, the logged error was basically "Bad file descriptor" while writing to fmt:: though.

deReeperJosh:
@ElBread3 haven't ever actually played lego dimensions, but if someone has done some documentation on the USB protocols for it I can try to pick it up :)

ETE-Design:
@deReeperJosh Have you looked at the Disney Infinity Emulation yet?

deReeperJosh:
@ETE-Design yes, I have a branch on my repo where the portal is recognised (no character loading yet) but I don't have a Wii U and games to test with any more, so I can't get past the EULA Error screen

ElBread3:
@deReeperJosh I've had to modify the code to load the infinity base and dump sys, but I've gotten past the error. However I can't test the characters cause the game doesn't render. Do you guys want me to record the boot up with debug overlay and put it here?

Squall-Leonhart:
skylanders renders just fine on current code though.....

ElBread3:
@Squall-Leonhart I was referring to Disney Infinity. An interesting note is that Disney Infinity seems to have nearly the same issue with rendering as Cars 3: Driven To Win, also developed by the same studio and therefore probably runs on the same engine. But yes, the Skylanders games work great

deReeperJosh:
Not sure if there is much more left on this PR but I have become largely too busy to be able to continue working on fun side projects such as this - I will keep this PR open and continue to update with commits to main when I get the chance but won't be able to pay as much attention to it as previously

iamc24:
@deReeperJosh Thanks for your work on this, I'm almost done playing Skylanders giants on your fork, and it has worked very well so far. I plan on playing the following 4 entries as well, so I'll let you know about any instability or other issues. Don't worry about fixing anything until/unless you have time for it; I totally understand life getting in the way of hobbies and side projects.

deReeperJosh:
@ETE-Design I have a branch on my repo where I have my infinity base progress, there isn't much there yet (no loading/removing figures or anything) but it would be good if you could test getting in game, you need to dump the erreula from your console iirc to get past the menu

Squall-Leonhart:
The skylander portal speaker is underflowing when playing audio, leading to the issue reported in #1146

deReeperJosh:
@Squall-Leonhart that is mostly an issue with the Windows HID transfers, afaik libusb doesn't have the same problem, would be ideal if windows users also had the option to use libusb in the future

Squall-Leonhart:
Quite possibly, i can't recall what the barrier for using it on windows is, but building with -DENABLE_NSYSHID_WINDOWS_HID=OFF hasn't demonstrated any odd behaviors in lego dimensions to date, alas i don't have skylanders to verify if the portal sounds work better with it.

Exzap:
Changing the flag is not enough. Last time I tried, libusb via vcpkg had issues linking on Windows. It's probably solvable but then more work needs to be done so we can support windows HID API + libusb devices in parallel. At any rate this topic is outside the scope of this PR.

deReeperJosh:
@Exzap can't seem to directly request a review from you without being a member of the Cemu team, but just tagging you here to add to your backlog

deReeperJosh:
Latest push adds in a Create Skylander Dialog to the window, which was missing previously while I was still figuring wxWidgets out.

src/Cafe/OS/libs/nsyshid/Skylander.h
```diff
@@ -0,0 +1,98 @@
...
+
+		bool SetProtocol(uint32 ifIndex, uint32 protocol) override;
+
+		bool SetReport(ReportMessage* message) override;
+
+	  private:
+		bool m_IsOpened;
+	};
+
+	extern const std::map<const std::pair<const uint16, const uint16>, const std::string> listSkylanders;
```
Exzap, line 39: `listSkylanders` should be renamed to `s_listSkylanders` to match our naming convention. Even better would be to avoid using a loose global variable altogether. Instead, accessing it via a static getter method in SkylanderUSB would be preferable. What I mean is accessing it like this:
```
nsyshid::SkylanderUSB::GetSkylanders();
```
Which imho looks better than `nsyshid::skylanders`

src/Cafe/OS/libs/nsyshid/Skylander.cpp
```diff
@@ -0,0 +1,939 @@
...
+		uint32 skySerial = 0;
+		for (int i = 3; i > -1; i--)
+		{
+			skySerial <<= 8;
+			skySerial |= buf[i];
+		}
+		uint8 foundSlot = 0xFF;
+
+		// mimics spot retaining on the portal
+		for (auto i = 0; i < 16; i++)
```
Exzap, line 800:
```suggestion
		for (auto i = 0; i < m_skylanders.size(); i++)
```

Exzap (commented):
This looks mostly good to me. My only complaints are of cosmetic nature in regards to variable naming and code structure. See my other annotations below.

Additionally, I think extra validation in many of the functions that work with raw data wouldn't hurt, especially in the `SkylanderUSB` class, but I couldn't spot any potential out-of-bounds access so it's fine.

A lot of the manual data juggling and hardcoded array indices could be avoided by defining the Skylanders data as a struct instead of `std::array<uint8, 0x40 * 0x10>` or raw C uint8* pointers. Similar to how we are already doing it [for Amiibos](https://github.com/cemu-project/Cemu/blob/f3d20832c191f90fe4ad6f9b64a3ff21eb477b02/src/Cafe/OS/libs/nn_nfp/nn_nfp.cpp#L99).

At any rate, I think this PR has been open long enough and we don't need to delay it any further for purely surface level code changes. I will merge it as-is. If you want you can address my feedback in a future PR.

As mentioned already, portal emulation will make a lot of people happy :smile: Thanks for your contribution!

PR #972, opened 2023-09-20
Vulkan: only create imgui render pass once

goeiecool9999:
The render pass got created every time the swapchain is recreated (including resize). I couldn't find anywhere that actually destroys the old objects (including imgui code) so I think it just leaks. Imgui example code also seems to manually manage the render pass lifetime.

### DIFF
--- a/src/Cafe/HW/Latte/Renderer/Vulkan/VulkanRenderer.cpp
+++ b/src/Cafe/HW/Latte/Renderer/Vulkan/VulkanRenderer.cpp
@@ -1508,34 +1508,37 @@ void VulkanRenderer::DeleteNullObjects()
 
 void VulkanRenderer::ImguiInit()
 {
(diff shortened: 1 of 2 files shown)


goeiecool9999:
Maybe I should change it so it does get recreated every time? When I read the TODO comment I figured that format changing wasn't working for render passes but on second thought it might actually work currently since a format change in the swapchain would also trigger a new render pass to be created. I could change it to a check that destroys the current imgui render pass if it already exists.

goeiecool9999:
Can the gamepad and the TV use different formats? In that case we might have an issue because it doesn't look like imgui supports multiple render pass objects. Changing to code so it keep recreating the render pass every swapchain update would not be a proper implementation of format switching in that case so it wouldn't be that much better than leaving the TODO and always reusing the same render pass.

Exzap:
It might be easier to solve the format problem with dynamic rendering (VK_KHR_dynamic_rendering)?

PR #977, opened 2023-09-25
Catch exception in Nsyshid on Windows and log it in logDebug

Squall-Leonhart:
This keeps Cemu from crashing when nsyshid's logging encounters an error on Windows and prints the type of exception encountered to the log.

Currently a bad descriptor error is occurring on some toypad accesses, I have not addressed this (Don't know how), but this makes cemu no longer crash when they happen.


ssievert42:
Hey, it looks like I managed to forget that call to `fmt::print` in there when hacking on nsyshid.
Since that call shouldn't be there in the first place, does removing it fix the crash as well?

Squall-Leonhart:
Superseded by #971

PR #994, opened 2023-10-14
Fix several language selection issues

### LINKED ISSUE #580: Right-to-left UI interface
I set system locale in windows in Arabic (Egypt) and the interface is shown right-to-left.
Even if set the language to English, it still shows right to left interface.
When i change the language other than English (French for example), it shows left-to-right inteface normally.
Please fix this locale bug.

Screenshot 2022-12-16 134303
Screenshot 2022-12-16 134347
Screenshot 2022-12-16 134906

### LINKED ISSUE #616: [Linux] Cannot set locale to language "English" warning (version 2.0-25)
I am using Cemu experimental version [2.0-25](https://github.com/cemu-project/Cemu/commit/33bd10b4e0b9c27881fa7d6bf38908791f84d3b2). Whenever I open the appimage build of Cemu, a warning box pops up saying "Cannot set locale to language 'English'".

 This error happens on the appimage build but does not happen on the unpackaged binary ("Ubuntu" archive). In neither builds there is an actual problem with the UI language.

There are no specific steps to reproduce this bug other than to launch the appimage build of version 2.0-25

I am using Debian (testing/bookworm) with kernel version 6.0.12 and KDE Plasma under x11, with the european portuguese and american english locales installed on my system. I can provide more system information if needed but for this kind of issue it doesn't seem necessary.

### LINKED ISSUE #717: [Cemu-Language] : "English" resource file missing in Cemu releases causes the UI language setting to reset back to "Default" and show the "emulator should be restarted" prompt twice
When I choose "English" as my UI language and save, I get the prompt that I need to restart Cemu, as expected.
Then I restart Cemu.
Then I go back to the UI language setting, it's now been changed to "Default".
I don't touch it. I close the setting. But I get the prompt again that I need to restart Cemu.
I restart Cemu again. Now I won't get the prompt again if I go to settings again. But now the setting is stuck at "Default" when it should probably stay as "English" as I had chosen.

It shouldn't be going back to "Default" after having chosen "English".

My system language is Turkish. I don't know if that has anything to do with the default detection.

I checked and confirm this is still happening on the latest version.

Fs00:
This PR simplifies the logic for loading UI translations, which should fix several issues people have encountered.
First, we now translate UI strings using the [wxTranslations](https://docs.wxwidgets.org/3.2/classwx_translations.html) class directly: this makes language selection work with languages that aren't installed/supported on the operating system (this mostly applies to Linux: before this PR, choosing a language not installed on the machine didn't do anything).
Furthermore, the logic used to apply the UI localization has been fixed so that when the "Default" language is selected, the UI uses the system default language (and not English as it did before). The "English" language option now also works as intended: it stays selected upon closing and reopening the settings window and when it's selected, no localization is applied.

While I was at it, I also took the opportunity to format the "last played" date in the game list using the currently selected locale.

Some testing on macOS would be appreciated.

Fixes #717 
Fixes #616
Should fix #580

### DIFF
--- a/src/gui/CemuApp.cpp
+++ b/src/gui/CemuApp.cpp
@@ -238,7 +238,7 @@ void CemuApp::LocalizeUI()
 
 std::vector<const wxLanguageInfo*> CemuApp::GetAvailableTranslationLanguages(wxTranslations* translationsMgr)
 {
-	wxFileTranslationsLoader::AddCatalogLookupPathPrefix(ActiveSettings::GetDataPath("resources").generic_string());
+	wxFileTranslationsLoader::AddCatalogLookupPathPrefix(wxHelper::FromPath(ActiveSettings::GetDataPath("resources")));
 	std::vector<const wxLanguageInfo*> languages;
 	for (const auto& langName : translationsMgr->GetAvailableTranslations("cemu"))
 	{
--- a/src/gui/CemuApp.cpp
+++ b/src/gui/CemuApp.cpp
@@ -228,7 +228,7 @@ void CemuApp::LocalizeUI()
 		translationsMgr->SetLanguage(static_cast<wxLanguage>(configuredLanguage));
 		translationsMgr->AddCatalog("cemu");
 
-		if (wxLocale::IsAvailable(configuredLanguage))
+		if (translationsMgr->IsLoaded("cemu") && wxLocale::IsAvailable(configuredLanguage))
 			m_locale.Init(configuredLanguage);
 
 		// This must be run after wxLocale::Init, as the latter sets up its own wxTranslations instance which we want to override
--- a/src/gui/CemuApp.cpp
+++ b/src/gui/CemuApp.cpp
@@ -99,29 +99,7 @@ bool CemuApp::OnInit()
 
 	wxInitAllImageHandlers();
 
-
-	m_languages = GetAvailableLanguages();
-
-	const sint32 language = GetConfig().language;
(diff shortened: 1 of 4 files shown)


Fs00:
I have added a commit to prevent initializing the locale if Cemu has no UI translation for the default language, which should fix #580.

src/gui/CemuApp.cpp
```diff
@@ -231,33 +209,44 @@ int CemuApp::FilterEvent(wxEvent& event)
...
-		result.emplace_back(lang_info);
+		// This must be run after wxLocale::Init, as the latter sets up its own wxTranslations instance which we want to override
+		wxTranslations::Set(translationsMgr.release());
 	}
+}
 
-	return result;
+std::vector<const wxLanguageInfo*> CemuApp::GetAvailableTranslationLanguages(wxTranslations* translationsMgr)
+{
+	wxFileTranslationsLoader::AddCatalogLookupPathPrefix(ActiveSettings::GetDataPath("resources").generic_string());
```
Exzap, line 241: I don't trust `.generic_string()` to do the right conversion here in all situations. We have a helper which avoids implementation defined behavior: `wxString wxHelper::FromPath(const fs::path& path)`.

Fs00: Ok! Now it should be fine

hauntek:
image

Thank you very much. The actual testing did indeed fix the issue of using the default local language for macOS, and both the switching were effective

PR #1010, opened 2023-10-29
Fix  graphicPacks load failed if use utf8 exe path

lijunyu-cn:
Test passed when exe path use chinese or japanese.


Squall-Leonhart:
Also solves the issue reported in https://github.com/cemu-project/Cemu/issues/1018

Exzap:
Sorry for the long silence. I have addressed the issue in a broader rewrite (See c380975f308d4d4b070cf0e1cf91f2bab6edddd3) which supersedes this PR. Appreciate your effort nevertheless.

PR #1012, opened 2023-10-29
Vulkan: Change double buffered VSync to match display without blocking

goeiecool9999:
This PR attempts to eliminate unnecessary blocking on the GPU thread while using the FIFO present mode.
In order to still match the frame rate of the display like blocking FIFO does we do the following:
Tracks how fast frames are being presented and take over signalling vsync if the presentation interval is within 1% of the emulated vsync interval for more than 60 frames. Returns to the fixed emulated timer if it is more than 1% from the desired interval for 60 frames.
Tripple buffered VSync and VSync off behaviour remains unchanged and use the old method of driving present with SwapBuffers, which allows VRR to reduce stuttering and external tools to see exact frametimes

Implementation:
- Use two vulkan textures as a representation of a front and back buffer and let games only draw to the back buffer.
- When FIFO present mode is active, SwapBuffer only swaps the two textures
- Set AcquireImage timeout to 0. GPU thread never blocks on presentation code anymore.
- When the FIFO present mode is active call vkPresentImage from handleTimedVsync instead of SwapBuffer
- When FIFO present mode is active AND 60 frames had desired pacing, take over vsync signalling from fixed timer

Why:
- No blocking on GPU thread, more opportunity to do work.

Why not:
- A lot of work for a theoretical improvement
- Games may receive sporadic VSync timing because of the pacing detection system, which is inaccurate
- Pacing detection is fragile in general, it relies on handleTimedVsync being called more often than the framerate, which may not always be the case. To implement this properly presentation would need to be moved to a new thread, which is difficult and unusual and may expose driver bugs


lbibass:
Any update on this? Seems like it would be useful for steam deck.

goeiecool9999:
when I wrote the PR description I oversold it's capabilities considerably, because I didn't fully understand the nuances of this change. I have edited the PR description to be more accurate.
That said. I don't think this will give considerable improvements on any platform, in hindsight it doesn't actually give Cemu new capabilities. And with that I am closing this PR.

PR #1025, opened 2023-11-19
Set macOS min version to 12.0 Monterey

shinra-electric:
I've noticed in Discord that the minimum macOS version is 12 Monterey, but the actual minimum version set on the app bundle is 10.10 Yosemite. 

This PR adds the minimum version key to Info.plist, with a MACOSX_MINIMUM_SYSTEM_VERSION var. 
This var is then set to 12.0 in CmakeLists.txt

The purpose of setting the minimum version is so that users on Big Sur or lower will not be able to open the app, and they will be informed that it can't be run on their OS and they will be prompted to update. 

This should hopefully reduce the number of issues reported on GitHub and Discord. 

Testing: 
Ensure the CI builds and check that the app bundle advertises the required OS to be 12 or later. 
Screenshot 2023-11-19 at 23 34 40

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -81,6 +81,7 @@ if (MACOS_BUNDLE)
 	set(MACOSX_BUNDLE_COPYRIGHT "Copyright © 2023 Cemu Project")
 
 	set(MACOSX_BUNDLE_CATEGORY "public.app-category.games")
+ 	set(MACOSX_MINIMUM_SYSTEM_VERSION "12.0")
 
 	set_target_properties(CemuBin PROPERTIES
 		MACOSX_BUNDLE true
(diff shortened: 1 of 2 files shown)


shinra-electric:
I'd like to check how the CI does, but I suspect it might fail as CMake only handles a few of the Info.plist variables:
https://cmake.org/cmake/help/latest/prop_tgt/MACOSX_BUNDLE_INFO_PLIST.html

The way around this is to just set a global variable somewhere.

shinra-electric:
The changes seems to have worked for the macOS bundle: 
Screenshot 2023-11-23 at 13 10 09

(I doubt if the ubuntu CI failure is related to this PR...)

Exzap:
thanks!

PR #1027, opened 2023-11-21
nsyshid: Fix SetProtocol and SetReport for Libusb Backend

deReeperJosh:
The initial implementation of the libusb backend didn't implement SetProtocol and SetReport, and these methods are important for the Skylanders Games as the portal device uses these methods to receive commands from the game, and then respond via the HIDRead methods.

I have tested this as working on my Mac, with Giants and Swap Force. Will also require testing via linux. At the moment I have hard coded the values in the libusb_control_transfer method, but I will attempt to find the correct enum values for this.

### DIFF
--- a/src/Cafe/OS/libs/nsyshid/BackendLibusb.cpp
+++ b/src/Cafe/OS/libs/nsyshid/BackendLibusb.cpp
@@ -694,22 +694,58 @@ namespace nsyshid::backend::libusb
 			return false;
 		}
 
-		// ToDo: implement this
-#if 0
-		// is this correct? Discarding "ifIndex" seems like a bad idea
-		int ret = libusb_set_configuration(handleLock->getHandle(), protocol);
-		if (ret == 0) {
-			cemuLog_logDebug(LogType::Force,
-							 "nsyshid::DeviceLibusb::setProtocol(): success");
-			return true;
+		struct libusb_config_descriptor* conf = nullptr;
+		libusb_device* dev = libusb_get_device(handleLock->GetHandle());
+		int getConfig = libusb_get_active_config_descriptor(dev, &conf);
+		if (getConfig == LIBUSB_SUCCESS)
+		{
+			for (uint8 i = 0; i < conf->bNumInterfaces; ++i)
+			{
+				int releaseSuccess = libusb_release_interface(handleLock->GetHandle(), i);
+				if (releaseSuccess < LIBUSB_SUCCESS)
+				{
+					cemuLog_logDebug(LogType::Force, "nsyshid::DeviceLibusb::SetProtocol(): Failed to release interface %i", i);
+					return false;
+				}
+			}
 		}
-		cemuLog_logDebug(LogType::Force,
-						 "nsyshid::DeviceLibusb::setProtocol(): failed with error code: {}",
-						 ret);
-		return false;
-#endif
+		else
+		{
+			cemuLog_logDebug(LogType::Force, "nsyshid::DeviceLibusb::SetProtocol(): Failed to Get Active Config");
+		}
 
-		// pretend that everything is fine
+		const int setConfig = libusb_set_configuration(handleLock->GetHandle(), protocol);
+		if (setConfig == LIBUSB_SUCCESS)
+		{
+			getConfig = libusb_get_active_config_descriptor(dev, &conf);
+			if (getConfig == LIBUSB_SUCCESS)
+			{
+				for (uint8 i = 0; i < conf->bNumInterfaces; ++i)
+				{
+					int detachSuccess = libusb_detach_kernel_driver(handleLock->GetHandle(), i);
+					if (detachSuccess < LIBUSB_SUCCESS && detachSuccess != LIBUSB_ERROR_NOT_FOUND &&
+						detachSuccess != LIBUSB_ERROR_NOT_SUPPORTED)
+					{
+						cemuLog_logDebug(LogType::Force, "nsyshid::DeviceLibusb::SetProtocol(): failed to detach kernel driver");
+						return false;
+					}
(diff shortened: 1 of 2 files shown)


deReeperJosh:
@ssievert42 fyi - this has worked on my Mac for the set protocol and set report methods, needed to release all interfaces, set the specified configuration then claim the interfaces for the specified configuration again in the set protocol, then in set report control transfers work :)

ssievert42:
Neat!

src/Cafe/OS/libs/nsyshid/BackendLibusb.cpp
```diff
@@ -694,22 +694,53 @@ namespace nsyshid::backend::libusb
...
-#if 0
-		// is this correct? Discarding "ifIndex" seems like a bad idea
-		int ret = libusb_set_configuration(handleLock->getHandle(), protocol);
-		if (ret == 0) {
-			cemuLog_logDebug(LogType::Force,
-							 "nsyshid::DeviceLibusb::setProtocol(): success");
-			return true;
+		struct libusb_config_descriptor* conf = nullptr;
+		libusb_device* dev = libusb_get_device(handleLock->GetHandle());
+		int ret = libusb_get_active_config_descriptor(dev, &conf);
```
ssievert42, line 699: This probably leaks some memory: The [docs](https://libusb.sourceforge.io/api-1.0/group__libusb__desc.html#ga425885149172b53b3975a07629c8dab3) say that `conf` must be freed with `libusb_free_config_descriptor()`.
However, the result of `libusb_get_device` does not need to be freed. The joys of programming in C :)
To cite the docs: "do not feel compelled to unreference it when you are done".

src/Cafe/OS/libs/nsyshid/BackendLibusb.cpp
```diff
@@ -694,22 +694,53 @@ namespace nsyshid::backend::libusb
...
+		}
+		else
+		{
+			cemuLog_logDebug(LogType::Force, "nsyshid::DeviceLibusb::SetProtocol(): Failed to Get Active Config");
+		}
+
+		const int setConfig = libusb_set_configuration(handleLock->GetHandle(), protocol);
+		if (setConfig == LIBUSB_SUCCESS)
+		{
+			libusb_get_active_config_descriptor(dev, &conf);
```
ssievert42, line 720: Same as above, plus conf should only be valid if this returns 0.

src/Cafe/OS/libs/nsyshid/BackendLibusb.cpp
deReeperJosh: good catch, have added that in now!

src/Cafe/OS/libs/nsyshid/BackendLibusb.cpp
```diff
@@ -694,22 +694,58 @@ namespace nsyshid::backend::libusb
...
+		int getConfig = libusb_get_active_config_descriptor(dev, &conf);
+		if (getConfig == LIBUSB_SUCCESS)
+		{
+			for (uint8 i = 0; i < conf->bNumInterfaces; ++i)
+			{
+				int releaseSuccess = libusb_release_interface(handleLock->GetHandle(), i);
+				if (releaseSuccess < LIBUSB_SUCCESS)
+				{
+					cemuLog_logDebug(LogType::Force, "nsyshid::DeviceLibusb::SetProtocol(): Failed to release interface %i", i);
+					return false;
```
ssievert42, line 708: Returning here is problematic: We still need to tell libusb to free the struct that `conf` points to!
Since you are only using `conf->bNumInterfaces` from that struct, how about calling `libusb_get_active_config_descriptor()`, making a copy of `conf->bNumInterfaces` and then immediately calling `libusb_free_config_descriptor()`?

src/Cafe/OS/libs/nsyshid/BackendLibusb.cpp
```diff
@@ -694,22 +694,58 @@ namespace nsyshid::backend::libusb
...
+		}
+		else
+		{
+			cemuLog_logDebug(LogType::Force, "nsyshid::DeviceLibusb::SetProtocol(): Failed to Get Active Config");
+		}
+
+		const int setConfig = libusb_set_configuration(handleLock->GetHandle(), protocol);
+		if (setConfig == LIBUSB_SUCCESS)
+		{
+			getConfig = libusb_get_active_config_descriptor(dev, &conf);
```
ssievert42, line 720: This (in its current form) causes us to lose all references to the struct that `conf` previously pointed to.
Just made a small test program, to confirm that I am not talking utter nonsense, and valgrind agrees with me that the following snippet leaks:
```c++
struct libusb_config_descriptor *conf = nullptr;
libusb_get_active_config_descriptor(dev, &conf);
libusb_get_active_config_descriptor(dev, &conf);
```

deReeperJosh:
Unfortunately having to abandon this PR - I have become extremely busy due to personal reasons, but if someone wanted to pick this work up in future and get these methods working for libusb hopefully this can be a starting point.

Exzap:
No problem and the effort is appreciated regardless. I recently got more USB portals so that I can help with this myself. Will see if I can pick this up.

Caspersonn:
What is the current state with skylander portal on linux?

deReeperJosh:
@Caspersonn I don't think the methods have been created yet to handle all skylander portal commands on linux yet so the games won't work just yet

Caspersonn:
Ooh that sucks, is there any way i could help?

deReeperJosh:
If you have dev experience I would try and have a go at trying to figure out the structure of libusb_control_transfer methods to send commands to the portal :)

kirbeh73:
Any updates?

PR #1028, opened 2023-11-22
Reduce default Deadzone value

j8r:
25% is far too much for most controllers, a saner default can be 15%.

I have several controllers to test on (even no-names) and I don't recall having ones above 10%, the worsts may be Xbox 360 controllers.

### DIFF
--- a/src/input/api/Controller.h
+++ b/src/input/api/Controller.h
@@ -130,7 +130,7 @@ public:
 
 	struct AxisSetting
 	{
-		AxisSetting(float deadzone = 0.25f) : deadzone(deadzone) {}
+		AxisSetting(float deadzone = 0.15f) : deadzone(deadzone) {}
 		float deadzone;
 		float range = 1.0f;
 	};


Squall-Leonhart:
defaults should be per stick,  right stick on some controls is 25+ %

j8r:
Yes it can, however most controllers outhere - Dualshocks/sense and Xbox controllers have decent quality joysticks.

High deadzone controllers exist, but of course they are not as pleasant to use.

The UI already allows to define a different deadzone per stick.

EgoBizarro:
After some time of usage, I have to go to 25% to not get any drift with my Xbox One Controller; even when it was new, I had to have 20% to never get false inputs.

j8r:
@EgoBizarro maybe it was required to recalibrate the neutral point in software instead of increasing the deadzone?
From what I understand the deadzone is useful when joysticks are loose (when one can move it wirhout any/few resistance).

I may be alone then, I have tens of different controllers, and 10 is enough on my side.
May be fine to leave as-is then.

Issue #886, opened 2023-06-30
[Feature Request] Make Alt+F4 close the Emulator


lunatic-gh:
So, in previous versions of Cemu (everything before the 2.0 Release) had alt+f4 work perfectly fine, closing the app as expected.
But now in 2, it does exactly... nothing ._.

I've read in other issues its supposed to be an issue on Steam Decks only, but i'm not using a Steamdeck, nor am i using Linux in any form (alltho its the same behavior on my arch machine).

I don't think it matters, but for completion i'll put my full System specs below:

Windows 11 Pro
AMD Ryzen 7 5800X3D
AMD Radeon RX 6900XT
16GB DDR4 RAM
Latest Cemu version (2.0-42)

chloecdn:
maybe i should add: this only appears when in a game. when just opening the emulator itself, i can close it with alt+f4 perfectly fine.

aem111607:
Not really a feature request as much as a bug

Squall-Leonhart:
the context for the window accelerators are broken in some cases when the emulator transitions to running a game from the library list, this is already known but it requires some time and effort to figure out how and why, this will likely get looked at in the future when the rest of the DPI scaling is also looked at.

chloecdn:
This is weird, because even if Cemu is ran with a specific game via arguments (Cemu.exe -f -g "path/to/game"), it will still refuse to close, the same way as when opened from the game library. But maybe it's because of how the emulator behaves when opening games from arguments, i don't know.

jonbarrow:
Have you tried a more recent release? I'm on Ubuntu 22.04 and alt+F4 closes the emulator for me just fine, even in games

chloecdn:
> Have you tried a more recent release? I'm on Ubuntu 22.04 and alt+F4 closes the emulator for me just fine, even in games

Yes, i'm always on the latest version, right now on 2.0-45, and pressing alt+f4 does nothing when ingame. I'd take a wild guess and say it's a windows-only issue? Or maybe even Windows11-Only?

spudpiggy:
I'm on Windows 10 running 2.0-47 and I can't alt+f4 either.
Alt+Space then C is a lesser known way to close a window, and I can confirm that it works for me. There's a workaround, although I know it isnt ideal cus muscle memory.

biblicabeebli:
Summary/Update

Bug still present in 2.0-59.
Bug appears to be Wandows only.
Bug occurs only when a game is being actively emulated.
[...]

PR #1035, opened 2023-12-01
make altf4/cmd-q propogate to the parent after MainWindow::CreateCanvas has set focus

Squall-Leonhart:
Makes the force exit key combo for various OS work when Cemu is running a game as a window or fullscreen


Squall-Leonhart:
closes #886

ExpatUK:
Can't wait for this to be merged. This behaviour is a huge QoL fix when using launchers from the couch.

PR #1039, opened 2023-12-09
Mac: Add wua filetype to info.plist

shinra-electric:
Declaring support for `wua` filetypes in the info.plist will associate wua files with Cemu. 

Needs testing with the CI: 

- [x] Viewing a `wua` file in Finder should display a document icon with a Cemu icon superimposed on it. 

Example from OpenEmu: 
Screenshot 2023-12-09 at 12 16 42

- [x] Double-clicking on a `wua` file should now cause Cemu to open. 

The action we would expect is to run the `wua` on launch. 
Edit: Or else to first add to the games list and then launch...
Edit 2: Cemu opens, but nothing else happens. Functionality can be added in a separate PR.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -82,7 +82,7 @@ if (MACOS_BUNDLE)
 
 	set(MACOSX_BUNDLE_CATEGORY "public.app-category.games")
  	set(MACOSX_MINIMUM_SYSTEM_VERSION "12.0")
-  	set(MACOSX_BUNDLE_TYPE_EXTENSIONS "wua")
+  	set(MACOSX_BUNDLE_TYPE_EXTENSION "wua")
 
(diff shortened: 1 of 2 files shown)


shinra-electric:
I checked the CI: 

- The association is working
- The icon is the full Cemu icon, not the document one I was expecting
- Double-clicking the wua file simply opens Cemu. Will need to add a new function to handle this properly.

Screenshot 2023-12-09 at 13 00 17

shinra-electric:
Removing explicit mention of `cemu.icns` gives the expected document icon: 

Screenshot 2023-12-09 at 13 16 20

I think maybe the function for opening the file can be done in another PR.

Exzap:
I assume the issue with games not launching is that Cemu expects the `-g` option in front of the launch path. But I agree that this can be solved as a separate PR. Thanks!

PR #1042, opened 2023-12-13
Latte code clean up + Proper support for SAMPLE_LB shader instruction

Exzap:
This PR is mostly for cleaning up old OpenGL mannerisms in the shared renderer code. While doing this I noticed that the SAMPLE_LB shader instruction was never correctly implemented so I added that now. The only shaders that I could find which use this instruction are the terrain shaders in BotW. But I couldn't really see a visual difference after the fix.

Also for OpenGL we now have some wrappers to emulate DSA-style API even if the driver doesn't provide the actual extension. This simplifies the GL renderer code a bit since we don't have to worry about altering the render state when creating or updating textures


Exzap:
Thanks for letting me know. It crashes on my end too, I'll look into it

Exzap:
d2ba4e65c5d7802e53faac21922be1bf9347f325 should fix it.

MaranBr:
Confirmed. The game is working again, thank you!

PR #1047, opened 2023-12-17
Play bootSound.btsnd while shaders/pipelines are compiling

goeiecool9999:


Adamillo:
Tested on Windows but I hear nothing from bootSound.btsnd, is this option enabled by default?

Adamillo:
And I found another bug. XAudio2 does not work properly with this PR no matter what latency you set it to

goeiecool9999:
Latest version should work on all API's and latency settings.
Unrelated but I think before merging these changes it's probably a good idea to make this functionality disabled by default. I think it would be startling for users who update to suddenly hear audio when loading a game. When someone has short loading times it will play just a little bit of the audio and then abruptly cut out which also isn't very polished. I also heard that audio dropouts like that can blow up speakers in exceptional cases.

Adamillo:
The audio drop out is definitely a problem that could make your PR get blocked from merging (I say could, because I am not a maintainer). Is there a way to make the audio fade out smoothly without complicating the code?

goeiecool9999:
I may have overstated how bad of an issue it is. Going from sound to silence like that happens every time you pause a YouTube video. So I recon anyone dealing with audio systems that could actually be damaged from that know to be extra careful when hooking up computer software to it. Main concerns are the fact that it's unexpected and unpolished.

Adamillo:
I see, makes sense. I was a little worried with the issue but I see what you mean now

Adamillo:
Any plans about reviews/merges for this PR?

src/audio/XAudio27API.cpp
```diff
@@ -33,8 +33,8 @@ XAudio27API::XAudio27API(uint32 device_id, uint32 samplerate, uint32 channels, u
 	m_wfx.Format.nChannels = channels;
 	m_wfx.Format.nSamplesPerSec = samplerate;
 	m_wfx.Format.wBitsPerSample = bits_per_sample;
-	m_wfx.Format.nBlockAlign = (m_wfx.Format.nChannels * m_wfx.Format.wBitsPerSample) / 8; // must equal (nChannels � wBitsPerSample) / 8
-	m_wfx.Format.nAvgBytesPerSec = m_wfx.Format.nSamplesPerSec * m_wfx.Format.nBlockAlign; // must equal nSamplesPerSec � nBlockAlign.
+	m_wfx.Format.nBlockAlign = (m_wfx.Format.nChannels * m_wfx.Format.wBitsPerSample) / 8; // must equal (nChannels × wBitsPerSample) / 8
+	m_wfx.Format.nAvgBytesPerSec = m_wfx.Format.nSamplesPerSec * m_wfx.Format.nBlockAlign; // must equal nSamplesPerSec × nBlockAlign.
```
goeiecool9999, line 37: I changed these characters to UTF-8. These were the only non-UTF8 characters in the file but GitHub doesn't detect encoding changes.

GabIsAwesome:
is this ever getting merged?? 🫤

Adamillo:
Any updates on this PR?

Exzap:
Can you add a fade-out effect instead of abruptly stopping the audio? Similar to how it is on console:
[TWW_Intro.webm](https://github.com/user-attachments/assets/66974d9d-7ace-4464-a8b7-053b8102ff64)

goeiecool9999:
I didn't see your comment until now. I had that idea but didn't implement it because I thought it would be too complicated. I'll look into it.

goeiecool9999:
Latest commit has a 10 second linear fadeout. 10 seconds is likely too long and I'm also curious what type of curve the Wii U uses for it's fadeout.
Because the life of the playback thread now extends beyond the duration of the cache loading screen I've changed the structure of the code to use an extra class which manages the thread. This gives a guarantee that when main returns and the destructor of the global is called the thread joins properly/timely.

goeiecool9999:
I've done my best to approximate the fade-out. To get more accurate someone would need to make a homebrew that uses a pure tone as a bootsnd or reverse engineer that part of the system software. But I think this is good for now.
I also wonder if titles expect a certain syscall to trigger the end the bootscreen and start the fade-out (perhaps GX2 initialisation?) The new structure allows us to trigger the fadeout anywhere so if we figure that out we could move it there instead.

PR #1048, opened 2023-12-18
Update vcpkg.json

Jim-dll:


Exzap:
We generally handle vcpkg and library updates internally.

PR #1051, opened 2023-12-23
windows-build-error-fix

fuhongcui:
update vcpkg for windows build


Squall-Leonhart:
This is a duplicate of https://github.com/cemu-project/Cemu/pull/1048, Exzap prefers to handle vcpkg updates internally.

fuhongcui:
> What exactly are you fixing here? Cemu already builds on Windows.

The submodule vcpkg version in the current code has dependencies on libraries that are no longer present on the server.

image
image

Squall-Leonhart:
Ah, i see, this only affects mingw builds then, but not msvc builds.

fuhongcui:
> > Ah, i see, this probably only affects mingw builds then, but not msvc builds - i just built cemu just fine with msvc without any failures.
> 
> 
> 
> It does in fact affect MSVC use.
> 
> Building may be fine, CMake configuration is not, because the vcpkg download keeps failing. You probably already have the dependency downloaded from an earlier configuration

Yes, that's the issue I've encountered.

Exzap:
Superseded by #1064

PR #1064, opened 2024-01-13
Update vcpkg and libraries

Exzap:
| Libname   | Previous version      | New version |
|-----------|------------|------------|
| pugixml   | 1.13.0     | 1.14       |
| zlib      | 1.2.13     | 1.3        |
| zstd      | 1.5.4      | 1.5.5      |
| libzip    | 1.9.2      | 1.10.1     |
| rapidjson | 2022-06-28 | 2023-07-17 |
| SDL2      | 2.26.5     | 2.28.5     |
| boost     | 1.81.0     | 1.83.0     |
| fmt       | 9.1.0      | 10.2.1     |
| hidapi    | 0.13.1     | 0.14.0     |
| libpng    | 1.6.39     | 1.6.40     |
| glm       | 0.9.9.8    | 0.9.9.8    |
| glslang   | 12.1.0     | 13.1.1     |
| zstd      | 1.5.4      | 1.5.5      |
| openssl   | 3.1.0      | 3.2.0      |
| curl      | 7.88.1     | 8.5.0      |
| libusb    | 1.0.26     | 1.0.26.11791 |
| wxWidgets | 3.2.2.1    | 3.2.4      |

We use vcpkg by default for package management. So these changes only apply when compiling Cemu without explicitly turning vcpkg off. (e.g. `-DENABLE_VCPKG=off`)

If you are compiling Cemu locally you might run into issues during cmake bootstrapping vcpkg. In that case try deleting the vcpkg binary from `dependencies/vcpkg/` or run `dependencies/vcpkg/bootstrap-vcpkg`


amayra:
this typo or glm lib never get update ?

Exzap:
0.9.9.8 is still the last release https://github.com/g-truc/glm/releases

Jim-dll:
I don't know if my comment will be useful but glm has been updated.

Exzap:
We only very lightly make use of glm so for our needs there is no real advantage of updating immediately. But we did run into an issue (now resolved) where some Linux distros sourced glm from the git directly instead of 0.9.9.8 and that caused some incompatibilities. At any rate, we'll handle the glm update with the next bulk dependency update

Issue #1065, opened 2024-01-14
Cemu  crashes on opening on Mac - "Library missing" 


MetallicaPixel:
Cemu 2.0-62

Cemu will crash when trying to launch it.

macOS 13.6.3 
Intel Mac

Crash report -

Crashed Thread:        0

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000

Termination Reason:    Namespace DYLD, Code 1 Library missing
Library not loaded: /usr/local/opt/libusb/lib/libusb-1.0.0.dylib
Referenced from: <DCBE5229-ADF0-39A9-BE0F-F52230D1EF1F> /Users/USER/*/Cemu.app/Contents/MacOS/Cemu
Reason: tried: '/usr/local/opt/libusb/lib/libusb-1.0.0.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/usr/local/opt/libusb/lib/libusb-1.0.0.dylib' (no such file), '/usr/local/opt/libusb/lib/libusb-1.0.0.dylib' (no such file), '/usr/local/lib/libusb-1.0.0.dylib' (no such file), '/usr/lib/libusb-1.0.0.dylib' (no such file, not in dyld cache)
(terminated at launch; ignore backtrace)

Thread 0 Crashed:
0   dyld                          	    0x7ff81806dc52 __abort_with_payload + 10
1   dyld                          	    0x7ff818087fd7 abort_with_payload_wrapper_internal + 82
2   dyld                          	    0x7ff818088009 abort_with_payload + 9
3   dyld                          	    0x7ff81800c8f0 dyld4::halt(char const*) + 375
4   dyld                          	    0x7ff818009b71 dyld4::prepare(dyld4::APIs&, dyld3::MachOAnalyzer const*) + 4526
5   dyld                          	    0x7ff8180083bd start + 1805

Thread 0 crashed with X86 Thread State (64-bit):
  rax: 0x0000000002000209  rbx: 0x0000000000000000  rcx: 0x00007ff7b9dc6518  rdx: 0x00007ff7b9dc6980
  rdi: 0x0000000000000006  rsi: 0x0000000000000001  rbp: 0x00007ff7b9dc6560  rsp: 0x00007ff7b9dc6518
   r8: 0x00007ff7b9dc6580   r9: 0x0000000000000000  r10: 0x000000000000007d  r11: 0x0000000000000246
[...]

PR #1066, opened 2024-01-16
Fix macos missing dylib file

qurious-pixel:
Fixes https://github.com/cemu-project/Cemu/issues/1065


JaimieVandenbergh:
Tested and launches fine on macOS 14/M1.

Exzap:
The overlay port is probably not necessary if we just exclude libusb from vcpkg.json like this:
```
    {
      "name": "libusb",
      "platform": "!osx"
    },
```

PR #1071, opened 2024-01-18
Add support for portable directory without build flag.

Steveice10:
Removes the need for a compile-time portable flag, instead supporting using either a portable or system directory at runtime.

The behavior is as follows:
* First, on all platforms, check for the directory "portable" in the same directory as the executable (or app bundle on macOS, if running from an app bundle). If that directory exists, use it for all paths.
  * This is how most other emulators approach this, e.g. the Citra/Yuzu "user" directory, the Ryujinx "portable" directory, etc.
* Otherwise, match the old behavior:
  * Windows: Use the directory the executable is in.
  * Linux: Use XDG directories where possible, plus the other directories from wxWidgets.
  * macOS: Use the standard directories from wxWidgets.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -75,7 +75,7 @@ jobs:
         
     - name: "cmake"
       run: |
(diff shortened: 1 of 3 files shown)


Exzap:
On a quick glance this looks good. I will give this a test run soon and merge it if no issues pop up. Thanks!

Steveice10:
Any updates on this?

Exzap:
Sorry for the wait! LGTM

PR #1072, opened 2024-01-18
Vulkan Headers: Update from 1.3.245 to 1.3.275

Jim-dll:
I updated Vulkan Headers because I think it could provide better performance.
I had some problems that's why it's a bit chaotic.


Exzap:
We handle dependency and library updates internally so I will close this, but the effort is appreciated regardless.
That said, updating vulkan headers will not improve performance, they only describe the API interface to the Vulkan loader and driver.

Issue #1075, opened 2024-01-19
glslang has reduced it's public headers


FireBurn:
glslang only installs it's public headers now:

https://github.com/KhronosGroup/glslang/pull/3397 https://github.com/KhronosGroup/glslang/commit/1dcb072cda091180a5b8b03c030bcbe83a54f8e2

Which means #include <glslang/Include/Types.h> is no longer available

it's used in:

src/Cafe/HW/Latte/Renderer/Vulkan/VKRPipelineInfo.cpp
src/Cafe/HW/Latte/Renderer/Vulkan/VulkanRenderer.cpp

PR #1076, opened 2024-01-19
Vulkan: Don't use glslang internal headers

FireBurn:
With these changes I'm able to build against glslang 1.3.275

I think this fixes https://github.com/cemu-project/Cemu/issues/1075


Exzap:
Thanks!

PR #1080, opened 2024-01-22
Update CMake to 3.28.1

Jim-dll:


Exzap:
If you make changes to the build scripts or dependencies then please elaborate on why you are doing so. Either way I dont think setting a maximum version is really beneficial since cmake generally has very good backwards compatibility.

PR #1101, opened 2024-02-24
add llvm package for Fedora

leifliddy:
Fedora requires the llvm compiler in order to compile Cemu with `clang`. 
Installing `clang` will install `llvm-libs` but not `llvm`.
Interestingly enough it also installs `gcc` and `gcc-c++` as dependencies -- so we don't need to install those separately (if compiling with `gcc`) 
```
========================================================================================================================================
 Package                                   Architecture          Version                                   Repository              Size
========================================================================================================================================
Installing:
 clang                                     x86_64                17.0.6-2.fc39                             updates                 76 k
Installing dependencies:
 ....
 gcc                                       x86_64                13.2.1-3.fc39                             fedora                  34 M
 gcc-c++                                   x86_64                13.2.1-3.fc39                             fedora                  13 M
 ....
 llvm-libs                                 x86_64                17.0.6-3.fc39                             updates                 27 M

```


Exzap:
Thanks!

PR #1112, opened 2024-03-10
Fix linux OpenGL VSync crash

goeiecool9999:


src/Cafe/HW/Latte/Renderer/OpenGL/OpenGLRenderer.cpp
```diff
@@ -241,6 +244,18 @@ void LoadOpenGLImports()
 #undef GLFUNC
 #undef EGLFUNC
 }
+
+#if BOOST_OS_LINUX
+// dummy function for all code that is statically linked with cemu and attempts to use eglSwapInterval
+// used to suppress wxWidgets calls to eglSwapInterval
+// Real eglSwapInterval is a function pointer defined as eglSwapIntervalCemu
```
Exzap, line 251: This comment needs updating?

goeiecool9999: Indeed

PR #1114, opened 2024-03-10
Fix crash when setting console language while game is running

### LINKED ISSUE #816: [Bug Report] Changing system language while a game is running crashes Cemu.
Hello, if you change the console language (Options -> Console language) while a title is running (e.g. homebrew), Cemu crashes immediately.

The problem has been happening for several versions, including the most recent 2.0-39 (experimental).
Thank you.

goeiecool9999:
Fixes #816

### DIFF
--- a/src/gui/MainWindow.cpp
+++ b/src/gui/MainWindow.cpp
@@ -1019,8 +1019,11 @@ void MainWindow::OnConsoleLanguage(wxCommandEvent& event)
 	default:
 		cemu_assert_debug(false);
 	}
-	m_game_list->DeleteCachedStrings();
-	m_game_list->ReloadGameEntries(false);
+	if(!IsGameLaunched())
+	{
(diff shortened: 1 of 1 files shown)


src/gui/MainWindow.cpp
```diff
@@ -1019,8 +1019,11 @@ void MainWindow::OnConsoleLanguage(wxCommandEvent& event)
 	default:
 		cemu_assert_debug(false);
 	}
-	m_game_list->DeleteCachedStrings();
-	m_game_list->ReloadGameEntries(false);
+	if(!IsGameLaunched())
```
Exzap, line 1022:
```suggestion
	if (m_game_list)
```
Probably better to just check if the list exists?

Exzap:
Since you are at it, maybe gray out the console language option once a game is running. Changing it mid-emulation is generally a bad idea and I dont think our code gracefully handles it

Issue #911, opened 2023-07-23
Interpreter mode no longer activating


retrobenny:
#907 

On Windows, Interpreter mode is not working in build 2.0-45 from Actions as I didn't notice releases had it too until just now.
When switching CPU mode on Game Config for a game such as Super Smash Bros. 4 (v304/1.1.7) it just runs in one of the recompiler modes instead of interpreter.
After noticing this, I randomly tried setting it on version 1.26.2 of Cemu that I had already and that one ran on interpreter mode.

I read that the interpreter is needed for correct read/write breakpoints and it definitely worked for that on 1.26.2.
If I should, I will try testing again from the releases build when I get the time to.

Edit: The initial 2.0 release version from releases works on interpreter as well, time to bisect I guess.
And I can confirm the issue happens on the releases copy of 2.0-45 as well.

retrobenny:
Interpreter selection is working in 2.0-41 so it comes down to commits between that and 2.0-45 where interpreter stopped working, as it just boots into recompiler instead.
Bisecting narrowed down, now to fully bisect and edit in the results.

Exzap:
I'm using interpreter mode a lot of debugging and haven't noticed any issues. How are you enabling it? Via command line `--force-interpreter` or via game profile option? How do you know it's running in recompiler mode? Does it say that in the log.txt?

retrobenny:
I set it to use interpreter on the game via the GUI on the game profile menu.
I will have to check the log file when I next get the chance to test again.

Though I can tell that it was still using recompiler because performance was not at all impacted on 2.0-45 and from the game not getting stuck on the menu for a long time like it does on 2.0-44 using interpreter, and read/write breakpoints actually went to the correct asm when interpreter was set on 2.0-44 while 2.0-45 still went to an incorrect spot.

Getting stuck waiting for the main menu is possibly because of starved resources through a single core used for all threads, setting cpu timing to 8x for a moment allows me to continue through the main menu then load VS mode and start a match.
Not sure why it can't just run it in a balanced way, costing more performance during loading but avoiding getting stuck there.

Pausing in a match has UI elements missing which never appear even after changing the cpu timing setting and trying to end the match on pause ends up soft-locking the game.
The game runs poorly as expected for interpreter on 2.0-44 but also is sped up with frame skips/drops and laggy audio, it even has bad audio in the rare case of reaching 60fps on a lighter stage with two of the same character.

PR #1115, opened 2024-03-11
CafeSystem: call PPCRecompiler_init after SetupExecutable

goeiecool9999:
Fixes: #911
This makes sure that the gameprofile is loaded when the recompiler initialises and uses the selected mode.


Exzap:
Nice catch!

Issue #1013, opened 2023-10-30
Application has no proper title and ignores SIGTERM


willbelr:
In the Linux version (cemu v2.0.59-1), I use a script to close all running emulators at once, and I noticed that cemu is not affected by the `killall` command. Instead, it require the full bin path;

```
killall cemu  # cemu: no process found
killall /usr/bin/cemu  # process found, but won't close
```

When running cemu, using `ps -eo comm` shows that it is actually called `MainThread`, instead of `cemu`.

Regarding the signals, oddly cemu won't close with `SIGTERM`, and instead requires `SIGQUIT` (which also dumps the core);

```
killall /usr/bin/cemu  # process gets the signals, but it keeps running
killall -s SIGQUIT /usr/bin/cemu  # this works
killall -s SIGQUIT MainThread  # also works
```

PR #1116, opened 2024-03-11
Linux: Exit on SIGTERM

goeiecool9999:
Addresses one concern of #1013
The issue also points out that linux (or at least utilities like killall) look(s) at the thread name and not the executable name.
We could change the name of the main thread to something ``cemu`` on linux. Which seems sensible because threads that don't set their name inherit the name of the thread that created them. So many Cemu threads are now called MainThread, which also isn't very helpful. But that's for another PR I think.


Exzap:
Changing main thread name to `cemu` sounds good to me

PR #1118, opened 2024-03-11
InputSettings: Fix controller type counter to restore WPAD limit

goeiecool9999:
get_emulated_controller_types always returned zero for the number of wpads. It _was_ able to count VPAD's because EmulatedController::VPAD is the only enum value that evaluates to false.
Additionally, create all pages of the input settings menu when it is first opened, so that the counter can see controllers on different pages before the user viewed those pages.
This restores the WPAD limit of 7.

### DIFF
--- a/src/gui/input/InputSettings2.cpp
+++ b/src/gui/input/InputSettings2.cpp
@@ -745,11 +745,13 @@ void InputSettings2::on_emulated_controller_dropdown(wxCommandEvent& event)
 	wxWindowUpdateLocker lock(emulated_controllers);
 
 	bool is_gamepad_selected = false;
+	bool is_wpad_selected = false;
 	const auto selected = emulated_controllers->GetSelection();
 	const auto selected_value = emulated_controllers->GetStringSelection();
 	if(selected != wxNOT_FOUND)
 	{
 		is_gamepad_selected = selected_value == to_wxString(EmulatedController::type_to_string(EmulatedController::Type::VPAD));
+		is_wpad_selected = !is_gamepad_selected && selected != 0;
 	}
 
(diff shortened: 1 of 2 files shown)


Squall-Leonhart:
whats the chance of this fixing pro pads in black ops 2?

Exzap:
> Additionally, create all pages of the input settings menu when it is first opened
I didn't look at the code yet but is this change necessary? As far as I remember this was a deliberate optimization by Petergov to make the loading of the input window faster. Giving your PR a quick test I can indeed see that opening the window now takes about half a second on my system, whereas it used to be near instant.

goeiecool9999:
> > Additionally, create all pages of the input settings menu when it is first opened
> 
> I didn't look at the code yet but is this change necessary?

The function that counts the types of currently configured controllers, `InputSettings2::get_emulated_controller_types()`, uses `page_data->ref().m_controller->type();` So if you for example set controller 7 and 8 to a Wii U gamepad, then close and reopen the window, then only page one is initialised so when the dropdown calculates what controller types should be selected it can't see that on pages 7 and 8 a Wii U game pad was configured and lets you select another one. Only after the user opened each page individually does get_emulated_controller_types return accurate results.

To fix this problem without initialising every page when opening the input settings would require a bit of more complicated restructuring.

goeiecool9999:
It turns out that InputManager has the same function that the input settings window has, but instead of looking at the pages it looks at the actual state of InputManager. That InputManager function was unused. I think someone prepared this function to replace the InputSettings function but forgot to actually patch it in. I've now done that. Hopefully that's faster than initialising all pages. (which was is instant on linux so I didn't notice)

Issue #573, opened 2022-12-12
Wayland HiDPI renders at double the size of the window


rbutoi:
Trying out the native Wayland build (using Sway 1.7 myself), I found that the game would render twice as large, as the screenshot shows: ![2022-12-12-183824_screenshot](https://user-images.githubusercontent.com/5759233/207183270-a3716d40-119c-46e3-9a7d-a2cecee8b6a9.png)

My main monitor is a 4K monitor with 2x scaling in Sway. This is on the latest HEAD build, c78b3da. I've tried setting `GDK_SCALE` but it seems to be ignored.

Is this happening to anyone else? My current workaround is to run on the preview version's AppImage that runs in Xwayland.

rbutoi:
The fix at https://github.com/SSimco/Cemu/tree/waylandcanvassize works for me!

MrRinkana:
I was testing this on a build from 072c18a and I can confirm this behaviour. 

If the game is fullscreened with f11 the issue disappears, and everything is visible (while in fullscreen).

There is some weirdness though, especially in the case of multiple monitor with different scaling:
My setup is one monitor with 100% scaling and one with 175%.

If the cemu window is created on the 100% scaling screen it will be correctly scaled, however maximising and then "pulling the window down" from maximised (back to starting size) it will about 66% of the time bug out and scale incorrectly. Redoing this procedure (maximising and then "unmaximising") can fix the issue and return to proper scaling. Maximised always scales correctly on this screen (100% scaling).

The window can be dragged from the 100% to the 175% scaled screen, keeping correct scaling, if it was in that state on the 100% monitor (the game is within the window).

Maximising, toggling f11 twice or creating the cemu window while on the 175% screen will always result in incorrect scaling. Maximising and unmaximising does not fix the scaling.

However the window can be dragged back to the 100% scaled screen and fixed there by repeating the maximising/unmaximising process until it happens to scale correctly again.


This might have been a confusing text, I can maybe record the behaviour later if needed but I am not sure how helpful my observations are. In essence things are a bit quirky on this kind of setup, but the quirks might be the fault of kwin. Getting it to scale correclty on single monitor setups will likely fix all these quirks aswell.

aaron-rust:
[...]

PR #1123, opened 2024-03-16
Wayland: use physical units on subsurface resize

goeiecool9999:
potential fix for: #573
I've had roughly this patch shelved in my IDE since june 2023 but I never submitted it because I don't have a HiDPI wayland DE setup to test with.
Could someone test if this fix works?


myownfriend:
I just tried it out and the game rendered outside of the window and stretched larger than both of my monitors.

PR #1125, opened 2024-03-16
Improve current Exception Handler

EpicUsername12:
# Purpose 

This PR aims to improve the CEMU exception handler

# Features

- [x] Add a game crash message dialog before exiting
- [x] Add a CEMU crash message dialog before exiting
- [ ] Copy ``log.txt`` to clipboard
  - [x] Win32
  - [ ] Linux
  - [ ] MacOS
- [ ] Make sure it works and doesn't break anything else
  - [x] Win32 debug/release
  - [ ] POSIX  debug/release
  
# Description

The changes spawn a thread on CafeSystem initialization that will continously read a boolean and yield if the boolean is ``false``.

This boolean is set to ``true`` if a unhandled exception occurs, shutting down CafeSystem (to avoid the game still running in the background and other synchronization issues occuring) then displaying the message boxes.

# Pictures

| image | 
|:--:| 
| **Example CEMU crash** (forced crash in H264 decoder) |

### DIFF
--- a/src/Cafe/CafeSystem.cpp
+++ b/src/Cafe/CafeSystem.cpp
@@ -25,6 +25,7 @@
 #include "Common/FileStream.h"
 #include "GamePatch.h"
 #include "HW/Espresso/Debugger/GDBStub.h"
+#include "Common/ExceptionHandler/ExceptionHandler.h"
 
 #include "Cafe/IOSU/legacy/iosu_ioctl.h"
 #include "Cafe/IOSU/legacy/iosu_act.h"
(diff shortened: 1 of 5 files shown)


Crementif:
This seems like a really nice feature to have! I do think that telling users to report crashing issues to GitHub is pretty risky. Generally, these (recurring) crashes are more often a user setup or game files error then anything else. We currently also have plenty of issues and don't really have enough maintainers to manage/fix even more.

I think this referral should really just be to the troubleshooting channel in our discord or on our subreddit, if anything at all. I'm kinda worried that even those places will see too big of an influx in users, especially if you don't give them any tips on what they could do themselves before asking for help.

The ideal solution in my opinion (also have spoken a bit about this with Exzap) is that:
1. It should tell the user if it was a game crash or a Cemu crash (aka vulkan API errors, other common exceptions I can't think off), if possible
2. It should tell the users to check if they are running the latest updates of those games and to check their files using the title manager for corruption issues, to check if their hardware is supported and their drivers are up-to-date and tells them to check the game's compatibility rating on the wiki BEFORE asking for help in the troubleshooting channel.

EpicUsername12:
> 1. It should tell the user if it was a game crash or a Cemu crash (aka vulkan API errors, other common exceptions I can't think off), if possible

- It currently says whether it is a game crash or a CEMU crash!

---

> I do think that telling users to report crashing issues to GitHub is pretty risky. Generally, these (recurring) crashes are more often a user setup or game files error then anything else. We currently also have plenty of issues and don't really have enough maintainers to manage/fix even more.

- I agree with that

> 2. It should tell the users to check if they are running the latest updates of those games and to check their files using the title manager for corruption issues, to check if their hardware is supported and their drivers are up-to-date and tells them to check the game's compatibility rating on the wiki BEFORE asking for help in the troubleshooting channel.

-  I will try to add some of these features in future commits

Adamillo:
Are there any updates about this PR?

PR #1127, opened 2024-03-17
Improve setting of dpd_enable_fg

capitalistspz:
`dpd_valid_fg` was  (and probably still is) reported incorrectly. 

From testing on my own Wii U, I observed:
- `1` is hard to get consistently
- `2` is returned when aiming normally
- `-1` when only one IR cluster is visible
-  `0`  when not aiming at the sensor bar at all and IR position values no longer change

I haven't tested in any games on Cemu (my Wiimotes refuse to connect to my PC).
Hopefully other people can test it.

### DIFF
--- a/src/input/api/Controller.h
+++ b/src/input/api/Controller.h
@@ -2,6 +2,7 @@
 
 #include "input/InputManager.h"
 #include "input/motion/MotionSample.h"
+#include "input/api/ControllerState.h"
 
 namespace pugi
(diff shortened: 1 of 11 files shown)


Fancy2209:
imagem
this fixes the pointer in Little Inferno

Exzap:
I think this should be good to merge? Or is there any reason why this is still a draft

PR #1130, opened 2024-03-18
Updated the DXGI version to 1.6

Jim-dll:
I made this change realizing that the latest DXGI version was 1.6. I don't know if this is useful. Otherwise, sorry for the inconvenience.


Exzap:
#1048 #1072 #1080
Not going to repeat myself

PR #1132, opened 2024-03-22
Vulkan: several fixes and refactors

goeiecool9999:
* make surfaces entirely managed by SwapchainInfoVK (and actually destroy the surfaces in destructor)
 * destroy temporary surface for device enumeration
 * simplify SwapchainInfoVk constructor
 * change drawbackbuffer render pass load operation to OP_LOAD, so that when the swapchain image was cleared the black background is kept (formally, it always was in practice)
 * fix structure type enum
 * move queue family code back into VulkanRenderer

### DIFF
--- a/src/Cafe/HW/Latte/Renderer/Vulkan/SwapchainInfoVk.cpp
+++ b/src/Cafe/HW/Latte/Renderer/Vulkan/SwapchainInfoVk.cpp
@@ -1,15 +1,34 @@
 #include "SwapchainInfoVk.h"
 
 #include "config/CemuConfig.h"
+#include "gui/guiWrapper.h"
 #include "Cafe/HW/Latte/Core/Latte.h"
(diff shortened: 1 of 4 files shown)


amayra:
can we have "VK_KHR_synchronization2" as option in settings

Exzap:
> can we have "VK_KHR_synchronization2" as option in settings

There isn't really a point to it being available as an option. The reason we want to use it is that it's just a **tiny** bit faster. But if we make the code path dynamic that introduces branching overhead and it would offset the benefit while also making the code a lot more complicated.

Exzap:
On a side note including anything from `gui` in non-GUI code is something we really want to move away from. The Vulkan renderer shouldn't be tightly coupled to our wxWidgets/UI code and in the long term we should find a more abstract interface to handle the canvas. I know you just moved things around and didn't introduce this, just a FYI

PR #1134, opened 2024-03-24
DebugSymbolStorage: fix infinite loop

goeiecool9999:
This fixes a bug where ClearRange goes into an infinite loop when length is not divisible by 4. Because length is unsigned it wraps around to UINT32_MAX and never becomes zero. Because of this bug, if a user disables a graphic pack like "Graphics" for botw while the game is running length is equal to 9 and the UI thread freezes and never recovers.

### DIFF
--- a/src/Cafe/HW/Espresso/Debugger/DebugSymbolStorage.h
+++ b/src/Cafe/HW/Espresso/Debugger/DebugSymbolStorage.h
@@ -43,16 +43,14 @@ public:
 		return t;
 	}
 
-	static void ClearRange(MPTR address, uint32 length)
+	static void ClearRange(MPTR base, uint32 length)
 	{
 		s_lock.lock();
-		while (length > 0)
+		for (MPTR address = base; address < base + length; address += 4)
 		{
 			auto itr = s_typeStorage.find(address);
 			if (itr != s_typeStorage.end())
 				s_typeStorage.erase(itr);
-			address += 4;
-			length -= 4;
 		}
 		s_lock.unlock();
 	}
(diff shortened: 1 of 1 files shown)


Crementif:
Huh, seems like a good find!

Exzap:
This will overflow when `base` + `length` equals 0x1'00000000 (e.g. clearing the last dword of the address space with `base = 0xFFFFFFFC length = 4`) In practice there shouldn't really be any clears requested in high address space but it still should be either fixed or caught with an assert

goeiecool9999:
I have fixed the overflow. I had to brush up on my integral promotion rules but this should be right (I checked).

src/Cafe/HW/Espresso/Debugger/DebugSymbolStorage.h
```diff
@@ -45,12 +45,17 @@ class DebugSymbolStorage
...
+			return;
 		s_lock.lock();
-		while (length > 0)
+		for (;;)
 		{
 			auto itr = s_typeStorage.find(address);
 			if (itr != s_typeStorage.end())
 				s_typeStorage.erase(itr);
+			// break if length has surpassed zero and wrapped around
+			if (length - (uint32)4u > length)
```
Exzap, line 57: either `(uint32)4` or `4u` is enough, you dont need both.

But also why not just do `if (length < 4)`

goeiecool9999: > you dont need both.

I knew that but I was thinking "an unsigned integer is not guaranteed to be 32-bit"
> But also why not just do `if (length < 4)`

You make a good point...

PR #1138, opened 2024-03-25
coreinit: fix race condition that causes crash

goeiecool9999:
On linux cemu sometimes randomly crashes on startup. Looking at the coredump in a debugger showed that it crashed in emplace_back for a vector because of a double free. Write access to a vector is not thread-safe, but when cemu is in multi-core recompiler mode when the timing is just right two or more threads simultaneously emplace in the vector. Added a mutex to protect access.


Exzap:
The timing on this race condition is extremely narrow, so it's interesting that you got it to trigger. Thanks for the fix

Issue #998, opened 2023-10-18
[feature request] Auto update Cemu's GNU/Linux Appimage


davidhedlund:
It would be nice if the Appimage in GNU/Linux could auto-update itself, like the PlayStation 3 emulator [RPCS3](https://github.com/RPCS3/rpcs3).

Auto update only exists for Windows at this writing.

* https://github.com/cemu-project/Cemu/issues/952

PR #1145, opened 2024-03-31
Linux/Mac Auto-Updater

qurious-pixel:
Linux / Mac Auto-updater

- Mac build ~~WIP~~ (ready for testing)
- ~~Linux still requires xz~~

fixes https://github.com/cemu-project/Cemu/issues/998


Exzap:
In the current state this would try to update the ubuntu and flatpak builds too using the appimage code path?

qurious-pixel:
I tested with the Ubuntu update path as well, which would be similar to Windows, but put that idea to the side because of permissions. Have to examine if the user would have permission to write to `/usr/bin` to replace the installed app. 

AppImage does not require elevated permissions in standard paths.

Mac code (which hasn't been updated in the PR yet) has to handle Apple app sandbox.

Exzap:
Looks good to me. Sorry for the delay

PR #1154, opened 2024-04-05
ELF loader: allocate PPC-recompiler jump table for executable sections

goeiecool9999:
This fixes crashing when homebrew applications are started from an ELF file with the recompiler enabled.


aboood40091:
Very cool addition, but also very little homebrew use ELF nowadays.

PR #1157, opened 2024-04-06
Avoid current app from showing up as "???" for others in Friend List + View friends' status

47463915:
iosu_fpd.cpp, nexFriends.cpp and nexFriends.h are edited to have these changes:
- Starting a friend session on a non-game application (system app, etc) will force myPresence.gameKey.titleId and myPresence.gameKey.ukn to be 0 in order to avoid the app from showing up as "???" for other people in Friend List and instead have no icon just like in real hardware
- Updating the gamemode will force myPresence.gameKey.titleId and myPresence.gameKey.ukn to be the current foreground title id/version (or 0 if not a game app) in order to avoid the app from showing up as "???" for other people in Friend List (the titleid is incorrect when the gamemode is updated, not sure why)
- Renamed nexFriend->gameModeMessage to nexFriend->comment since it is the status comment of an account, nothing to do with GameMode
- When calling NexFriendToFPDFriendData, the gamemode description and status comment from NexFriend will be copied over to FriendData, allowing friends comments and gamemode description to be read in Friend List
- Somewhat implemented GetMyPlayingGame (might be not too accurate)

### DIFF
--- a/src/Cafe/IOSU/legacy/iosu_fpd.cpp
+++ b/src/Cafe/IOSU/legacy/iosu_fpd.cpp
@@ -217,9 +217,8 @@ namespace iosu
 			auto fixed_presence_msg = '\0' + frd->presence.msg; // avoid first character of comment from being cut off
 			friendData->friendExtraData.gameModeDescription.assignFromUTF8(fixed_presence_msg);
 			
-			auto comment_utf16 = StringHelpers::FromUtf8(frd->comment.commentString);
-			comment_utf16.insert(0, 1, '\0'); // avoid first character of comment from being cut off
-			memcpy(friendData->friendExtraData.comment, comment_utf16.c_str(), 36);
+			auto fixed_comment = '\0' + frd->comment.commentString; // avoid first character of comment from being cut off
+			friendData->friendExtraData.comment.assignFromUTF8(fixed_comment);
 			
 			// set valid dates
 			friendData->uknDate.year = 2018;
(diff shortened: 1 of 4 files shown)


47463915:
I forgot to change the other instance of gameModeMessage in nexFriend.readData causing the build to fail, sorry about that!

src/Cafe/IOSU/legacy/iosu_fpd.cpp
```diff
@@ -214,6 +214,13 @@ namespace iosu
 			friendData->friendExtraData.gameKey.ukn08 = frd->presence.gameKey.ukn;
 			NexPresenceToGameMode(&frd->presence, &friendData->friendExtraData.gameMode);
 
+			auto fixed_presence_msg = '\0' + frd->presence.msg; // avoid first character of comment from being cut off
+			friendData->friendExtraData.gameModeDescription.assignFromUTF8(fixed_presence_msg);
+			
+			auto comment_utf16 = StringHelpers::FromUtf8(frd->comment.commentString);
+			comment_utf16.insert(0, 1, '\0'); // avoid first character of comment from being cut off
+			memcpy(friendData->friendExtraData.comment, comment_utf16.c_str(), 36);
```
Exzap, line 222: We have `friendExtraData.comment` defined as a char array, but here you are assigning it the bytes from a wide string. If this is working then we should redefine `friendExtraData.comment` as the proper string type. Consider using the helper class `CafeWideString` since that avoids having to manually memcpy

Exzap:
thanks!

PR #1159, opened 2024-04-08
nn_sl: Stub GetDefaultWhiteListAccessor__Q2_2nn2slFv

Maschell:
Stub GetDefaultWhiteListAccessor__Q2_2nn2slFv to avoid crash in Wii U Menu when an online account is used


src/Cafe/OS/libs/nn_sl/nn_sl.cpp
```diff
@@ -0,0 +1,121 @@
...
+#include "Cafe/OS/libs/coreinit/coreinit_MEM.h"
+#include "config/ActiveSettings.h"
+#include "Cafe/CafeSystem.h"
+
+namespace nn
+{
+
+#ifdef __GNUC__
+#define PACK(__Declaration__) __Declaration__ __attribute__((__packed__))
+#endif
```
Maschell, line 12: Is something like this globally defined? I couldn't find anything

PR #1173, opened 2024-04-12
nn_fp: Implement GetMyComment and UpdateCommentAsync

squelchiee:
- When starting a friend session, myComment (nexComment) will be stored in the session (NexFriends) variable
- Using GetMyComment will then get session->myComment and convert the commentString to UTF16
- Implemented nexComment.writeData
- UpdateCommentAsync and UpdatePreferenceAsync are incredibly similiar to each other, so most of the code is copied over from that in order to make it work

### DIFF
--- a/src/Cafe/IOSU/legacy/iosu_fpd.cpp
+++ b/src/Cafe/IOSU/legacy/iosu_fpd.cpp
@@ -733,7 +733,7 @@ namespace iosu
 				g_fpd.nexFriendSession->getMyComment(myComment);
 				auto comment_utf16 = StringHelpers::FromUtf8(myComment.commentString);
 				comment_utf16.insert(0, 1, '\0'); // avoid first character of comment from being cut off
-				memcpy(vecOut->basePhys.GetPtr(), comment_utf16.c_str(), MY_COMMENT_LENGTH*sizeof(uint16be));
+				memcpy(vecOut->basePhys.GetPtr(), comment_utf16.c_str(), MY_COMMENT_LENGTH * sizeof(uint16be));
 				return 0;
 			}
 
@@ -1147,7 +1147,6 @@ namespace iosu
 				return FPResult_Ok;
 			}
 
-
 			nnResult CallHandler_UpdateCommentAsync(FPDClient* fpdClient, IPCIoctlVector* vecIn, uint32 numVecIn, IPCIoctlVector* vecOut, uint32 numVecOut)
 			{
 				std::unique_lock _l(g_fpd.mtxFriendSession);
@@ -1156,12 +1155,12 @@ namespace iosu
 				if (!g_fpd.nexFriendSession)
 					return FPResult_RequestFailed;
 				uint32 messageLength = vecIn[0].size / sizeof(uint16be);
-				if (messageLength == 0 || (vecIn[0].size % sizeof(uint16be)) != 0)
+				DeclareInputPtr(newComment, uint16be, messageLength, 0);
+				if (messageLength == 0 || newComment[messageLength-1] != 0)
 				{
 					cemuLog_log(LogType::Force, "UpdateCommentAsync: Message must contain at least a null-termination character");
 					return FPResult_InvalidIPCParam;
 				}
-				DeclareInputPtr(newComment, uint16be, messageLength, 0);
 				IPCCommandBody* cmd = ServiceCallDelayCurrentResponse();
 
 				auto utf8_comment = StringHelpers::ToUtf8(newComment, messageLength);
--- a/src/Cafe/IOSU/legacy/iosu_fpd.cpp
+++ b/src/Cafe/IOSU/legacy/iosu_fpd.cpp
@@ -1147,6 +1147,7 @@ namespace iosu
 				return FPResult_Ok;
 			}
 
+
 			nnResult CallHandler_UpdateCommentAsync(FPDClient* fpdClient, IPCIoctlVector* vecIn, uint32 numVecIn, IPCIoctlVector* vecOut, uint32 numVecOut)
 			{
 				std::unique_lock _l(g_fpd.mtxFriendSession);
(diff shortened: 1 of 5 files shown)


47463915:
I'll have to convert this as a draft for now, still seems to have a few issues and I opened this PR a bit too soon.

47463915:
Something I have noticed, this does not work too well without nn_fp.rpl placed in cafeLibs/, GetMyComment doesn't get called (meaning you can't read your own comment) and UpdateCommentAsync seems to corrupt myPresence->GameKey on the client (comment still updates for other friends and the title remains untouched serverside), I'm not exactly sure on what to do about this.

src/Cafe/OS/libs/nn_fp/nn_fp.cpp
```diff
@@ -607,6 +607,14 @@ namespace nn
 			return resultBuf != 0 ? 1 : 0;
 		}
 
+		nnResult UpdateCommentAsync(char16_t* newComment, void* funcPtr, void* customParam)
+		{
+			FP_API_BASE();
+			auto ipcCtx = std::make_unique<FPIpcContext>(iosu::fpd::FPD_REQUEST_ID::UpdateCommentAsync);
+			ipcCtx->AddInput(newComment, 36);
```
Exzap, line 614: `newComment` is a null-terminated big-endian wide string and our current convention for those is to define them as `uint16be*`
But also I took a quick look into nn_fp.rpl and it behaves differently here. The length passed to `AddInput` is dynamically calculated. I think this is how its implemented:
```
uint32 commentLen = CafeStringHelpers::Length(newComment, 17);
if(commentLen>=17)
    return FPResult_InvalidIPCParam;
ipcCtx->AddInput(newComment, sizeof(uint16be) * (commentLen + 1));    
```
So it seems to work similar to how `UpdateGameModeWithUnusedParam` passes it's string.

src/Cafe/IOSU/legacy/iosu_fpd.cpp
```diff
@@ -1143,6 +1147,31 @@ namespace iosu
...
 			}
 
+			nnResult CallHandler_UpdateCommentAsync(FPDClient* fpdClient, IPCIoctlVector* vecIn, uint32 numVecIn, IPCIoctlVector* vecOut, uint32 numVecOut)
+			{
+				std::unique_lock _l(g_fpd.mtxFriendSession);
+				if (numVecIn != 1 || numVecOut != 0)
+					return FPResult_InvalidIPCParam;
+				if (!g_fpd.nexFriendSession)
+					return FPResult_RequestFailed;
+				DeclareInputPtr(newComment, char16_t, (vecOut[0].size / 2), 0);
```
Exzap, line 1157: `char16_t` should be `uint16be`
The way this reads the string from the buffer doesn't match IOSU behavior. Take a look at `CallHandler_UpdateGameMode` for the general pattern.

src/Cafe/IOSU/legacy/iosu_fpd.cpp
```diff
@@ -1143,6 +1147,31 @@ namespace iosu
...
+				IPCCommandBody* cmd = ServiceCallDelayCurrentResponse();
+
+				auto utf8_comment = StringHelpers::ToUtf8((const uint16be*)newComment, vecIn[0].size);
```
Exzap, line 1160: You are passing the size in bytes to `ToUtf8`, but it wants the size in wide characters.

Exzap (changes requested):

Exzap:
There seem to be some out of bounds reads but no bad writes (unless MY_COMMENT_LENGTH is wrong or we are using it in the wrong places?) So not sure what could be causing the corruption. The call to `GetMyComment` probably depends on some other state from the friend lib that we provide incorrectly/differently.

47463915:
> The call to `GetMyComment` probably depends on some other state from the friend lib that we provide incorrectly/differently.

Is there a clear indicator to find where that state is?

47463915:
Another note, the `GameKey` corruption seems to come from `UpdatePresenceAsync` and not from `UpdateCommentAsync`

squelchiee:
I know it has been a really long time since I last pushed a commit, I'd say the PR is almost finished, the only problem I have seen is that the comment doesn't update for the client in Friend List (it still updates for every other friend and if you restart the app the comment does change), I have a feeling this has something to do with the graphics and not the nn_fp library itself, I even tried patching out some friend list functions on real hardware to see where the text gets updated but I wasn't able to find anything.
Also the latest commit changed multiple lines of `CallHandler_GetMyComment`, but for some reason Github is only showing that I removed a comment, the Files Changed tab does show the actual changes I made though.

src/Cafe/OS/libs/nn_fp/nn_fp.cpp
squelchiee: I checked nn_fp.rpl and yeah, that implementation is accurate, though commentLen is added by 2 instead of 1

src/Cafe/IOSU/legacy/iosu_fpd.cpp
squelchiee: I changed the code using `CallHandler_UpdateGameMode` as a reference, and it works fine

src/Cafe/OS/libs/nn_fp/nn_fp.cpp
```diff
@@ -472,6 +480,14 @@ namespace nn
 			return ipcCtx->Submit(std::move(ipcCtx));
 		}
 
+		nnResult GetMyComment(iosu::fpd::FPDPreference* myPreference)
+		{
+			FP_API_BASE();
+			auto ipcCtx = std::make_unique<FPIpcContext>(iosu::fpd::FPD_REQUEST_ID::GetMyComment);
+			ipcCtx->AddOutput(myPreference, sizeof(uint16be)*0x12);
```
Exzap, line 487: Can you use `MY_COMMENT_LENGTH` here instead of hardcoding 0x12. The definition of `MY_COMMENT_LENGTH` would have to be moved from `CallHandler_GetMyComment` into iosu_fpd.h

src/Cafe/OS/libs/nn_fp/nn_fp.cpp
```diff
@@ -607,6 +623,20 @@ namespace nn
 			return resultBuf != 0 ? 1 : 0;
 		}
 
+		nnResult UpdateCommentAsync(uint16be* newComment, void* funcPtr, void* customParam)
+		{
+			FP_API_BASE();
+			auto ipcCtx = std::make_unique<FPIpcContext>(iosu::fpd::FPD_REQUEST_ID::UpdateCommentAsync);
+			uint32 commentLen = CafeStringHelpers::Length(newComment, 17);
+			if (commentLen >= 17)
```
Exzap, line 631: Replacing 17 with `MY_COMMENT_LENGTH+1` makes it clearer where the size comes from

Exzap: Actually nvm, MY_COMMENT_LENGTH is 18 and not 16. So it's using a different size as the base

Exzap:
Is this ready to be merged or is more work being done? Asking since it's still marked as draft.

PR #1175, opened 2024-04-14
CI: Don't strip AppImage symbols

goeiecool9999:
As it turns out fixing the symbols in the debug output for AppImage builds was very simple.
Linux deploy strips binaries by default. They helpfully provide the ``NO_STRIP`` environment variable to not strip...
https://github.com/linuxdeploy/linuxdeploy/issues/72
``NO_STRIP`` not set:
image
``NO_STRIP`` set to 1:
image


Exzap:
This is great. Thanks!

PR #1188, opened 2024-04-28
nsysnet: Some improvements

GaryOderNichts:
This PR changes the following:
- Adds missing WU sockopt types
- Avoids always raising an assert for unimplemented types
- Properly handle `NULL` `src_addr` and `addrlen` in `recvfrom`
- Implements `getsockopt` with `SO_TYPE` to query the socket type


Exzap:
Looks good to me, thanks!

Issue #962, opened 2023-09-06
Homebrew Format Support


TurtleP:
Hi,

I'm the developer of [LÖVE Potion](https://github.com/lovebrew/lovepotion). I didn't see any issues open (or closed, unless the GitHub app isn't showing them) about this.

Essentially, I prefer to test a lot of my features and bug fixes on official hardware. However, Wii U is a bit of a pain point for me because there's a few work arounds and delays (which are semi-minor) compared to 3DS and Switch development.

Anyways, `wut` has added a new format, `wuhb` which allows homebrew to be bundled into a single application. This works wonderfully on hardware via Aroma cfw, but if I wanted to test on Cemu I cannot since it doesn't load the format.

One of the original ideas was for me to have the same tree structure on the virtual sd card (e.g. code, content, etc), but as of right now it doesn't even load. It used to work when I was doing initial work on the Wii U version. That's a separate issue I need to debug.

Regardless, supporting the `wuhb` format would be very convenient for homebrew developers since everything is bundled into one binary. I understand Cemu is still experimental, but as mentioned before, I can easily run my 3DS or Switch homebrew on their emulators if I wanted.

aboood40091:
(I do see your point in wanting `wuhb` support, but the 3DS and Switch arguments are irrelevant.)

PR #1190, opened 2024-04-30
add support for WUHB file format

goeiecool9999:
Closes: #962
This work is heavily inspired by the existing code to support WUA.
Some parts are copied and modified from WUT's wuhbtool (structures and hash calculation function)

Questions left to be answered:
- is std::filesystem::path safe to use across platforms in this way? https://github.com/cemu-project/Cemu/blob/37cd32d33477b36e9d27d1dcfb47294a0873a4a2/src/Cafe/Filesystem/WUHB/WUHBReader.cpp#L130
- ~~Should fil/dir table entries be cached?~~ A: It's probably fine.
- ~~How to assign a title ID to homebrew apps? Some kind of hash function?~~ A: CRC32(author+longName+shortName)


Crementif:
Oh, this is awesome to have. Nice work!

Maschell:
> How to assign a title ID to homebrew apps? Some kind of hash function?

On console for the "homebrew_on_menu" plugin I'm using a hash of the relative path to the .wuhb (e.g. "wiiu/apps/test.wuhb") with this hash function to get the lower title id and 0005000F for the first 4bytes : https://github.com/wiiu-env/homebrew_on_menu_plugin/blob/d0581407d69575d5205ba40ee188867e6c0dcdc1/src/utils/utils.cpp#L7

e.g. 0005000F-[HASH]

I think CEMU could/should come up with it's own implementation for the "lower tid", but they should also start with0005000F for consistency?

Maschell:
Not sure to whiche extend Cemu is even handling the cos.xml, but .wuhb under Aroma are running with these cos.xml values:

https://github.com/wiiu-env/MochaPayload/blob/6f47f2f9bb29b4bf649bf52f60ceb83876c8a45f/source/ios_mcp/source/mcp_loadfile.c#L476-L498

src/util/helpers/helpers.cpp
```diff
@@ -437,3 +439,40 @@ std::string GenerateRandomString(const size_t length, const std::string_view cha
...
+	stream.opaque = Z_NULL;
+	stream.avail_in = compressed.size();
+	stream.next_in = (Bytef*)compressed.data();
+	err = inflateInit2(&stream, 32); // 32 is a zlib magic value to enable header detection
+	if (err != Z_OK)
+		return {};
+
+	do
+	{
+		decompressed.resize(decompressed.size() + 1024);
```
Exzap, line 460: Increasing by a fixed 1024 bytes at a time seems rather inefficient? For a 1MB file thats a thousand vector resize operations.
I recommend starting out with a smallish size like 32KB (maybe even as a size hint parameter?) and then just doubling the size each loop iteration.

goeiecool9999: See latest commit. I know defining a variable and assigning the argument is redundant but I think it helps readability and it should be trivial for a compiler to optimize. I set the hint in the game list for the icons to 70KiB since one extracted game icon TGA I looked at is slightly more than 64KiB.

Exzap:
Looks good. Thanks!

Issue #686, opened 2023-03-06
Pokemon Rumble U Black Screen


TylerJaacks:
Cemu Version: 2.0-28
Specs: RTX 2080 Ti | 5950X | 64GB | Windows 11 22H2

When loading into a new level in Pokemon Rumble [TitleId: 00050000-10147700] [v0] there is a Black Screen this happens in both Vulkan and OpenGL. This game has worked in older versions of Cemu around 2 to 3 years ago.

![Screenshot 2023-03-05 222705](https://user-images.githubusercontent.com/1505374/223019490-20fdc479-f4c1-4e64-9565-93949b2397d0.png)

Squall-Leonhart:
if it works in -27, the issue is already known.

Squall-Leonhart:
https://github.com/cemu-project/Cemu/releases/tag/v2.0-27

user18081972:
> if it works in -27, the issue is already known.
The same issue occurs in v2.0-27, and also v1.26.2f
![image](https://user-images.githubusercontent.com/124319511/224491933-0c0078d9-2325-484d-bac8-8614fcbb657a.png)
![image](https://user-images.githubusercontent.com/124319511/224492158-af2c3ff8-6257-4694-9543-3842d6e92a0e.png)

It has been reported as early  as v1.16.1 https://wiki.cemu.info/wiki/Pok%C3%A9mon_Rumble_U
Here is a video from 2020 with the same issue: https://www.youtube.com/watch?v=JMSMsaw6xAk

Someone in the discord said the following about the bug:
"Arnie - 01/23/2022 1:33 AM
it uses unimplemented functions
or there is a race condition (I can't remember what was said about that game specifically when I asked about it a few months ago) that causes it to crash"

TylerJaacks:
Is this related to Amiibo then?

user18081972:
the game has optional Amiibo integration, so it is likely.

Squall-Leonhart:
amiibo's are fine on cemu.

MonochromeKirby:
> [16:43:59.215] Unsupported lib call: ntag.NTAGIsInit
> [16:43:59.215] Unsupported lib call: ntag.NTAGInit
> [16:43:59.215] Unsupported lib call: ntag.NTAGSetTagDetectCallback
> [16:43:59.215] Unsupported lib call: ntag.NTAGSetFormatSettings

[...]

PR #1193, opened 2024-05-04
Add initial NTAG and NFC implementation

GaryOderNichts:
This PR adds support for NTAG which is used by Pokemon Rumble U to read and write tags.
NTAG is built on top of NFC, which this PR also adds basic support for.

Checklist:
- [x] Properly implement states and callbacks to allow playing Rumble U (fixes https://github.com/cemu-project/Cemu/issues/686)
- [x] Fully support reading from tags
- [x] Support writing to tags (needs testing)
- [ ] Support for formatting tags (When does Rumble U use `NTAGFormat`?)
- [ ] Properly integrate nn_nfp with NTAG and NFC


Exzap:
I see this also implements the /dev/ccr_nfc IOSU interface and driver, very nice work :thumbsup:
There is some redundancy between the added Stream helper and `src/util/helpers/Serializer.h` and `Deserializer.h` but it's not a big deal and everything else looks good to me so lets merge this now. Thanks!

PR #1198, opened 2024-05-10
Update libraries

Xphalnos:
| Libname  | Previous version | New version
| ------------- | ------------- | ------------- |
| zlib  | 1.3  | 1.3.1  |
| zstd  | 1.5.5  | 1.5.6  |
| sdl2  | 2.28.5  | 2.30.0  |
| boost  | 1.83.0  | 1.85.0  |
| libpng  | 1.6.40  | 1.6.43  |
| glm  | 0.9.9.8  | 1.0.1  |
| glslang  | 13.1.1  | 14.2.0  |
| openssl  | 3.2.0  | 3.3.0  |
| curl  | 8.5.0  | 8.7.1  |
| libusb  | 1.0.26.11791  | 1.0.27  |

I would like to point out that I made this update because boost 1.85.0 was released and also because I updated glslang with VCPKG. Link : https://github.com/microsoft/vcpkg/pull/38638

I tested compiling it on Visual Studio Community and it compiles fine.

### DIFF
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -1,7 +1,7 @@
 {
   "name": "cemu",
   "version-string": "1.0",
-  "builtin-baseline": "53bef8994c541b6561884a8395ea35715ece75db",
+  "builtin-baseline": "cbf4a6641528cee6f172328984576f51698de726",
   "dependencies": [
     "pugixml",
     "zlib",


Exzap:
Normally we handle dependency updates internally because it's easy to overlook issues. Compiling alone isn't really indicative of everything still working. But since we planned to update vcpkg anyway we can merge this. The CI fails right now because it has vcpkg commits hardcoded, but I'll fix that with a follow-up commit after merging this.

Issue #1200, opened 2024-05-11
Graphic pack parser incorrectly parses commas in strings


splatoon1enjoyer:
### Current Behavior

In graphic pack patches, when defining a string:

`.string "Foo ,"`

Cemu's parser fails to accept commas as part of the string, which causes this error at runtime:

`[20:09:22.680] Line X: Error in assembler: String constants must end with a quotation mark. Example: "text"`

This is a problem when writing patches for games such as Splatoon, which have enums hard coded as string literals that get parsed at runtime, requiring the user to use commas as part of the string.

### Expected Behavior

The string should be sanitized properly and consider the comma or any other reasonable character as part of the string.

### Steps to Reproduce

Define a string with a comma.

goeiecool9999:
I've looked at the code.
From the perspective of cemu's PPC assember it looks like it's a data directive with two arguments: ``"Foo`` and ``"``. ```"foo``` doesn't end in a ``"`` so it treats it as invalid.
https://github.com/cemu-project/Cemu/blob/cf41c3b136ab7272e6801991d081c9d2c69c7143/src/Cemu/PPCAssembler/ppcAssembler.cpp#L2419-L2470

PR #1201, opened 2024-05-11
Fix commas edge case in strings when parsing an assembly line

splatoon1enjoyer:
This patch should fix the issue I described in #1200.


goeiecool9999 (approved):
Simple fix. LGTM

Exzap:
thanks!

PR #1209, opened 2024-05-20
OpenGL: fix loading cached shaders from GL cache files

goeiecool9999:
The code that uses glProgramBinary to load compiled shaders from shaderCache/precompiled/\<TitleID\>_gl.bin files is unreachable. The fileSize variable is never changed so the function always fails. This means Cemu has been relying entirely on driver cache implementations for some time. (they must be pretty effective since nobody seems to have noticed this)


Exzap:
Nice catch.

In our experience, `glProgramBinary` is pretty broken/useless across the board on Windows (unsure about Linux drivers). On 2 out of the 3 big vendors we had an issue where shaders would not be restored with all their state permutations. So you'd still get stutter during gameplay occasionally. The driver internal caches do not have this problem but iirc they do get invalidated on driver update. Either way thanks for the fix

PR #1210, opened 2024-05-20
Include cstdarg in precompiled.h

FireBurn:
This fixes the following build error:

src/Common/unix/FileStream_unix.cpp: In member function ‘void FileStream::writeStringFmt(const char*, ...)’: src/Common/unix/FileStream_unix.cpp:184:9: error: ‘va_start’ was not declared in this scope
  184 |         va_start(args, format);
      |         ^~~~~~~~

I'm guessing cstdarg is no longer automatically pulled into the includes with gcc 14.1

### DIFF
--- a/src/Common/precompiled.h
+++ b/src/Common/precompiled.h
@@ -31,6 +31,7 @@
 #endif
 
 // c includes
+#include <cstdarg>
 #include <cstdint>
 #include <cstdlib>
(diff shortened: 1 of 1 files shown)


Exzap:
#1193 already includes a fix for this and will be merged soon. The effort is appreciated regardless, thanks!

Issue #1214, opened 2024-05-22
cemu2.0-83 crashes on opening on Mac -"Namespace DYLD, Code 1 Library missing"


wataru2024:
### Current Behavior

When using earlier versions, there is a black screen when opening the game. Therefore, I downloaded the latest version and encountered more serious issues, even unable to open CEMU.
Used versions: 2.0-25, 2.0-80
When using 2.0-25, manually install Vulkan SDK to link to CEMU and open the game with a black screen
2.0-80, black screen when opening game
2.0-83, unable to open cemu
I'm curious if it's possible to use only CEMU for M3 Pro and 14.5MAC OS. Can you please advise me on how to do it.
The following is an error report:
-------------------------------------
Translated Report (Full Report Below)
-------------------------------------

Process:               Cemu [31047]
Path:                  /Applications/Cemu.app/Contents/MacOS/Cemu
Identifier:            info.cemu.Cemu
Version:               2.0.0 (2.0.0)
Code Type:             X86-64 (Translated)
Parent Process:        launchd [1]
User ID:               501

Date/Time:             2024-05-22 13:45:02.9930 +0800
OS Version:            macOS 14.5 (23F79)
Report Version:        12
Anonymous UUID:        1447B8A2-5BBB-FC50-744C-D50549656E5D

Sleep/Wake UUID:       812A57E3-0F65-40F2-ADF8-81A458760044

Time Awake Since Boot: 9400 seconds
Time Since Wake:       913 seconds

System Integrity Protection: enabled

Notes:
PC register does not match crashing frame (0x0 vs 0x105506A78)

Crashed Thread:        0

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000

Termination Reason:    Namespace DYLD, Code 1 Library missing
Library not loaded: /Users/*/libusb-1.0.0.dylib
Referenced from: <B91BA442-DEFA-3E91-B581-5C048E644BBA> /Applications/Cemu.app/Contents/MacOS/Cemu
Reason: tried: '/Users/*/libusb-1.0.0.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/Users/*/libusb-1.0.0.dylib' (no such file), '/Users/*/libusb-1.0.0.dylib' (no such file), '/usr/local/lib/libusb-1.0.0.dylib' (no such file), '/usr/lib/libusb-1.0.0.dylib' (no such file, not in dyld cache)
[...]

PR #1219, opened 2024-05-24
set VCPKG path to libusb for MacOS

qurious-pixel:
Sets the absolute path to the library and changes the binary to search the Frameworks path of the bundle.

Fixes https://github.com/cemu-project/Cemu/issues/1214


Exzap:
thanks!

PR #1224, opened 2024-05-27
Fix GamePad window size

ColinKinloch:
The gamepad window was the wrong size.

I don't know wxWidgets well enough to say why this fix works or whether it will cause issues.

| before | after |
|-|-|
| Screenshot from 2024-05-27 11-34-59 | Screenshot from 2024-05-27 11-33-21 |


zhangbo8418:
The problem is that you can drag and adjust the size of this window with the mouse, but your modification prevents it from being smaller.

ColinKinloch:
I'll take a closer look.
I think the problem is that the window is created at the correct size not accounting for decorations.

This commit fixes that by forcing a resize once the canvas is created.
It would be better to the the window to map at the correct size initially and not automatically resize when the canvas is ready.

PR #1234, opened 2024-06-14
Added total downloads

Xphalnos:
I added the total number of downloads.
I was impressed because Cemu reached 4.1 million downloads !

Preview:
Capture d'écran 2024-06-15 082618


Exzap:
Maybe it's just me, but I always felt that presenting download counts in the readme is a form of boasting. Especially since the added counter is inflated because it counts individual users many times due to the different update releases.

I appreciate the contribution but overall I rather not have downloads tracked on the readme.

PR #1243, opened 2024-06-26
nsyshid: Libusb SetProtocol and SetReport

deReeperJosh:
Round 2 of me attempting to implement the SetProtocol and SetReport methods for libusb. Will need user testing on linux for the Skylanders games, and some testing to confirm I haven't broken anything for Lego Dimensions also.

### DIFF
--- a/src/Cafe/OS/libs/nsyshid/BackendLibusb.cpp
+++ b/src/Cafe/OS/libs/nsyshid/BackendLibusb.cpp
@@ -230,6 +230,17 @@ namespace nsyshid::backend::libusb
 		return nullptr;
 	}
 
+	std::pair<int, ConfigDescriptor> MakeConfigDescriptor(libusb_device* device, uint8 config_num)
+	{
+		libusb_config_descriptor* descriptor = nullptr;
+		const int ret = libusb_get_config_descriptor(device, config_num, &descriptor);
+		if (ret == LIBUSB_SUCCESS)
+			return {ret, ConfigDescriptor{descriptor, libusb_free_config_descriptor}};
+
+		return {ret, ConfigDescriptor{nullptr, [](auto) {
+									  }}};
+	}
+
 	std::shared_ptr<Device> BackendLibusb::CheckAndCreateDevice(libusb_device* dev)
 	{
 		struct libusb_device_descriptor desc;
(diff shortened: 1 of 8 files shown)


deReeperJosh:
@ssievert42 your review would be much appreciated here :)

ssievert42:
@deReeperJosh I still don't own any Skylanders games, but I'll try to take a look at this when I can :)

src/Cafe/OS/libs/nsyshid/BackendLibusb.h
```diff
@@ -3,7 +3,7 @@
 
 #include "nsyshid.h"
 
-#if NSYSHID_ENABLE_BACKEND_LIBUSB
+#if 1
```
ssievert42, line 6: If this is to use the libusb backend on Windows by default, a cleaner way would probably be setting the following to `OFF` in `CMakeLists.txt`:
https://github.com/cemu-project/Cemu/blob/64b0b85ed5fe15d17cfa6b5fce0044000b013a07/CMakeLists.txt#L103
Otherwise we'd see each USB device twice!
I honestly have no idea what a game would do in that case; from connecting two Lego Dimensions portals at the same time, my guess would be that one is just chosen at random.

src/Cafe/OS/libs/nsyshid/BackendLibusb.h
```diff
@@ -44,6 +44,11 @@ namespace nsyshid::backend::libusb
 										bool& endpointOutFound, uint8& endpointOut, uint16& endpointOutMaxPacketSize);
 	};
 
+	template<typename T>
+	using UniquePtr = std::unique_ptr<T, void (*)(T*)>;
+
+	using ConfigDescriptor = UniquePtr<libusb_config_descriptor>;
```
ssievert42, line 50: That's such a neat way to deal with C memory management :)

src/Cafe/OS/libs/nsyshid/BackendLibusb.cpp
```diff
@@ -241,14 +252,31 @@ namespace nsyshid::backend::libusb
...
+		std::vector<ConfigDescriptor> config_descriptors{};
+		for (uint8 i = 0; i < desc.bNumConfigurations; ++i)
+		{
+			auto [ret, config_descriptor] = MakeConfigDescriptor(dev, i);
+			if (ret != LIBUSB_SUCCESS || !config_descriptor)
+			{
+				cemuLog_log(LogType::Force, "Failed to make config descriptor {} for {:04x}:{:04x}: {}",
+							i, desc.idVendor, desc.idProduct, libusb_error_name(ret));
+			}
+			config_descriptors.emplace_back(std::move(config_descriptor));
```
ssievert42, line 264: This puts invalid config descriptors in the `config_descriptors` vector.
I guess this might cause problems later on.

src/Cafe/OS/libs/nsyshid/BackendLibusb.cpp
```diff
@@ -680,32 +698,84 @@ namespace nsyshid::backend::libusb
...
 	}
 
-	bool DeviceLibusb::SetProtocol(uint32 ifIndex, uint32 protocol)
+	template<typename Configs, typename Function>
+	static int DoForEachInterface(const Configs& configs, uint8 config_num, Function action)
+	{
+		int ret = LIBUSB_ERROR_NOT_FOUND;
+		if (configs.size() <= config_num || !configs[config_num])
+			return ret;
+		for (uint8 i = 0; i < configs[config_num]->bNumInterfaces; ++i)
```
ssievert42, line 707: See my other comment about invalid config descriptors above;
This looks like the place where having invalid config descriptors, that contain just a `nullptr`, will blow up.

ssievert42 (changes requested):
Cool that you got it to work!
Just some small nitpicks, otherwise LGTM :+1:
I'm currently doing a test build, and will report back once I've tested if this breaks anything for Lego Dimensions.

ssievert42:
Lego Dimensions on Linux still works :+1:

src/Cafe/OS/libs/nsyshid/BackendLibusb.h
deReeperJosh: Ah this was just me because my IDE wasn't recognising the libusb backend enabled and all the code was very dim, will add this back in

src/Cafe/OS/libs/nsyshid/BackendLibusb.h
deReeperJosh: Wish I could claim this code as my own but this os borrowed from another libusb implementation

PR #1244, opened 2024-06-28
nsyshid: Skylander Tidyups

deReeperJosh:
Just some tidyups for the Skylander code based on some reviews from initial PR, as well as setting the filestream to 0 before writing in the Save() method

### DIFF
--- a/src/gui/EmulatedUSBDevices/EmulatedUSBDeviceFrame.cpp
+++ b/src/gui/EmulatedUSBDevices/EmulatedUSBDeviceFrame.cpp
@@ -1,7 +1,6 @@
 #include "gui/EmulatedUSBDevices/EmulatedUSBDeviceFrame.h"
 
 #include <algorithm>
-#include <random>
 
 #include "config/CemuConfig.h"
 #include "gui/helpers/wxHelpers.h"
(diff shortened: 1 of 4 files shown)


src/gui/EmulatedUSBDevices/EmulatedUSBDeviceFrame.cpp
```diff
@@ -251,46 +240,14 @@ CreateSkylanderDialog::CreateSkylanderDialog(wxWindow* parent, uint8 slot)
...
-		wxFileOutputStream output_stream(saveFileDialog.GetPath());
-		if (!output_stream.IsOk())
+		
+		if(!nsyshid::g_skyportal.CreateSkylander(_utf8ToPath(m_filePath.utf8_string()), skyId, skyVar))
 		{
-			wxMessageDialog saveError(this, "Error Creating Skylander File");
-			return;
+			wxMessageDialog errorMessage(this, "Failed to create file");
+			errorMessage.ShowModal();
+			this->EndModal(0);
```
Exzap, line 248: missing `return` after this? I dont think its intentional to call EndModal twice

deReeperJosh: added this back in

PR #1252, opened 2024-07-07
Windows default to non-portable + Reworked MLC handling and related UI

Exzap:
The Windows version of Cemu is currently the only version that is portable by default. While personally I am a fan of applications that are portable, this has been a constant troubleshooting burden with users losing their files due to deleting the Cemu directory or moving stuff around. It's especially bad with the mlc01 location which isn't really an intuitive concept and is easily "misplaced".

Since directory and MLC handling are intertwined with the UI I ended up overhauling parts of that as well. In summary the changes are: 
- The Windows version is now non-portable by default (files stored to `%appdata%/Roaming/Cemu`). Similar to Linux/MacOS portable mode is still supported by creating a directory `portable` next to the executable. If `settings.xml` is detected in the same directory as Cemu.exe, it will operate in the previous portable mode to avoid breaking existing installations
- UI is no longer encouraging the user to set a custom MLC, but the option is still there
- Getting Started dialog and General Settings UI have been reworked to align with these changes
- Some more UI tweaks here and there. Added extra description texts and message boxes to clarify things

I would appreciate if people could test this build and let me know if there are any issues with their existing installs regardless of what OS is being used


Exzap:
Some screenshots of UI. 
Getting Started before:
image
After:
image

General Settings before:
image
After:
image

Fs00:
When I first tried portable mode, I encountered a bug that caused UI translations not to be loaded because Cemu was looking for the _resources_ folder inside the portable folder, instead of near the executable. I can't tell by looking at the code if that's fixed now, but it's worth checking before merging.

Also, IMO it would be nice to make portable mode more discoverable. A way to do that could be to have a message along the lines of _"Cemu will save your data to [path]. If you wish to have them stored close to Cemu executable, create a folder named 'portable' in that location"_ in place of _"Cemu is running in portable mode"_ when no portable folder is found.

Exzap:
`resources` lookup is still broken under some conditions, thanks for letting me know. I'll look into fixing it before merging

exverge-0 (changes requested):

PR #1255, opened 2024-07-11
macOS: Apple Silicon (interpreter only)

exverge-0:
Fixes(?) #396, supersedes #1211 (I don't own BOTW and haven't been able to reproduce the issue on other games, but this fixes a similar error that only occurred on Apple Silicon)

The re-compiler hasn't been implemented and only the interpreter works. (and there's already an aarch64 recompiler WIP, assuming that they plan on merging into this repo)


Exzap:
I'll be honest, it will probably be a while until we have a decent ARM recompiler that will outperform Rosetta so this seems a bit early. But I generally don't want to stand in the way of adding new target platforms. That said, the changes made here are detrimental for other platforms. Just on a quick glance some obvious issues:
- Putting gx2WriteGatherPipe access behind a mutex is absolutely gonna wreck performance. Better to work with atomic pointers or manually insert memory barriers where needed.
- `mmuRange_HIGHMEM` is supposed to match Wii U's memory map. I get that there is a 16KB page size restriction but other platforms don't have this so it should be conditional.
- Lots of changes in the h264 decoder, would be nice if you could give an explanation?

exverge-0:
> Lots of changes in the h264 decoder, would be nice if you could give an explanation?

On arm64, C functions are compiled with a leading underscore and the header files define the ASM functions as a normal C function, resulting in functions 
Admittedly now I see this could probably be done by just wrapping it in an `extern` block, I'll do that instead
> Putting gx2WriteGatherPipe access behind a mutex is absolutely gonna wreck performance. Better to work with atomic pointers or manually insert memory barriers where needed.

Makes sense, I'll revert that and try something else
> `mmuRange_HIGHMEM` is supposed to match Wii U's memory map. I get that there is a 16KB page size restriction but other platforms don't have this so it should be conditional.

Ah, I wasn't sure but that's what I somewhat assumed.

src/util/MemMapper/MemMapperUnix.cpp
```diff
@@ -42,6 +42,12 @@ namespace MemMapper
 
 	void* AllocateMemory(void* baseAddr, size_t size, PAGE_PERMISSION permissionFlags, bool fromReservation)
 	{
+		uint64 page_size = sysconf(_SC_PAGESIZE);
+		if ((uint64) baseAddr % page_size != 0)
+		{
+			baseAddr = (void*) ((uint64)baseAddr & ~(page_size - 1));
+		}
+
```
Exzap, line 50: Silently allocating a different range than the one requested is probably not the way to go. Wii U page size is actually 128KB, therefore 16KB pages shouldn't be an issue unless we accidentally hardcoded in a 4K requirement somewhere. Sorry for not properly communicating this before.

`mmuRange_HIGHMEM` is a bit special but the previous workaround was fine imho. But let me know if there were any other issues and we can probably find a way to make all systems compatible with 16KB page size.

src/Cafe/HW/Latte/Renderer/Vulkan/VulkanPipelineCompiler.cpp
```diff
@@ -988,7 +988,8 @@ bool PipelineCompiler::Compile(bool forceCompile, bool isRenderThread, bool show
 	pipelineInfo.pDynamicState = &dynamicState;
 	pipelineInfo.pRasterizationState = &rasterizer;
 	pipelineInfo.pMultisampleState = &multisampling;
-	pipelineInfo.pColorBlendState = &colorBlending;
+	if (!_IsVkIntegerFormat(m_renderPassObj->GetColorFormat(0)))
+		pipelineInfo.pColorBlendState = &colorBlending;
```
Exzap, line 992: This should iterate each color attachment and when an integer format is used then disable blending only for that specific attachment. Having one color attachment in integer format does not prevent blending for all the non-integer attachments.

src/util/MemMapper/MemMapperUnix.cpp
exverge-0: > Silently allocating a different range than the one requested is probably not the way to go. 

I didn't mean to put this for allocation, only for the `mprotect` call as the entire page has to have the same permissions, that's my mistake. I'll update it now.

> Wii U page size is actually 128KB, therefore 16KB pages shouldn't be an issue unless we accidentally hardcoded in a 4K requirement somewhere. 

Looking into WiiUBrew's documentation of the memory map, it seems that this memory space isn't allocated on it's own, as the entire area from `0xFFFF0000` to `0xFFFFFFFF`, which includes `mmuRange_HIGHMEM`, appears to be mapped at once.

> `mmuRange_HIGHMEM` is a bit special but the previous workaround was fine imho

I removed the previous workaround in favor of this fix because this should be more flexible for other page sizes other than 4KB and 16KB and wouldn't be Apple-specific

src/Cafe/HW/Latte/Renderer/Vulkan/VulkanPipelineCompiler.cpp
Exzap: I see. I don't want to chalk this down as a bug in MoltenVk yet because it seems to be too big of an oversight to me. I checked [their code](https://github.com/KhronosGroup/MoltenVK/blob/3b9e335fe8fc8a72cf6099acbd54128889445c0a/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm#L1689) and I couldn't really see any logic that would enable blending if we didn't specifically ask for it. But more digging is required to understand this fully.

In the meantime an acceptable temporary workaround would be to disable all blending if ANY of the color attachments are integer format instead of only checking index 0. I also suggest making the workaround apply to MacOS/MoltenVk only.

src/util/MemMapper/MemMapperUnix.cpp
Exzap: > Looking into WiiUBrew's documentation of the memory map, it seems that this memory space isn't allocated on it's own, as the entire area from 0xFFFF0000 to 0xFFFFFFFF, which includes mmuRange_HIGHMEM, appears to be mapped at once.

You looked at the wrong memory map. The relevant PPC userspace one is on [this page](https://wiiubrew.org/wiki/Cafe_OS). But even there it's defined as starting from 0xFFFE0000 which is exactly one PPC page (128KB). This is a bit of a tangent, but the issue is that we don't really support emulating per-thread memory (yet) and the only reason we have defined it in Cemu at all is to prevent Project Zero from crashing since it writes to the very last byte of the address space due to a bug in the game's code. I don't want to define the memory range with the proper full size because that is still incorrect as it's not per-thread. Hence we limited it to as small a size possible so any unsupported access to it would be caught and not go through silently and lead to hard-to-debug crashes later on.

Regardless, I still believe that code using `ReserveMemory`, `AllocateMemory` and `FreeMemory` should be aware of page boundaries and that internally it shouldn't mess with the parameters or allow unaligned base or size. I know the existing code doesn't assert on this right now but ideally it would.

src/gui/MainWindow.cpp
```diff
@@ -2197,7 +2198,7 @@ void MainWindow::RecreateMenu()
 	debugLoggingMenu->AppendSeparator();
 
 	wxMenu* logCosModulesMenu = new wxMenu();
-	logCosModulesMenu->AppendCheckItem(0, _("&Options below are for experts. Leave off if unsure"), wxEmptyString)->Enable(false);
+	logCosModulesMenu->AppendCheckItem(MAINFRAME_MENU_ID_DEBUG_LOGGING_MESSAGE, _("&Options below are for experts. Leave off if unsure"), wxEmptyString)->Enable(false);
```
neebyA, line 2201: Thanks for this, I believe it resolves the debug build issue from #1547

src/config/ActiveSettings.cpp
```diff
@@ -74,10 +74,14 @@ CPUMode ActiveSettings::GetCPUMode()
 
 	if (mode == CPUMode::Auto)
 	{
+#ifdef __aarch64__
+		mode = CPUMode::SinglecoreInterpreter;
+#else
```
neebyA, line 79: I believe this can be removed now

dependencies/ih264d/common/armv8/ih264_intra_pred_luma_8x8_av8.s
```diff
@@ -1030,8 +1030,13 @@ ih264_intra_pred_luma_8x8_mode_horz_u_av8:
     ext       v4.16b, v2.16b , v2.16b , #1
     mov       v5.d[0], v4.d[1]
 
+#if defined(__APPLE__) && defined(__aarch64__)
```
neebyA, line 1033: Nit: I believe the `&& defined(__arch64__)` is unnecessary here, as cmake to begin with will only add these armv8 files to the target sources in the case of `aarch64` or `arm64`:
https://github.com/cemu-project/Cemu/blob/61484598fc21fb127d8499691b72309942ee5e97/dependencies/ih264d/CMakeLists.txt#L149-L153

dependencies/ih264d/common/ih264_deblk_edge_filters.h
```diff
@@ -63,133 +75,143 @@ typedef void ih264_deblk_chroma_edge_bslt4_ft(UWORD8 *pu1_src,
...
                                               const UWORD8 *pu1_cliptab_cr);
 
-typedef void ih264_deblk_chroma_edge_bs4_ft(UWORD8 *pu1_src,
+#if defined(__APPLE__) && defined(__aarch64__)
+#define ih264_deblk_chroma_edge_bslt4_ft(arg) _ih264_deblk_chroma_edge_bslt4_ft arg __asm__(#arg);
+#else
+#define ih264_deblk_chroma_edge_bslt4_ft(arg) _ih264_deblk_chroma_edge_bslt4_ft arg;
+#endif
+
+typedef void _ih264_deblk_chroma_edge_bs4_ft(UWORD8 *pu1_src,
```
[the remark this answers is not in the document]
exverge-0: I hadn't thought of this, and this will probably produce a much cleaner result. Thanks for the suggestion!

Exzap:
@neebyA Try changing the start of the loop to this as a workaround:
```
		for (auto&& [_jumpStart, _jumpInfo] : jumps)
		{
			auto& jumpStart = _jumpStart;
			auto& jumpInfo = _jumpInfo;
```
I am not familiar with macOS development so I am not sure what the best strategy is in regards of which xcode version to target. But in general I am against bumping a compiler version when something can be worked around easily.

exverge-0:
I used @Exzap 's workaround in order to avoid said issue, and it seems to compile fine.

Otherwise, with the addition of the aarch64 recompiler, this PR is essentially done. I haven't done much testing, from what I've tested it seems to run smoothly with no new issues and the only real difference I've noticed being that the arm64 seemingly has less lag spikes, whereas under Rosetta it would stutter while compiling shaders, but otherwise it's seemingly the same. (though Rosetta already ran at 60/30 fps from what I tested, so any performance improvement wouldn't be noticeable)
I also reverted the change from https://github.com/cemu-project/Cemu/issues/1255#discussion_r1709465498 so as to separate it from this PR to not have to include the odd workaround (which hadn't prevented any games from running/caused any visible issues in my testing, only console errors)

dependencies/ih264d/common/armv8/ih264_weighted_bi_pred_av8.s
```diff
@@ -424,17 +432,24 @@ ih264_weighted_bi_pred_chroma_av8:
...
+    ldr       w9, [sp, #88]             //Load ofst1 in w9
+    ldr       w10, [sp, #96]            //Load ofst2 in w10
+    ldr       w11, [sp, #104]           //Load ht in w11
+    ldr       w12, [sp, #112]           //Load wd in w12
+#else
+    ldr       w8, [sp, #80]             //Load wd in w8
+    ldr       w9, [sp, #84]             //Load ht in w9
+    ldr       w10, [sp, #88]            //Load offst2 in w10
+    ldr       w11,  [sp, #92]           //Load offst1 in w11
+    ldr       w12,  [sp, #96]           //Load offst1 in w12
```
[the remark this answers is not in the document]
exverge-0: I made a mistake with the comments which is why they are misleading, however it is still loading the same values into the same registers, the only difference is that on non-Apple, the offsets for each argument on the stack are always 8 bytes as specified by the ABI, whereas Apple seemingly does not follow this convention.

> [Procedure Call Standard for the Arm® 64-bit Architecture (AArch64)](https://github.com/ARM-software/abi-aa/blob/main/aapcs64/aapcs64.rst#parameter-passing)
> If the size of the argument is less than 8 bytes then the size of the argument is set to 8 bytes. The effect is as if the argument was copied to the least significant bits of a 64-bit register and the remaining bits filled with unspecified values.

hauntek:
```
  build-macos:
    runs-on: macos-14
    steps:
    - name: "Checkout repo"
      uses: actions/checkout@v4
      with:
        submodules: "recursive"

    - name: "Select Xcode 16.2"
      run: sudo xcode-select -s /Applications/Xcode_16.2.app/Contents/Developer

    - name: "Verify Xcode Version"
      run: xcodebuild -version
        
    - name: Setup release mode parameters
      run: |
        echo "BUILD_MODE=release" >> $GITHUB_ENV
        echo "BUILD_FLAGS=" >> $GITHUB_ENV
        echo "Build mode is release"

    - name: Setup build flags for version
      if: ${{ inputs.next_version_major != '' }}
      run: |
        echo "[INFO] Version ${{ inputs.next_version_major }}.${{ inputs.next_version_minor }}"
        echo "BUILD_FLAGS=${{ env.BUILD_FLAGS }} -DEMULATOR_VERSION_MAJOR=${{ inputs.next_version_major }} -DEMULATOR_VERSION_MINOR=${{ inputs.next_version_minor }}" >> $GITHUB_ENV
        
    - name: "Install system dependencies"
      run: |
        brew update
        brew install ninja nasm automake libtool

    - name: "Install molten-vk"
      run: |
        curl -L -O https://github.com/KhronosGroup/MoltenVK/releases/download/v1.3.0/MoltenVK-macos.tar
        tar xf MoltenVK-macos.tar
        sudo mkdir -p /usr/local/lib
        sudo cp MoltenVK/MoltenVK/dynamic/dylib/macOS/libMoltenVK.dylib /usr/local/lib

    - name: "Setup cmake"
      uses: jwlawson/actions-setup-cmake@v2
      with:
        cmake-version: '3.29.0'

    - name: "Bootstrap vcpkg"
      run: |
        bash ./dependencies/vcpkg/bootstrap-vcpkg.sh
        
    - name: 'Setup NuGet Credentials for vcpkg'
      shell: 'bash'
      run: |
        mono `./dependencies/vcpkg/vcpkg fetch nuget | tail -n 1` \
        sources add \
        -source "https://nuget.pkg.github.com/${{ github.repository_owner }}/index.json" \
        -storepasswordincleartext \
        -name "GitHub" \
        -username "${{ github.repository_owner }}" \
        -password "${{ secrets.GITHUB_TOKEN }}"
        mono `./dependencies/vcpkg/vcpkg fetch nuget | tail -n 1` \
        setapikey "${{ secrets.GITHUB_TOKEN }}" \
        -source "https://nuget.pkg.github.com/${{ github.repository_owner }}/index.json"
        
    - name: "cmake x64"
      run: |
        mkdir build
        cd build
        cmake .. ${{ env.BUILD_FLAGS }} \
        -DCMAKE_BUILD_TYPE=${{ env.BUILD_MODE }} \
        -DCMAKE_OSX_ARCHITECTURES=x86_64 \
        -DMACOS_BUNDLE=ON \
        -G Ninja
        
    - name: "Build Cemu x64"
      run: |
        cmake --build build

    - name: "Move x64 artifact"
      run: |
        mkdir bin/x64
        mv bin/Cemu_release.app bin/x64/Cemu.app
        mv bin/x64/Cemu.app/Contents/MacOS/Cemu_release bin/x64/Cemu.app/Contents/MacOS/Cemu
        sed -i '' 's/Cemu_release/Cemu/g' bin/x64/Cemu.app/Contents/Info.plist
        chmod a+x bin/x64/Cemu.app/Contents/MacOS/{Cemu,update.sh}
        codesign --force --deep --sign - bin/x64/Cemu.app

    - name: "cmake arm64"
      run: |
        rm -rf build
        mkdir build
        cd build
        cmake .. ${{ env.BUILD_FLAGS }} \
        -DCMAKE_BUILD_TYPE=${{ env.BUILD_MODE }} \
        -DCMAKE_OSX_ARCHITECTURES=arm64 \
        -DMACOS_BUNDLE=ON \
        -G Ninja

    - name: "Build Cemu arm64"
      run: |
        cmake --build build

    - name: "Move arm64 artifact"
      run: |
        mkdir bin/arm64
        mkdir bin/universal
        mv bin/Cemu_release.app bin/arm64/Cemu.app
        cp -R bin/arm64/Cemu.app bin/universal/Cemu.app
        mv bin/arm64/Cemu.app/Contents/MacOS/Cemu_release bin/arm64/Cemu.app/Contents/MacOS/Cemu
        sed -i '' 's/Cemu_release/Cemu/g' bin/arm64/Cemu.app/Contents/Info.plist
        chmod a+x bin/arm64/Cemu.app/Contents/MacOS/{Cemu,update.sh}
        codesign --force --deep --sign - bin/arm64/Cemu.app

    - name: "Create Universal binary"
      run: |
        rm bin/universal/Cemu.app/Contents/MacOS/Cemu_release
        lipo -create -output bin/universal/Cemu.app/Contents/MacOS/Cemu \
        bin/x64/Cemu.app/Contents/MacOS/Cemu \
        bin/arm64/Cemu.app/Contents/MacOS/Cemu
        rm bin/universal/Cemu.app/Contents/Frameworks/libusb-1.0.0.dylib
        lipo -create -output bin/universal/Cemu.app/Contents/Frameworks/libusb-1.0.0.dylib \
        bin/x64/Cemu.app/Contents/Frameworks/libusb-1.0.0.dylib \
        bin/arm64/Cemu.app/Contents/Frameworks/libusb-1.0.0.dylib
        sed -i '' 's/Cemu_release/Cemu/g' bin/universal/Cemu.app/Contents/Info.plist
        chmod a+x bin/universal/Cemu.app/Contents/MacOS/{Cemu,update.sh}
        codesign --force --deep --sign - bin/universal/Cemu.app
        rm -rf bin/x64/Cemu.app bin/arm64/Cemu.app

    - name: Prepare artifacts
      run: |
        mkdir bin/Cemu_app_universal
        mv bin/universal/Cemu.app bin/Cemu_app_universal/Cemu.app
        ln -s /Applications bin/Cemu_app_universal/Applications
        hdiutil create ./bin/tmp.dmg -ov -volname "Cemu" -fs HFS+ -srcfolder "./bin/Cemu_app_universal"
        hdiutil convert ./bin/tmp.dmg -format UDZO -o bin/Cemu_universal.dmg
        rm bin/tmp.dmg
              
    - name: Upload universal artifact
      uses: actions/upload-artifact@v4
      with:
        name: cemu-bin-macos-universal
        path: ./bin/Cemu_universal.dmg
```
I suggest building a universal version would be better, to avoid distinguishing between x86_64 and arm64 versions. After all, the universal version can run directly on both x86_64 and arm64 platforms. Moreover, on arm64 platforms, the universal version can switch to running the x86_64 version through the Rosetta 2 option.

Exzap:
@hauntek Feel free to open a separate PR for building the universal binary

RedBlackAka:
Individual arch builds have the benefit of being smaller and launching faster. They should be keep, even if universal is regarded as the main distribution

Issue #217, opened 2022-09-09
H264 videos get stuck if the CPU isn't fast enough


Adamillo:
Since the inclusion of the new h264 software video decoder this issue has made games like Rayman Legends not playable as it's impossible to skip those video cutscenes. The previous implementation of the h264 decoder doesn't have this issue

Exzap:
Originally we wanted to replace DXVA with Vulkan's video decoding API. But unfortunately H264 decoding adoption was and still is [minimal](http://vulkan.gpuinfo.org/listdevicescoverage.php?extension=VK_EXT_video_decode_h264&platform=all).

DXVA had it's fair share of problems in regards to buggy drivers so I rather not bring that back, especially since it's Windows only. In general I would prefer software decoding over GPU accelerated decoding as its much more robust and not prone to driver issues.

Potential alternative options: 
- ffmpeg's h264 decoder because it has better performance than the current ih264d decoder we use
- OpenH264 might become viable in the future. Right now it's still lacking full main profile and high profile support.
- Some other MPL friendly library we could use? We have more options now that we are open-source.

There may also be potential to squeeze extra performance out of ih264d. Since it's an undocumented black box it's capabilities are mostly a mystery without digging into the internals.

Haxorzz:
It seems yuzu uses FFMPEG for its video decoding, and has support for AMD and Nvidia GPUs on both windows and linux

Zopolis4:
`ih264d` is also a large implement to removing the `dependencies` folder, so I am in favor of looking for a more traditional replacement.

JaimieVandenbergh:
Just a note that the initial Rayman Legends video wedges on M1 Mac too, in 2.0-47. So perhaps it's not core performance based.
<img width="1161" alt="image" src="https://github.com/cemu-project/Cemu/assets/82263742/1e699a13-89f8-48cf-b809-896759a3bcfc">

foxsouns:
i am running into this issue with most cutscenes in breath of the wild. i have generally underpowered specs (i5-7400@3GHz 4core, nvidia gtx1060 3gb, 16gb ram) & can only maintain a consistent 20 fps (peaking at 30) in this game, turning off all graphics mods (fps+, config, xbox buttons) does not fix the issue, thus i believe it is likely performance based. i see there was talking about testing replacements for the h264 renderer: has anyone else had success in this? i may attempt to set up a dev environment and fiddle around with building it myself with ffmpeg to see if that helps.

PR #1257, opened 2024-07-15
h264: Use asynchronous decoding when possible

Exzap:
By design `h264.rpl` exposes only a synchronous API where decoding only happens during `H264DECExecute` calls. However, if H264DEC runs in buffered mode we can use asynchronous decoding under the hood while still simulating a synchronous API on top of it. In most cases this lets us hide the latency of our software decoder which helps with games that are picky about decoding times.

Should help with #217

Marking this as draft for now since I still have to go through all the games with videos to make sure there are no regressions.

### DIFF
--- a/src/Cafe/CMakeLists.txt
+++ b/src/Cafe/CMakeLists.txt
@@ -374,7 +374,9 @@ add_library(CemuCafe
   OS/libs/gx2/GX2_Texture.h
   OS/libs/gx2/GX2_TilingAperture.cpp
   OS/libs/h264_avc/H264Dec.cpp
+  OS/libs/h264_avc/H264DecBackendAVC.cpp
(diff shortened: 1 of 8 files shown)


Adamillo:
Super Smash Bros For Wii U intro cutscene still gets easily stuck even with this PR

Exzap:
Thanks for testing this. Can you enable `Debug -> CafeOS Modules logging -> h264 API` and then upload log.txt after it freezes

Adamillo:
[log.txt](https://github.com/user-attachments/files/16365705/log.txt)

Exzap:
The good news is that from the log it looks like the PR is working and decoding times are very low. Bad news is that this means there are other factors outside the decoder itself that can trigger the freezes. Maybe slow file reading speed? I will do some testing when I get a chance

PR #1288, opened 2024-08-12
Add support for iterating directories in graphics pack content folders.

blueskythlikesclouds:
Allows games to see new files in graphics packs when iterating directories.

Example: Bayonetta 2, when you patch its code to load files outside CPKs, uses directory iterate functions to find files. This change allows it to find files in graphics packs.


Exzap:
Looks good to me. Thanks!

PR #1306, opened 2024-08-23
Preparations for 2.1

Exzap:
For 2.1, we are overhauling how we handle releases by dropping the distinction between stable and experimental builds. The previous setup had experimental releases coming out pretty frequently, but the stable release took so long to happen that the experimental version became the recommended version for people to use long ago, contrary to the original idea behind it.

The main issue with any stable release is that Cemu never feels quite “stable enough” and the extra work to maintain a separate release channel, write detailed changelogs, etc., just adds to the mental barrier of releasing one. So, from now on, there’s only going to be one update stream.

To avoid pushing out potentially buggy updates to **everyone** (we are a small team, things slip through the cracks), I have added a new option in the UI: `Receive untested updates` If enabled, the user will get the latest release as soon as it drops. If disabled, updates will be delayed by 1-3 weeks, giving us time to catch and fix any issues that pop up. We can also manually change the delay or skip specific updates if necessary. By default this option is off.

**Summary of the changes in this PR:**
- Dropping the distinction between stable and experimental releases
- Setting the CI and build system up for 2.1 and future releases
- From now on the release numbering will be 2.1, 2.2, 2.3, 2.4, etc.
- Each release on github will come with a changelog from now on
- Added an option "Receive untested updates" (as explained above) 
- Some last minute bug fixes and tweaks

A note for maintainers of unofficial Cemu packages: Instead of `-DEXPERIMENTAL_VERSION` there is now `-DEMULATOR_VERSION_MAJOR=...` and `-DEMULATOR_VERSION_MINOR=...` matching the SemVer naming of the version components. Patch is implied to be zero (2.1 is the same as 2.1.0, we just shorten it for simplicity sake)


Ammar-Sadaoui:
so we are ready to add "VK_KHR_synchronization2" back ?

Exzap:
> so we are ready to add "VK_KHR_synchronization2" back ?

This question is opening a can of worms and the whole discussion is outside the scope for this particular PR. There are indeed plans to bump the required Vulkan version so we can switch over to `VK_KHR_dynamic_rendering`. This would also give us access to `VK_KHR_synchronization2`. Whether it is worth doing this is up in the air still. Advantages on paper don't necessarily translate to real drivers.

Ammar-Sadaoui:
> This question is opening a can of worms and the whole discussion is outside the scope for this particular PR

sorry [Exzap](https://github.com/Exzap)
You're right; there are more important things to address first, such as developing a new shader translator that directly supports SPIR-V output and other crucial improvements.

nobleflame:
Out of interest, when I update CEMU via the prompt, it downloads the update, patches it, and then closes / reloads CEMU. All good, except that I get an application crash error in the applications section of event viewer.

It doesn’t appear to affect anything, but just wondering why this happens and if I can do anything to troubleshoot or fix it?

Again, this doesn’t affect functionality at all - it’s just a false positive error - but I don’t like errors.

war59312:
Every time I uncheck "Receive untested updates" it comes back checked. In 2.1 official release.

Exzap:
> Out of interest, when I update CEMU via the prompt, it downloads the update, patches it, and then closes / reloads CEMU. All good, except that I get an application crash error in the applications section of event viewer.

This is a known issue. Just super low priority so it hasn't been addressed yet.

> Every time I uncheck "Receive untested updates" it comes back checked. In 2.1 official release.

yeah I can replicate this, thanks for reporting. I'll fix it for the next release

nobleflame:
> This is a known issue. Just super low priority so it hasn't been addressed yet.

That’s great to know. Thanks for clarifying!

Home0x:
please add the usb emulated device dimensions toy pad

PR #1310, opened 2024-08-25
build: Better support for cross-compiling on ARM macOS.

squidbus:
## Description

This implements and utilizes better support for cross-compiling on ARM macOS.

On the build side:
* Fixes architecture detection for ih264d and CemuAsm by using the value of `CMAKE_OSX_ARCHITECTURES` when available. Also added support for detecting the `arm64` value used on Mac while I was there for future ARM support.
* Sets `CMAKE_OSX_DEPLOYMENT_TARGET` to the minimum supported OS version to make sure the output executable has the correct support.
* Updates build instructions:
  * An x86_64 shell is no longer necessary for cross-compiling; it works fine as long as you pass the right architecture to CMake and install x86_64 versions of MoltenVK and Boost. I split up the dependencies section into tools and libraries, put instructions for installing x86_64 libraries from Homebrew, and removed references to needing to compile in an x86_64 shell every time.
  * Homebrew LLVM is no longer necessary to compile Cemu on newer versions of Xcode. In fact I was having some trouble with LLVM 15 not working right with newer macOS SDK versions.

On the CI side:
* Updates to the latest macOS GitHub Actions runners, which use Apple Silicon Macs. It's likely that once GitHub deprecates the macOS versions that use Intel runners, there will only be free ARM runners left, so it's best to get support for this going before that happens. Plus these runners tend to perform better than the older Intel Mac runners.
* Speeds up build times by using the pre-built MoltenVK releases instead of building from source. We can still get the older v1.2.9 this way without spending a lot of time building everything (usually about 16 minutes), and for future use it is a universal binary so it works for any architecture.
* Adds architecture specifier and removes LLVM compiler override, for the same reasons as the build instructions updates.
* Removes redundant `cmake` install as it is installed by `setup-cmake-action` later.

## Testing

I tested both building on my Mac and building in GitHub Actions. From my testing, the Mac CI build now takes about 7 minutes and 30 seconds to complete with `vcpkg` dependencies cached.

I also downloaded the resulting GItHub Actions build and ran a few games to check everything is working.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -195,7 +195,7 @@ jobs:
         path: ./bin/Cemu.exe
 
   build-macos:
-    runs-on: macos-12
+    runs-on: macos-14
     steps:
     - name: "Checkout repo"
       uses: actions/checkout@v4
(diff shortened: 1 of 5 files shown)


Exzap:
This looks very good. Thanks for the detailed explanation. I will merge this after the upcoming 2.1 release

PR #1316, opened 2024-08-28
Allow the appimage to update with appimageupdate

Samueru-sama:
This change allows delta updates with [appimageupdatetool](https://github.com/AppImageCommunity/AppImageUpdate)

I noticed that there was old release info in the build script, but checking the releases it seems it never got implemented?

Also was there a specific reason to be using mkappimage instead of appimagetool? as far as I know the only difference used to be that mkappimage was pulling the static runtime, but that is something that now appimagetool does as well, so functionality wise there is no difference I think.

Note that ideally the zsync should be released on the latest release and not in the pre-releases, else the entire update info and the release tags would need to be changed to allow updates on pre-releases.

### DIFF
--- a/dist/linux/appimage.sh
+++ b/dist/linux/appimage.sh
@@ -6,8 +6,8 @@ fi
 
 curl -sSfLO "https://github.com/linuxdeploy/linuxdeploy/releases/download/continuous/linuxdeploy-x86_64.AppImage"
 chmod a+x linuxdeploy*.AppImage
-curl -sSfL https://github.com"$(curl https://github.com/probonopd/go-appimage/releases/expanded_assets/continuous | grep "mkappimage-.*-x86_64.AppImage" | head -n 1 | cut -d '"' -f 2)" -o mkappimage.AppImage
-chmod a+x mkappimage.AppImage
+curl -sSfL "https://github.com/AppImage/appimagetool/releases/download/continuous/appimagetool-x86_64.AppImage" -o appimagetool
+chmod a+x appimagetool
 curl -sSfLO "https://raw.githubusercontent.com/linuxdeploy/linuxdeploy-plugin-gtk/master/linuxdeploy-plugin-gtk.sh"
 chmod a+x linuxdeploy-plugin-gtk.sh
 curl -sSfLO "https://github.com/darealshinji/linuxdeploy-plugin-checkrt/releases/download/continuous/linuxdeploy-plugin-checkrt.sh"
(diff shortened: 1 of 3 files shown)


qurious-pixel:
For appimageupdate, the .zsync2 file has to be hosted to a known location. This can be the github releases page or a download webpage, but that info gets embedded into the appimage (which currently it is not).

The most recent update to 2.1 has changed somethings, so it may be wiser _for me_ to wait until 2.2 to see how the update broke.

dist/linux/appimage.sh
```diff
@@ -6,8 +6,8 @@ fi
 
 curl -sSfLO "https://github.com/linuxdeploy/linuxdeploy/releases/download/continuous/linuxdeploy-x86_64.AppImage"
 chmod a+x linuxdeploy*.AppImage
-curl -sSfL https://github.com"$(curl https://github.com/probonopd/go-appimage/releases/expanded_assets/continuous | grep "mkappimage-.*-x86_64.AppImage" | head -n 1 | cut -d '"' -f 2)" -o mkappimage.AppImage
-chmod a+x mkappimage.AppImage
+curl -sSfL "https://github.com/AppImage/appimagetool/releases/download/continuous/appimagetool-x86_64.AppImage" -o appimagetool
+chmod a+x appimagetool
```
qurious-pixel, line 10: The reason we use `mkappimage` instead of `appimagetool` is because of the change of libfuse2 no longer standard on many newer distros. mkappimage statically links this, avoiding another requirement for the host system.

dist/linux/appimage.sh
```diff
@@ -34,9 +34,10 @@ chmod +x AppDir/usr/bin/Cemu
 
 cp /usr/lib/x86_64-linux-gnu/{libsepol.so.1,libffi.so.7,libpcre.so.3,libGLU.so.1,libthai.so.0} AppDir/usr/lib
 
-export UPD_INFO="gh-releases-zsync|cemu-project|Cemu|ci|Cemu.AppImage.zsync"
+export UPD_INFO="gh-releases-zsync|$GITHUB_REPOSITORY_OWNER|Cemu|latest|Cemu-*x86_64.AppImage.zsync"
```
qurious-pixel, line 37: The author specifically recommends _against_ using `latest` and suggests using continuous or just using version tags.

Samueru-sama: Are you sure? Because I had this issue: https://github.com/AppImageCommunity/AppImageUpdate/issues/233

I originally tried to get it working but couldn't because Cemu used to release as pre-releases with varying tag names, I tried continuous, ci, testing, etc and none worked.

Now Cemu is going to make more stable releases instead of making pre-releases, so I rebased my changes and made this PR.

dist/linux/appimage.sh
Samueru-sama: Good that's what I meant with the static runtime. AppImage/appimagetool also pulls the static runtime now.

Samueru-sama:
> For appimageupdate, the .zsync2 file has to be hosted to a known location. This can be the github releases page or a download webpage, but that info gets embedded into the appimage (which currently it is not).
> 
> The most recent update to 2.1 has changed somethings, so it may be wiser _for me_ to wait until 2.2 to see how the update broke.

Hi, I don't know what you meant in the last phrase, what do you mean by wait to see how the update broke? you mean the internal update mechanism that Cemu has?

Anyway, how the release is setup right now it zsync works, you can see it in my fork:

https://github.com/Samueru-sama/Cemu/releases/tag/v2.1

image

qurious-pixel:
From your error message, I had assumed something broke due to the naming convention change from Cemu-2.0 to Cemu-2.1. I had to make some changes to scripts I manage, and flatpak maintainer did too. But the update still works from Cemu-2.0 to Cemu-2.1.

I also use appimageupdate for projects that do not self-update. Your changes appear to work including the new appimagetool, which recently switched to fuse3. 

Can you do a version 2.2 on your repo? Cemu-2.0-7 update fails. 
Version 2.1 updates, but yields the last commit from your `test6` branch, as version identifier. `Cemu-adf0abf-x86_64.AppImage`

Samueru-sama:
> From your error message, I had assumed something broke due to the naming convention change from Cemu-2.0 to Cemu-2.1. I had to make some changes to scripts I manage, and flatpak maintainer did too. But the update still works from Cemu-2.0 to Cemu-2.1.
> 
> I also use appimageupdate for projects that do not self-update. Your changes appear to work including the new appimagetool, which recently switched to fuse3.
> 
> Can you do a version 2.2 on your repo? Cemu-2.0-7 update fails. Version 2.1 updates, but yields the last commit from your `test6` branch, as version identifier. `Cemu-adf0abf-x86_64.AppImage`

Done, and yeah something goes wrong when updating: 

image

PR #1292, opened 2024-08-13
Update appimage.sh to support runtime libstdc++.so.6 loading

20943204920434:
Add [checkrt](https://github.com/darealshinji/linuxdeploy-plugin-checkrt) plugin in order to detect the correct `libstdc++.so.6` version at runtime. The current method of bundling `libstdc++.so.6` is wrong and may cause conflicts ([1](https://discourse.appimage.org/t/im-a-big-fan-of-this-but-graphics-driver-libstdc-conflict/171/3), [2](https://github.com/cemu-project/Cemu/issues/1280)) if the base system contains fresh packages.  For context, this is the [recommended](https://github.com/AppImage/AppImageKit/wiki/Creating-AppImages/cc2441518975caca23e9ce2dba6f08a22c678d1e#libstdcso6) way to do this.

PR #1319, opened 2024-08-30
appimage.sh: Do not copy `libstdc++.so.6` to `usr/lib/`

MoonlightWave-12:
This is supposed to be handled by checkrt now, but if the library is available in `usr/lib/` in the AppImage, it will be loaded from there instead.

Fixes https://github.com/cemu-project/Cemu/pull/1292

### DIFF
--- a/dist/linux/appimage.sh
+++ b/dist/linux/appimage.sh
@@ -50,7 +50,6 @@ fi
 echo "Cemu Version Cemu-${GITVERSION}"
 
 rm AppDir/usr/lib/libwayland-client.so.0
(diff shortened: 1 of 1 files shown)


Exzap:
On [this page](https://github.com/AppImage/AppImageKit/wiki/Creating-AppImages/cc2441518975caca23e9ce2dba6f08a22c678d1e#libstdcso6) it talks about loading libstdc conditionally from `usr/optional/libstdc++/libstdc++.so.6`. So maybe thats the location we should copy it to instead of omitting it completely?

PR #1324, opened 2024-09-02
nsysnet: Check for null timeout in select call

deReeperJosh:
After bisecting an error with Skylanders Superchargers and Imaginators, I found that after this commit https://github.com/cemu-project/Cemu/commit/bab1616565b8fab99c52e0bedb275ae90a7b53df was added, the nsysnet select method was being called with a non-null write fdset, but a null timeout. I have added a null check to the code to return early if this timeOut is null.

### DIFF
--- a/src/Cafe/OS/libs/nsysnet/nsysnet.cpp
+++ b/src/Cafe/OS/libs/nsysnet/nsysnet.cpp
@@ -1210,6 +1210,14 @@ void nsysnetExport_select(PPCInterpreter_t* hCPU)
 
 	timeval tv = { 0 };
 
+	if (timeOut == NULL)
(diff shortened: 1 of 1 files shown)


Exzap:
Thanks for tracing this down to select(). I am not happy with the implementation because, according to man pages, a NULL timeout is an infinite timeout. This is also doesn't simulate a zero timeout and instead quits right away before checking any of the sockets. Whether Wii U follows any of these behaviors is another question.

But at the end of the day this is acceptable as a workaround and our select() implementation is all sorts of hacky already anyway. So I will merge this as-is and look into fixing up the select implementation when I get a chance. Thanks again for looking into this!

PR #1344, opened 2024-09-16
[IOSU/BOSS] Fix BOSS not honoring the proxy_server setting.

AToska21:
Special thanks to Exzap in the Discord server for figuring this out!


Exzap:
thanks for the PR!

PR #1353, opened 2024-09-26
Linux: Allow connecting Wiimotes via L2CAP

capitalistspz:
Requires Bluez.

This supports 3rd-party Wiimotes that are "invisible to scans"

Pairing a controller is done through the `Pair Wii or Wii U Controller` button dialog. The pairing is temporary, and will only last as long as Cemu is open. 

If the controller is disconnected while Cemu is open, it can be reconnected by putting the controller into sync mode. You do not need to click `Pair Wii or Wii U Controller` again.

~~At the moment not recommended to merge, disconnecting the controller after pairing causes hitching every 2 seconds because `connect` takes too long to timeout.~~ Fixed by https://github.com/cemu-project/Cemu/pull/1353/commits/ae2f5d23b32c547b63c020dacac5c1f3fed8bf2c

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -39,7 +39,7 @@ jobs:
     - name: "Install system dependencies"
       run: |
         sudo apt update -qq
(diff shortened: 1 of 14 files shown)


amppz:
Commit #ae2f5d2 should have fixed the hitching

ilikefatpeople1026-source:
Yes please do

PR #1354, opened 2024-09-28
Migration from Vulkan-Headers to VCPKG + updating submodules

Xphalnos:
I moved **Vulkan-Headers** to **VCPKG** for easier maintenance.
This will also ensure that whenever **glslang** is updated, the latest **Vulkan-SDK** version will be usable
I could only update **imgui** to **1.90.1** because of compilation errors for later versions. Same for **fmt** and **OpenSSL**.

| Library | Old | New |
| ------- | --- | --- |
| imgui | 1.89.4 | 1.90.1 |
| Vulkan-Headers | 1.3.245 | 1.3.290 |
| glslang | 14.2.0 | 14.3.0 |
| Curl | 8.8.0 | 8.10.1 |
| libjpeg-turbo | 3.0.3 | 3.0.4 |

Tested with **Vulkan** on **Mario Kart 8** and **Zelda Breath of the Wild**:
Image1
Image2

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,10 +10,6 @@
 	path = dependencies/vcpkg
 	url = https://github.com/microsoft/vcpkg
 	shallow = false
-[submodule "dependencies/Vulkan-Headers"]
-	path = dependencies/Vulkan-Headers
-	url = https://github.com/KhronosGroup/Vulkan-Headers
-	shallow = true
 [submodule "dependencies/imgui"]
 	path = dependencies/imgui
 	url = https://github.com/ocornut/imgui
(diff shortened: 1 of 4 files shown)


Exzap:
We prefer to handle dependency updates internally because there is always a chance of things breaking in subtle ways, so thorough testing is necessary on all operating systems. I already mentioned this [last time](https://github.com/cemu-project/Cemu/pull/1198#issuecomment-2104082300) :wink:

I'll be closing this since I don't think any of the updated libraries really give us any immediate benefit but also because I disagree with moving Vulkan-Headers to vcpkg

PR #1365, opened 2024-09-30
Update GeneralSettings2.cpp: Add "and size"

davidhedlund:
Added "and size" to "Remember pad window position and size"

### DIFF
--- a/src/gui/GeneralSettings2.cpp
+++ b/src/gui/GeneralSettings2.cpp
@@ -161,12 +161,12 @@ wxPanel* GeneralSettings2::AddGeneralPage(wxNotebook* notebook)
 			};
 
 			const int topflag = wxALIGN_CENTER_VERTICAL | wxALL;
-			m_save_window_position_size = new wxCheckBox(box, wxID_ANY, _("Remember main window position and size"));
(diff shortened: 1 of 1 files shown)


Exzap:
Thanks for the pull request. I am not sure if we should merge this as-is because to me it looks off now because of how long the text is:
image

versus before:
image

Maybe there is a way to phrase it more concisely?

davidhedlund:
> Thanks for the pull request. I am not sure if we should merge this as-is because to me it looks off now because of how long the text is: image
> 
> versus before: image
> 
> Maybe there is a way to phrase it more concisely?

How about "Remember main window position/size"?

davidhedlund:
Please review the latest commit

Exzap:
Using slashes in UI text is problematic for readability and localization.
I thought about it some more and I believe the original text is ok for now. Maybe I can come up with something better in the future.
At any rate, the tooltip explains that it also affects the window size and in the 7 years of troubleshooting I have done nobody ever was confused about the phrasing of this option so I don't think this really needs to be changed anyway.

Thanks for taking the time to submit the issue and PR

PR #1367, opened 2024-10-03
make this script more readable

Drsheppard01:
Only cosmetic changes here, but I'd be happy to have this merged


Exzap:
imho this is less readable but thank you for the PR regardless

Issue #1268, opened 2024-07-23
LEGO Dimensions ToyPad Emulation


Fancy2209:
### Your suggestion

Adding the ability to emulate the Lego Dimensions Portal (Toy-Pad) like what done for the Skylanders Portal (https://github.com/cemu-project/Cemu/pull/971) and the Infinity Base (https://github.com/cemu-project/Cemu/pull/1246)
There is a [Hardware Level Emulator that uses a RPi](https://github.com/Berny23/LD-ToyPad-Emulator), but a Software one in CEMU would be neat.

PR #1371, opened 2024-10-06
nsyshid: Emulate Dimensions Toypad

deReeperJosh:
This PR aims to emulate the last mainstream toys-to-life peripheral, the Lego Dimensions Toypad. The Toypad is functionally very similar to the Disney Infinity Base, it has 2 endpoints, one for reading and one for writing, and all requests and responses are padded to 32 bytes.

The UI I have created functions mostly the same as the Skylander Portal and Infinity Base, but the layout is slightly different, to try to match the positions of a physical portal:

image

Users can Clear minifgures from the UI, Create either a new Character, or a 'Blank Tag' which can be used to store gadgets from the game, or Load an existing dump.

The game quite frequently requires figures to be moved to different positions on the Toypad, so I have added in a move button, which opens up a new UI window allowing users to quickly select a new position to move one of their loaded figures on to:

image

Any testing would be much appreciated, I have the game but a lot of different sections of the game have different ways to interact with the Toypad, so there is potential for some areas of the game to require slight variations to adding/removing responses etc.

For anyone interested, [here is a video](https://youtu.be/v-htDhg1Pas) of the tool in action

This PR should close https://github.com/cemu-project/Cemu/issues/1268.

### DIFF
--- a/src/Cafe/OS/libs/nsyshid/Dimensions.cpp
+++ b/src/Cafe/OS/libs/nsyshid/Dimensions.cpp
@@ -416,7 +416,10 @@ namespace nsyshid
 
 	Device::WriteResult DimensionsToypadDevice::Write(WriteMessage* message)
 	{
-		g_dimensionstoypad.SendCommand(message->data, message->length);
+		if (message->length != 32)
+			return Device::WriteResult::Error;
+
+		g_dimensionstoypad.SendCommand(std::span<const uint8, 32>{message->data, 32});
 		message->bytesWritten = message->length;
 		return Device::WriteResult::Success;
 	}
@@ -507,14 +510,13 @@ namespace nsyshid
 		{
 			if (!m_figureAddedRemovedResponses.empty())
 			{
-				memcpy(response.data(), m_figureAddedRemovedResponses.front().data(),
-					   0x20);
+				response = m_figureAddedRemovedResponses.front();
 				m_figureAddedRemovedResponses.pop();
 				responded = true;
 			}
 			else if (!m_queries.empty())
 			{
-				memcpy(response.data(), m_queries.front().data(), 0x20);
+				response = m_queries.front();
 				m_queries.pop();
 				responded = true;
 			}
@@ -527,7 +529,7 @@ namespace nsyshid
 		return response;
 	}
 
-	void DimensionsUSB::SendCommand(uint8* buf, sint32 originalLength)
+	void DimensionsUSB::SendCommand(std::span<const uint8, 32> buf)
 	{
 		const uint8 command = buf[2];
 		const uint8 sequence = buf[3];
@@ -547,13 +549,13 @@ namespace nsyshid
 		case 0xB1: // Seed
 		{
 			// Initialise a random number generator using the seed provided
-			g_dimensionstoypad.GenerateRandomNumber(&buf[4], sequence, q_result);
+			g_dimensionstoypad.GenerateRandomNumber(std::span<const uint8, 8>{buf.data() + 4, 8}, sequence, q_result);
 			break;
 		}
 		case 0xB3: // Challenge
 		{
 			// Get the next number in the sequence based on the RNG from 0xB1 command
-			g_dimensionstoypad.GetChallengeResponse(&buf[4], sequence, q_result);
+			g_dimensionstoypad.GetChallengeResponse(std::span<const uint8, 8>{buf.data() + 4, 8}, sequence, q_result);
 			break;
 		}
 		case 0xC0: // Color
@@ -579,13 +581,13 @@ namespace nsyshid
 		case 0xD3: // Write
 		{
 			// Write 4 bytes to page buf[5] to the figure at index buf[4]
-			g_dimensionstoypad.WriteBlock(buf[4], buf[5], &buf[6], q_result, sequence);
+			g_dimensionstoypad.WriteBlock(buf[4], buf[5], std::span<const uint8, 16>{buf.data() + 6, 16}, q_result, sequence);
 			break;
 		}
 		case 0xD4: // Model
 		{
 			// Get the model id of the figure at index buf[4]
-			g_dimensionstoypad.GetModel(&buf[4], sequence, q_result);
+			g_dimensionstoypad.GetModel(std::span<const uint8, 8>{buf.data() + 4, 8}, sequence, q_result);
 			break;
 		}
 		case 0xD0: // Tag List
@@ -673,7 +675,7 @@ namespace nsyshid
 			std::array<uint8, 8> valueToEncrypt = {uint8(id & 0xFF), uint8((id >> 8) & 0xFF), uint8((id >> 16) & 0xFF), uint8((id >> 24) & 0xFF),
 												   uint8(id & 0xFF), uint8((id >> 8) & 0xFF), uint8((id >> 16) & 0xFF), uint8((id >> 24) & 0xFF)};
 
-			std::array<uint8, 8> encrypted = Encrypt(valueToEncrypt.data(), figureKey);
+			std::array<uint8, 8> encrypted = Encrypt(valueToEncrypt, figureKey);
 
 			std::memcpy(&fileData[36 * 4], &encrypted[0], 4);
 			std::memcpy(&fileData[37 * 4], &encrypted[4], 4);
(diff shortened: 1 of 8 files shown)


src/Cafe/OS/libs/nsyshid/Dimensions.cpp
```diff
@@ -0,0 +1,1076 @@
...
+	{
+		uint32 e = m_randomA - std::rotl(m_randomB, 21);
+		m_randomA = m_randomB ^ std::rotl(m_randomC, 19);
+		m_randomB = m_randomC + std::rotl(m_randomD, 6);
+		m_randomC = m_randomD + e;
+		m_randomD = e + m_randomA;
+		return m_randomD;
+	}
+
+	std::array<uint8, 8> DimensionsUSB::Decrypt(const uint8* buf, std::optional<std::array<uint8, 16>> key)
```
goeiecool9999, line 766: This function blindly indexes a pointer and assumes it's an array that has at least 8 elements. This is fine as long as you validate that this function is only called with that precondition satisfied.
I know I risk this review being nothing but "use span" spam, but changing ``const uint8* buf`` to ``std::span<const uint8, 8>`` makes it harder to call the function with a buffer pointing to invalid memory in the first place, so you're less likely to make a mistake.
same goes for DimensionsUSB::Encrypt

src/Cafe/OS/libs/nsyshid/Dimensions.cpp
```diff
@@ -0,0 +1,1076 @@
...
+			scrambled = b + v4 + v5 - scrambled;
+		}
+		return {uint8(scrambled & 0xFF), uint8(scrambled >> 8 & 0xFF), uint8(scrambled >> 16 & 0xFF), uint8(scrambled >> 24 & 0xFF)};
+	}
+
+	uint32 DimensionsUSB::GetFigureId(const std::array<uint8, 0x2D * 0x04>& buf)
+	{
+		const std::array<uint8, 16> figureKey = GenerateFigureKey(buf);
+
+		const std::array<uint8, 8> decrypted = Decrypt(&buf[36 * 4], figureKey);
```
goeiecool9999, line 925: If you do change the argument type to std::span, ``&buf[36 * 4]`` would need to be replaced with something like ``std::span<const uint8, 8>{buf.begin() + 36 * 4, 8}`` or shorter but less C++-like ``(const uint8(&)[8])buf[36 * 4]``, both of which are fairly lengthy for a function argument.
Considering that in this case you access the same area of the buffer later in the function you could define the span as a local variable that you use every time you want to access something at 36 * 4. That way the argument for Decrypt can be short, and you repeat that index less often so the code is less prone to typo's. (not that it matters if you wrote it correctly, but "Do not Repeat Yourself" is good practice regardless)

src/Cafe/OS/libs/nsyshid/Dimensions.cpp
```diff
@@ -0,0 +1,1076 @@
...
+
+	void DimensionsUSB::GenerateRandomNumber(uint8* buf, uint8 sequence,
+											 std::array<uint8, 32>& replyBuf)
+	{
+		// Decrypt payload into an 8 byte array
+		std::array<uint8, 8> value = Decrypt(buf, std::nullopt);
+		// Seed is the first 4 bytes (little endian) of the decrypted payload
+		uint32 seed = uint32(value[3]) << 24 | uint32(value[2]) << 16 | uint32(value[1]) << 8 | uint32(value[0]);
+		// Confirmation is the second 4 bytes (big endian) of the decrypted payload
+		uint32 conf = uint32(value[4]) << 24 | uint32(value[5]) << 16 | uint32(value[6]) << 8 | uint32(value[7]);
```
goeiecool9999, line 706: This comment applies to all instances where you manually change endianness of a value. Cemu has utility code for this. Also loading a little endian value does not require extra work. You can just cast the reference.

```suggestion
		uint32 seed = uint32be::from_bevalue((uint32&)value[0]);
		// Confirmation is the second 4 bytes (big endian) of the decrypted payload
		uint32 conf = (uint32&)&value[4]
```

src/Cafe/OS/libs/nsyshid/Dimensions.cpp
deReeperJosh: Same as above, decrypt is called sometimes from requests from the game (where the message is a pointer) which is why the method takes in a pointer. If there is a way to convert the pointer to a span of size originalLength as above, then this can change the rest of the methods to allow for a span param

src/Cafe/OS/libs/nsyshid/Dimensions.cpp
goeiecool9999: > Does this apply cross platform (for the little endian code), because I thought byte order was dependent on platform?

This is true, but Cemu already makes the assumption that it's running on a little endian platform so making it portable across platforms of different endianness is already a huge task and making the assumption in one more place doesn't hurt in my opinion.

src/Cafe/OS/libs/nsyshid/Dimensions.cpp
deReeperJosh: I have had a go at using span across the class - let me know if my implementations look correct, haven't got a whole lot of C++ experience besides the occasional open source contribution, so learning new things is always a bonus here (especially best practices for C++ 20)

src/Cafe/OS/libs/nsyshid/Dimensions.cpp
```diff
@@ -547,13 +549,13 @@ namespace nsyshid
 		case 0xB1: // Seed
 		{
 			// Initialise a random number generator using the seed provided
-			g_dimensionstoypad.GenerateRandomNumber(&buf[4], sequence, q_result);
+			g_dimensionstoypad.GenerateRandomNumber(std::span<const uint8, 8>{buf.data() + 4, 8}, sequence, q_result);
```
goeiecool9999, line 552: ``data()`` should be changed to ``begin()``. They are subtly different. The std::span constructor expects an iterator type. data() returns a pointer type. In practice these are often equivalent, hence why it accepts the pointer as an argument, but this is not guaranteed anywhere. There are other places where you construct a span using a pointer and it should be changed there too.

src/Cafe/OS/libs/nsyshid/Dimensions.cpp
goeiecool9999: Looks good except for what I pointed out below in a different comment

src/Cafe/OS/libs/nsyshid/Dimensions.cpp
```diff
@@ -0,0 +1,1080 @@
...
+	{
+		std::lock_guard lock(m_dimensionsMutex);
+		const uint32 id = GetFigureId(buf);
+
+		DimensionsMini& figure = GetFigureByIndex(index);
+		figure.dimFile = std::move(file);
+		figure.id = id;
+		figure.pad = pad;
+		figure.index = index + 1;
+		memcpy(figure.data.data(), buf.data(), buf.size());
```
goeiecool9999, line 622: This memcpy can also be a simple assignment
```suggestion
		figure.data = buf;
```

deReeperJosh:
Will squash the commits once reviews have been done

goeiecool9999:
> Will squash the commits once reviews have been done

You don't have to. The Cemu repo has merge commits disabled so the options for merging PR's are squash or rebase. Usually squash is chosen so the individual commits do not enter the main tree. It might actually be helpful to keep the separate commits in the PR just in case.

src/Cafe/OS/libs/nsyshid/Dimensions.cpp
```diff
@@ -0,0 +1,1140 @@
...
+	bool DimensionsUSB::RemoveFigure(uint8 pad, uint8 index, bool save, bool lock)
+	{
+		DimensionsMini& figure = GetFigureByIndex(index);
+		if (figure.index == 255)
+		{
+			return false;
+		}
+
+		if (lock)
+			std::shared_lock lock(m_dimensionsMutex);
```
goeiecool9999, line 646: This is not the proper way to use std::shared_lock. The scope of the variable named lock is the if statement. There's only one statement so the scope of that if statement ends immediately. So in this case it grabs the lock on construction and immediately releases it on destruction.
I recommend defining ``std::optional<std::lock_guard> lock;`` unconditionally instead and then constructing it in the if statement.
```suggestion
		std::optional<std::lock_guard> scopeLock;
		if(lock)
			scopeLock.emplace(m_dimensionsMutex);
```
Same goes for all the other places where the lock is acquired like this.

Also I think m_dimensionsMutex doesn't have to be a shared_mutex. It can probably be a regular std::mutex. From [cppreference.com](https://en.cppreference.com/w/cpp/thread/shared_mutex):
> Shared mutexes are especially useful when shared data can be safely read by any number of threads simultaneously, but a thread may only write the same data when no other thread is reading or writing at the same time.

Since many of the functions that acquire the shared lock write to the protected data it could only theoretically cause clashing if multiple threads called these functions simultaneously for example.

goeiecool9999: Looks good. One more thing: DimensionsUSB::GetStatus accesses m_figureAddedRemovedResponses so it should acquire the lock too when it does so.

Issue #1040, opened 2023-12-12
[Win] BOTW Inventory Screen Severely Garbed on Release 2.0-52+ When Not Focused


JamShady:
I'm playing on W11, using a full screen window. Normally I move the mouse off-screen to hide the pointer.

From release 2.0-52 onwards, when I do this, the inventory screen is severely garbled and broken (visually, it still appears to work). The rest of the gameplay is fine, it's just limited to the inventory screen for some reason, and more particularly to the clothing items (although bows & shields do trigger it occasionally too).

![image](https://github.com/cemu-project/Cemu/assets/167395/f53138b2-5c07-4e36-a60e-b2159e92ffb7)
![image](https://github.com/cemu-project/Cemu/assets/167395/88d01f29-b398-46db-ae0a-89cfc9db868d)
![image](https://github.com/cemu-project/Cemu/assets/167395/0a666dc2-ab1f-4d62-835a-3f0f386a32eb)

Squall-Leonhart:
That isn't unique to 2.0, and signs point to third party injection being the cause, or OOM on some graphics card / driver combos.

https://www.reddit.com/r/cemu/comments/gct0vy/botw_inventory_graphic_glitch/

christianbittner:
used to have the same issue a handful of weeks ago. It only occurred on the armor screen for me where the icons just glitched out, causing weird flickering and red lights to appear.
I believe it was related to a graphics pack. I dont remember 100% but I know I changed a setting to fix it, maybe also look into your buffering setting, as well as if you use driver level settings like forcing the output layered on the DXGI swapchain or something like low latency mode. I just tried for 20+min to reproduce the issue but no matter which and how hard I crank the settings it wont appear anymore. 
It seems to be fixed on my end (even tried multiple res). Using a 5800X3D and a GTX 1060 6GB with the latest driver

christianbittner:
Never mind, I got it to appear again. But only when using the double buffer option under vsync.  @JamShady turn tripple buffering on and try again. During my testing I forgot to turn it to full screen all the time before clicking on my 2nd desktop (by default CEMU fades my mouse out after~2 sec, so I dont need to tab out). I retested a bunch of configs again and the ONLY thing making the glitch appear was the double buffer solution of CEMU.

JamShady:
@Squall-Leonhart I appreciate the insight; hard to argue with that evidence.

[...]

Issue #1011, opened 2023-10-29
[Xenoblade Chronicles X] [Vulkan] Graphical Corruption/Flickering in certain menus


Meepster-836:
[log.txt](https://github.com/cemu-project/Cemu/files/13198749/log.txt)
![Screenshot_20231029_102340](https://github.com/cemu-project/Cemu/assets/127622115/fd4dcefb-e8c2-4e1f-8222-13769eef2654)

When in the Game Options menu in Xenoblade X (in-game, not sure if this happens on the title screen as well), there is inconsistent but fast graphical corruption/flickering occuring. The corruption would also very in intensity at random, sometimes weak and only affecting small parts of the screen, and sometimes affecting all of it. The flickering was too fast for me to take a screenshot of. Moving around the options menu might amplify the issue, but I'm not sure.

When in the Skell menu (the terminal in the BLADE Barracks) and customizing the Skell's equipment, there is full screen flickering black and white, also affecting the UI (This is what the attached screenshot shows). I've also seen it occur isolated to just the Skell, flickering between fullbright white and black.

I have not tested Skell submenus other than equipment, but it seems to not happen in the main Skell menu, only when in the equipment submenu.

Detailed system specs are in the log, but I am using:

Ryzen 9 5900X
Radeon RX 6700 XT
32GB DDR4 RAM
KUbuntu 22.04.3 LTS
Cemu 2.0-59 AppImage
Vulkan Renderer/Asyc Compile on/Accurate Barriers on
Vsync is off

Meepster-836:
I played some more of the game and found 2 more scenarios in which this bug occurs:

1. Specifically when opening the Yes/No prompt for upgrading an Art

2. When going to the _Shop_ menu for Skells and browsing purchasable equipment

MNolan147:
I’m also having this issue (flickering in the equipment select screen). I’ve tried changing graphics pack settings (change resolution to default, 30fps, revert to default) and nothing has worked so far.

I’m using:

AMD Ryzen 7 7800X3D
Radeon RX 7800 XT
64GB DDR4 RAM
Arch Linux
Cemu built from GitHub
Vulcan Renderer/Async Compile On/Accurate Barriers On
V-sync Triple Buffering (issue still occurs with V-sync off)

This could be an issue with the graphics card as we are both using the same card.

AKKypher:
I am having the same issue "flickering in the equipment select screen" I have also tried changing graphics pack settings and more nothing has worked. 

[...]

PR #1378, opened 2024-10-13
Vulkan: Protect against uniform var ringbuffer overflow

goeiecool9999:
Check if the uniform ringbuffer is full and if it is wait for commands to finish.
In my testing Breath of the Wild can run correctly with a ring buffer as small as 10 KB (with terrible performance obviously). When using a buffer smaller than that debug builds can detect corruption and assert/log. Although it's definitely impossible above a certain ringbuffer size.

### DIFF
--- a/src/Cafe/HW/Latte/Renderer/Vulkan/VulkanRendererCore.cpp
+++ b/src/Cafe/HW/Latte/Renderer/Vulkan/VulkanRendererCore.cpp
@@ -1390,17 +1390,6 @@ void VulkanRenderer::draw_execute(uint32 baseVertex, uint32 baseInstance, uint32
 		return;
 	}
 
-	// process index data
-	const LattePrimitiveMode primitiveMode = static_cast<LattePrimitiveMode>(LatteGPUState.contextRegister[mmVGT_PRIMITIVE_TYPE]);
-
-	Renderer::INDEX_TYPE hostIndexType;
-	uint32 hostIndexCount;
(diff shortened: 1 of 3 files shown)


Exzap:
Looks good👍Will do some testing later today and after that give you the final ok to merge

This is an actual fix for #1040 and #1011. I suspect this will also fix a lot of mystery issues in regards to random graphics corruptions.

src/Cafe/HW/Latte/Renderer/Vulkan/VulkanRendererCore.cpp
```diff
@@ -455,26 +451,63 @@ void VulkanRenderer::uniformData_updateUniformVars(uint32 shaderStageIndex, Latt
...
+				{
+					if(m_cmdBufferUniformRingbufIndices[m_commandBufferIndex] != m_uniformVarBufferReadIndex)
+					{
+						draw_endRenderPass();
+						SubmitCommandBuffer();
+					}
+					else
+					{
+						// submitting work would not change readIndex, so there's no way for conditions based on it to change
+						cemuLog_logDebug(LogType::Force, "draw call overflowed and corrupted uniform ringbuffer. expect visual corruption");
```
Exzap, line 470: Maybe use `cemuLog_logOnce` here so that we can see in user logs if stalls are encountered but at the same time we don't want to spam the log file

Exzap:
All ok during testing. You can merge this when ready. Thanks for looking into this

Issue #1373, opened 2024-10-07
Add checkbox to toggle screen for controller


damil42:
### Your suggestion

Currently you can define a button to show the gamepad screen. It would be great if we could add a checkbox to toggle the button instead of holding it down. That means you only have to press once and the other screen will be shown permanently. Press again to go back. Just like CTRL+TAB on the keyboard. Just for the controller. I want to play with the controller and quickly switch screens without interrupting myself by using the keyboard.

![{205C94CE-D38E-4BA5-AEA0-5C2B8AC29FFB}](https://github.com/user-attachments/assets/f619faee-ca4c-4dea-9de0-a0bb7cfe1ad4)

NazaDwnt3d:
I 100% agree that this feature should be added, it would make Cemu much less of a hassle to use with just a controller and single monitor

Issue #916, opened 2023-07-27
[Feature request] Option to make the "show screen" bind a toggle


DiaBomb:
Currently, the show screen button is only a press and hold. This would be useful for games that show gameplay relevant things only on the gamepad, such as both Zelda HD versions if using a gamepad. I know you can toggle already using ctrl+tab, but as far as I know, that can't be bound to a controller button without the use of third party software.

PR #1383, opened 2024-10-17
Add toggle button to VPAD show screen input

goeiecool9999:
In addition to adding a checkbox this also alters the behaviour of pressing the "show screen" button. Currently if you enable the "Launch with gamepad view" option in the game profile, pressing the button does nothing. I've changed it so it acts like the TAB key and shows the TV screen.
image
image

I'm not happy with the alignment of the volume slider and the checkbox. That's the main reason why this is marked as draft. I think it needs to look a little more polished.
EDIT: I've added a windows screenshot since writing this and I think it doesn't look as bad on windows.
Closes: #1373
Closes: #916

### DIFF
--- a/src/Cafe/CafeSystem.cpp
+++ b/src/Cafe/CafeSystem.cpp
@@ -396,7 +396,7 @@ void cemu_initForGame()
 		// replace any known function signatures with our HLE implementations and patch bugs in the games
 		GamePatch_scan();
 	}
-	LatteGPUState.alwaysDisplayDRC = ActiveSettings::DisplayDRCEnabled();
+	LatteGPUState.isDRCPrimary = ActiveSettings::DisplayDRCEnabled();
 	InfoLog_PrintActiveSettings();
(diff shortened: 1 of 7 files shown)


goeiecool9999:
I figured out how to align the text so the PR is ready now.
image

Exzap:
I don't think it's clear to the user what `toggle screen` actually means in this context and that it relates to the show screen button. You should at the very least add a tooltip that explains what it does, but maybe you can think of a better solution that is more intuitive. I was never super happy with `show screen` either, but there is not enough space for `show GamePad screen`.

Also, put the label on the left and then have the text-less checkbox on the right of it. That way it will visually line up with the above options. I think that makes more sense than having it line up with the volume slider

goeiecool9999:
> but maybe you can think of a better solution that is more intuitive.

I tried making a box around it with wxStaticBox but I couldn't figure out how to make it look good :shrug:
This is what it looks like now:
image

Issue #869, opened 2023-06-18
DSUController settings crash Cemu


e00E:
Arch Linux, Cemu built through AUR

> cemu --version
> 2.0-353 (experimental)

Playing around with adding and removing DSUController inputs, I quickly see Cemu crashing. I'm not posting the stacktrace because it is useless due to missing debug symbols. The log doesn't contain relevant information either.

Crementif:
Can you try building Cemu from source and try wayland? The aur package isn't provided by us thus isn't officially supported.

e00E:
I can reproduce it with the official AppImage version. The stacktrace is still useless.

PR #1386, opened 2024-10-19
input: Fix crash when closing add controller dialog before search completes

goeiecool9999:
Fixes: #869

### DIFF
--- a/src/gui/input/InputAPIAddWindow.cpp
+++ b/src/gui/input/InputAPIAddWindow.cpp
@@ -212,7 +212,7 @@ void InputAPIAddWindow::on_api_selected(wxCommandEvent& event)
 
 void InputAPIAddWindow::on_controller_dropdown(wxCommandEvent& event)
 {
-	if (m_search_running)
+	if (m_search_thread.joinable())
 		return;
 
 	int selection = m_input_api->GetSelection();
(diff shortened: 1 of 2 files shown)


src/gui/input/InputAPIAddWindow.h
```diff
@@ -50,4 +51,10 @@ class InputAPIAddWindow : public wxDialog
 
 	std::vector<ControllerPtr> m_controllers;
 	std::atomic_bool m_search_running = false;
+	struct AsyncThreadData
+	{
+		std::atomic_bool discardResult = false;
+		std::mutex mutex;
```
Exzap, line 57: Suggestion: Make the mutex a member variable of `InputAPIAddWindow`. Tracking it per-thread only unnecessarily complicates things and is error prone.

src/gui/input/InputAPIAddWindow.cpp
```diff
@@ -162,6 +168,10 @@ void InputAPIAddWindow::on_api_selected(wxCommandEvent& event)
 	if (m_input_api->GetSelection() == wxNOT_FOUND)
 		return;
 
+	m_search_running = false;
+	if(m_search_thread_data)
+		m_search_thread_data->discardResult = true;
```
Exzap, line 173: You should acquire the mutex here just for consistency

src/gui/input/InputAPIAddWindow.cpp
```diff
@@ -114,6 +114,12 @@ InputAPIAddWindow::InputAPIAddWindow(wxWindow* parent, const wxPoint& position,
 	this->Bind(wxControllersRefreshed, &InputAPIAddWindow::on_controllers_refreshed, this);
 }
 
+InputAPIAddWindow::~InputAPIAddWindow()
+{
+	std::lock_guard lock{m_search_thread_data->mutex};
```
Exzap, line 119: `m_search_thread_data` can be null here if `on_add_button` wasn't called yet. This can be triggered when opening the window and just closing it again

Exzap (changes requested):

src/gui/input/InputAPIAddWindow.h
Exzap: Actually, this may then introduce a race condition where `this` is destroyed before the thread accesses it. Not sure, maybe there is a less fragile way to handle this in general.

goeiecool9999: > maybe there is a less fragile way to handle this in general.

Have a look at the latest commit. I switched to a jthread. I think I was reinventing the wheel.

goeiecool9999: I've reverted it. Not only because of the lack of macOS support but also because it has more theoretical race conditions than the mutex.

PR #1392, opened 2024-10-25
Vulkan: Make scaling shaders compatible + fixes

goeiecool9999:
Make built-in scaling shaders compatible with vulkan and incorporate the fixes from #879

### Changes for graphic pack developers
- Cemu now prepends a preamble containing all the uniforms so you can just write the shader parts :smile: 
- preamble uses GLSL version 430
- ``textureSrcResolution`` is now the actual resolution of the textureSrc sampler.
- ``inputResolution`` has been replaced by ``nativeResolution`` and gives the resolution that the game would render at without graphic pack overrides.

#### Full preamble (as of writing) for reference
https://github.com/cemu-project/Cemu/blob/e2e9e81c3eaa626b401a9a61f10daf7335f95d7f/src/Cafe/HW/Latte/Renderer/RendererOuputShader.cpp#L287-L302

### DIFF
--- a/src/Cafe/GraphicPack/GraphicPack2.cpp
+++ b/src/Cafe/GraphicPack/GraphicPack2.cpp
@@ -960,7 +960,7 @@ bool GraphicPack2::Activate()
 				auto option_upscale = rules.FindOption("upscaleMagFilter");
 				if(option_upscale && boost::iequals(*option_upscale, "NearestNeighbor"))
 					m_output_settings.upscale_filter = LatteTextureView::MagFilter::kNearestNeighbor;
-				auto option_downscale = rules.FindOption("NearestNeighbor");
+				auto option_downscale = rules.FindOption("downscaleMagFilter");
 				if (option_downscale && boost::iequals(*option_downscale, "NearestNeighbor"))
 					m_output_settings.downscale_filter = LatteTextureView::MagFilter::kNearestNeighbor;
 			}
--- a/src/Cafe/GraphicPack/GraphicPack2.cpp
+++ b/src/Cafe/GraphicPack/GraphicPack2.cpp
@@ -960,7 +960,7 @@ bool GraphicPack2::Activate()
 				auto option_upscale = rules.FindOption("upscaleMagFilter");
 				if(option_upscale && boost::iequals(*option_upscale, "NearestNeighbor"))
 					m_output_settings.upscale_filter = LatteTextureView::MagFilter::kNearestNeighbor;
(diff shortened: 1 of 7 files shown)


Ammar-Sadaoui:
a y plan to support spir-v directly ?

src/Cafe/GraphicPack/GraphicPack2.cpp
```diff
@@ -960,7 +960,7 @@ bool GraphicPack2::Activate()
 				auto option_upscale = rules.FindOption("upscaleMagFilter");
 				if(option_upscale && boost::iequals(*option_upscale, "NearestNeighbor"))
 					m_output_settings.upscale_filter = LatteTextureView::MagFilter::kNearestNeighbor;
-				auto option_downscale = rules.FindOption("NearestNeighbor");
+				auto option_downscale = rules.FindOption("downscaleMagFilter");
```
Exzap, line 963: This should be named "downscaleMinFilter" to match [OpenGL/Vulkan terminology](https://gdbooks.gitbooks.io/legacyopengl/content/Chapter7/MinMag.html). Internally the code is just abusing `LatteTextureView::MagFilter` but it should probably be its own enum for Graphic Packs.

Exzap:
After merging this some users might be confused about why their image quality suddenly changed. So I wonder if we should force-reset to bilinear as part of this PR? It would unfortunately also reset the scaling options on OpenGL but that might be the lesser evil.

src/Cafe/GraphicPack/GraphicPack2.cpp
goeiecool9999: Ah okay. The reason I went with downscaleMagFilter is because that's what I saw in all of the community graphic packs.

goeiecool9999:
> some users might be confused about why their image quality suddenly changed

Considering people already convinced themselves that hermite looks better than bilinear when it wasn't even implemented I don't think it's that big of a deal. You have to really pixel peep to tell the difference. I think it would probably be equally or slightly more confusing if a user remembered their settings and saw that they changed on their own.
If you do want to force bilinear after updating how would we go about that? Does Cemu have a mechanism to detect the first launch after updating?

Exzap:
Did some more comparisons and yeah it's probably fine. Hermite starts looking bad for really large upscale factors, but thats beyond what anyone would reasonably run into.

Adamillo:
There's a regression now with the Intel GPU drivers on Windows where it doesn't render anything correctly anymore in OpenGL after this PR got merged unfortunately

Issue #1305, opened 2024-08-20
Unsupported texture readback format 000a


Triticum0:
### Current Behavior

Cause graphical graphical issues in The Legend of Zelda: Phantom Hourglass

### Expected Behavior

Support texture readback format 000a

### Steps to Reproduce

1. boot game
2. what to get to opening cinematic

### System Info (Optional)

OS: Windows 10
GPU: 3060 ti

JaimieVandenbergh:
Cemu is a Wii U emulator. VC DS games are partly supported. You'd be better off extracting the rom from the VC game bundle and playing it in a DS emulator. https://wiki.hacks.guide/wiki/Wii_U:VC_Extract

sharkeylaser:
That may be true, @JaimieVandenbergh , but isn't the point of emulator development to mimic the functionalities and capabilities of the original system? Obviously something isn't entirely accurate if these games can be played on real hardware but not via the emulator. To me, this is a valid bug report.

Issue #1296, opened 2024-08-17
Unsupported texture readback format 0008


Triticum0:
### Current Behavior

The game Harvest Moon 64 exhibits this behavior when trying to render text leaving the text to flicker and break

### Expected Behavior

not break

### Steps to Reproduce

1.run game 

### System Info (Optional)

OS: Windows 10
GPU: 3060 ti

PR #1430, opened 2024-11-25
Readback Vulkan fixes

rcaridade145:
Tentative fixes, i could not verify as i don't own the games, for #1296 and #1305 .
Can you please test @Triticum0

### DIFF
--- a/src/Cafe/HW/Latte/Renderer/Vulkan/TextureReadbackVk.cpp
+++ b/src/Cafe/HW/Latte/Renderer/Vulkan/TextureReadbackVk.cpp
@@ -81,13 +81,10 @@ uint32 LatteTextureReadbackInfoVk::GetImageSize(LatteTextureView* textureView)
 	}
 	else if (textureView->format == Latte::E_GX2SURFFMT::R5_G6_B5_UNORM )
 	{
-		cemu_assert(textureFormat == VK_FORMAT_R8G8B8A8_UNORM);
-		return baseTexture->width * baseTexture->height * 2;
-	}
-	else if (textureView->format == Latte::E_GX2SURFFMT::R5_G5_B5_A1_UNORM )
-	{
-		cemu_assert(textureFormat == VK_FORMAT_R8G8B8A8_UNORM);
-		return baseTexture->width * baseTexture->height * 4;
+		if(textureFormat == VK_FORMAT_R5G6B5_UNORM_PACK16){
+			return baseTexture->width * baseTexture->height * 2;
+		}	
+		return 0;
 	}
 	else
 	{
--- a/src/Cafe/HW/Latte/Renderer/Vulkan/TextureReadbackVk.cpp
+++ b/src/Cafe/HW/Latte/Renderer/Vulkan/TextureReadbackVk.cpp
@@ -82,7 +82,7 @@ uint32 LatteTextureReadbackInfoVk::GetImageSize(LatteTextureView* textureView)
 	else if (textureView->format == Latte::E_GX2SURFFMT::R5_G6_B5_UNORM )
 	{
 		cemu_assert(textureFormat == VK_FORMAT_R8G8B8A8_UNORM);
-		return baseTexture->width * baseTexture->height * 4;
+		return baseTexture->width * baseTexture->height * 2;
 	}
 	else if (textureView->format == Latte::E_GX2SURFFMT::R5_G5_B5_A1_UNORM )
 	{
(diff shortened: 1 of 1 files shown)


Triticum0:
can I have a compiled build to test it

src/Cafe/HW/Latte/Renderer/Vulkan/TextureReadbackVk.cpp
```diff
@@ -79,6 +79,16 @@ uint32 LatteTextureReadbackInfoVk::GetImageSize(LatteTextureView* textureView)
 		// todo - if driver does not support VK_FORMAT_D24_UNORM_S8_UINT this is represented as VK_FORMAT_D32_SFLOAT_S8_UINT which is 8 bytes
 		return baseTexture->width * baseTexture->height * 4;
 	}
+	else if (textureView->format == Latte::E_GX2SURFFMT::R5_G6_B5_UNORM )
+	{
+		cemu_assert(textureFormat == VK_FORMAT_R8G8B8A8_UNORM);
```
Exzap, line 84: `Latte::E_GX2SURFFMT::R5_G6_B5_UNORM` is a bit tricky, because it can be emulated as either `VK_FORMAT_R5G6B5_UNORM_PACK16` or if that is not supported by the driver the Vulkan backend will fallback to `VK_FORMAT_R8G8B8A8_UNORM`. See the code [here](https://github.com/cemu-project/Cemu/blob/90eb2e01f405e502ed33fea330352e167c4a14db/src/Cafe/HW/Latte/Renderer/Vulkan/VulkanRenderer.cpp#L2372). The readback code needs to handle this correctly. For non-matching formats conversion is necessary. If you want to keep it simple you could ignore the more complicated case and return 0 when `textureFormat != VK_FORMAT_R5G6B5_UNORM_PACK16`

src/Cafe/HW/Latte/Renderer/Vulkan/TextureReadbackVk.cpp
```diff
@@ -79,6 +79,16 @@ uint32 LatteTextureReadbackInfoVk::GetImageSize(LatteTextureView* textureView)
...
+		return baseTexture->width * baseTexture->height * 4;
```
Exzap, line 85:
```suggestion
		return baseTexture->width * baseTexture->height * 2;
```
`Latte::E_GX2SURFFMT::R5_G6_B5_UNORM` is 2 bytes per texel

src/Cafe/HW/Latte/Renderer/Vulkan/TextureReadbackVk.cpp
```diff
@@ -79,6 +79,16 @@ uint32 LatteTextureReadbackInfoVk::GetImageSize(LatteTextureView* textureView)
...
+	}
+	else if (textureView->format == Latte::E_GX2SURFFMT::R5_G5_B5_A1_UNORM )
```
Exzap, line 87: For `R5_G5_B5_A1_UNORM` conversion is always necessary since the backend doesn't use Vulkan formats that are bit-compatible. See code [here](https://github.com/cemu-project/Cemu/blob/90eb2e01f405e502ed33fea330352e167c4a14db/src/Cafe/HW/Latte/Renderer/Vulkan/VulkanRenderer.cpp#L2383). Maybe just drop this case for now?

Exzap (changes requested):
I did tell you on discord that most readback formats follow the same pattern, but you happened to pick the formats that need specialized handling. Apologizes for putting you on the wrong track

src/Cafe/HW/Latte/Renderer/Vulkan/TextureReadbackVk.cpp
rcaridade145: Removed

Triticum0:
Can I have a test build? It has still not been approved to be able to test the fix.

rcaridade145:
> Can I have a test build? It has still not been approved to test the fix.

https://github.com/cemu-project/Cemu/actions/runs/12012239663/artifacts/2238726253 Does this work for you?

Triticum0:
I tested all of them except The Legend of Zelda: Phantom Hourglass seems version 2.4 has a regression meaning their graphical corruption and uses 5gb whereas 2.3 uses half of that. seems to happen on both DS games I tested so may affect all the DS games. So can't test the pr to see if it works. tested Harvest Moon 64 and seemed to render the same but didn't check the log for any difference if it still got hit or issue unrelated.

PR #1431, opened 2024-11-26
Set version for macOS bundle

neebyA:
On macOS, the version of Cemu is always reported as `2.0.0`, regardless of the version of the actual release. For example, this what the app version is reported as despite being from the 2.4 pre-release:
```
grep -A 1 'CFBundleVersion' /Applications/Cemu.app/Contents/Info.plist
    <key>CFBundleVersion</key>
    <string>2.0.0</string>
```

This is due to cmake using the variable `CMAKE_PROJECT_VERSION` to set the version of the app bundle, rather than the values of the environment variables passed to the GitHub workflow.

This PR sets the version for the macOS app bundle using the workflow environment variable, instead of the hardcoded `CMAKE_PROJECT_VERSION` of 2.0.0.


Exzap:
thanks!

PR #1445, opened 2024-12-09
Check whether portable is directory to avoid conflict with portable sandboxing framework

### LINKED ISSUE #1444: [LINUX] cemu conflicts with portable sandboxing framework software, mistakenly thinking that it is a portable configuration
### Current Behavior

From src/gui/CemuApp.cpp, if "portable" is found under the same directory of cemu app, cemu will think it is a portable configuration. However, https://github.com/Kraftland/portable provides a file called "portable".  This will become a problem when cemu and portable are managed by linux repos, where /usr/bin/cemu and /usr/bin/portable are both present. Cemu will mistakenly think it is under portable configuration and refuse to launch.

### Expected Behavior

Check whether "portable" is a directory

### Steps to Reproduce

E.g. in archlinux, install cemu and portable (https://aur.archlinux.org/packages/portable, https://aur.archlinux.org/packages/cemu) together and cemu will refuse to launch

### System Info (Optional)

OS: Archlinux

_No response_

### Logs (Optional)

_No response_

heavysink:
Portable sandboxing framework adds an executable called "portable" which may confuse the portable configuration check of cemu. Fix #1444

### DIFF
--- a/src/gui/CemuApp.cpp
+++ b/src/gui/CemuApp.cpp
@@ -88,7 +88,7 @@ void CemuApp::DeterminePaths(std::set<fs::path>& failedWriteAccess) // for Windo
 	fs::path exePath(wxHelper::MakeFSPath(standardPaths.GetExecutablePath()));
 	fs::path portablePath = exePath.parent_path() / "portable";
 	data_path = exePath.parent_path(); // the data path is always the same as the exe path
-	if (fs::exists(portablePath, ec))
+	if (fs::is_directory(portablePath, ec))
 	{
 		isPortable = true;
 		user_data_path = config_path = cache_path = portablePath;


goeiecool9999:
> a more resilient solution would be to add a build flag

Which cemu used to have. 9bbb7c8b deliberately removed the build flag and let Cemu determine if it should run portably at runtime. I personally prefer this. Maybe a reasonable compromise would be to add an overriding command line argument?

goeiecool9999:
> Determining whether to run portably at runtime can still be done, the flag can be used to prevent Cemu from being allowed to decide at runtime.

Sounds good to me. I figured that's probably what you meant after I already posted my reply.

Reality2byte (approved):

Exzap:
Superseded by #1464

PR #1468, opened 2025-01-07
Updated dependencies for fedora, added troubleshooting flag for alternative architectures

brysma1:
Changed bluez-libs to bluez-libs-devel and added wayland-protocols-devel on fedora dependencies command.
Added troubleshooting step for alternative architectures as required by vcpkg, found when building on a M1 macbook air running fedora asahi remix.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -57,7 +57,7 @@ At Step 3 in [Build Cemu using cmake and clang](#build-cemu-using-cmake-and-clan
    `cmake -S . -B build -DCMAKE_BUILD_TYPE=release -DCMAKE_C_COMPILER=/usr/bin/clang-15 -DCMAKE_CXX_COMPILER=/usr/bin/clang++-15 -G Ninja -DCMAKE_MAKE_PROGRAM=/usr/bin/ninja`
 
 #### For Fedora and derivatives:
-`sudo dnf install bluez-libs clang cmake cubeb-devel freeglut-devel git glm-devel gtk3-devel kernel-headers libgcrypt-devel libsecret-devel libtool libusb1-devel llvm nasm ninja-build perl-core systemd-devel zlib-devel zlib-static`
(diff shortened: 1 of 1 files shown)


Liubik (changes requested):
Typo in "* Environment variable `VCPKG_FORCE_SYSTEM_BINARIES` must **ve** set."
It should be "* Environment variable `VCPKG_FORCE_SYSTEM_BINARIES` must **be** set."

Exzap:
Passing `-DVCPKG_FORCE_SYSTEM_BINARIES=1` to cmake does not work since it's an environment variable and not a cmake variable. The easiest way to handle it is by prefixing the cmake command like this:
`VCPKG_FORCE_SYSTEM_BINARIES=1 cmake -S . -B build ...`

BUILD.md
```diff
@@ -120,6 +120,9 @@ This section refers to running `cmake -S...` (truncated).
 * Compiling failed during rebuild after `git pull` with an error that mentions RPATH
    * Add the following and try running the command again:
       * `-DCMAKE_BUILD_WITH_INSTALL_RPATH=ON`
+* Environment variable `VCPKG_FORCE_SYSTEM_BINARIES` must be set.
+   * Execute the folowing and then try running the command again:
+      * `VCPKG_FORCE_SYSTEM_BINARIES=1`
```
Exzap, line 125:
```suggestion
      * `export VCPKG_FORCE_SYSTEM_BINARIES=1`
```

Exzap:
Had to tweak it a bit but its merged now, thanks! Just FYI you can either execute `export VCPKG_FORCE_SYSTEM_BINARIES=1` or use the prefixing method I explained. But its not enough to just run `VCPKG_FORCE_SYSTEM_BINARIES=1` as a single command.

brysma1:
I tried it both in zsh and sh and it does work on both, but i guess the export is more specific, either way, thank you and sorry for the inconvenience for such a simple PR

PR #1471, opened 2025-01-11
nsyshid: Make Libusb the Windows backend

deReeperJosh:
After doing some extensive testing, I am opening this PR to make Libusb the default backed across all platforms, including now Windows. 

The motivations for this being, that other emulators also use Libusb as their default backend on all platforms, which on Windows requires the device to have libusb drivers installed. However, if users were jumping between emulators and playing games that required a usb peripheral, they would have to constantly uninstall and reinstall different drivers for their device.

Also, having one backend for all operating systems allows for any changes made to be applied to all OSs, which should make future updates to the nsyshid methods simpler.

Libusb on windows also fixes the Skylanders: Trap Team portal audio issues seen with Windows HID.

Updates to nsyshid methods and method parameters have been made based on updates to the WUT nyshid header file. Updates to the libusb backend (including the addition of set protocol and set idle) have been made based on the updates to the nsyshid parameters, and the information in section 7 of the [HID 1.1 specifications](https://www.usb.org/sites/default/files/hid1_11.pdf)


src/Cafe/OS/libs/nsyshid/nsyshid.cpp
```diff
@@ -305,75 +305,191 @@ namespace nsyshid
...
+			DoHIDTransferCallback(callbackFuncMPTR,
+								  callbackParamMPTR,
+								  device->m_hid->handle,
+								  -1,
+								  0,
+								  0);
+		}
+	}
+
 	void export_HIDGetDescriptor(PPCInterpreter_t* hCPU)
```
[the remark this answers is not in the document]
Exzap: It's easy to accidentally introduce subtle regressions while reworking HLE interfaces so I think this should ideally be done as a separate PR anyway.

src/Cafe/OS/libs/nsyshid/BackendLibusb.cpp
```diff
@@ -681,24 +690,63 @@ namespace nsyshid::backend::libusb
...
+
+		uint16 wValue = uint16(duration) << 8 | uint16(reportId);
+
+		// HID Set_Idle requests are handled via libusb_control_transfer
+		int ret = libusb_control_transfer(handleLock->GetHandle(),
+										  LIBUSB_REQUEST_TYPE_CLASS | LIBUSB_RECIPIENT_INTERFACE | LIBUSB_ENDPOINT_OUT,
+										  HID_CLASS_SET_IDLE, // Defined in HID Class Specific Requests (7.2)
+										  wValue,
+										  ifIndex,
+										  emptyData.data(),
```
Exzap, line 740: Can you not pass nullptr here instead of creating an empty array?

Exzap:
Looks good to me. Regarding switching a few instances from `cemuLog_logDebug` to `cemuLog_log` in the libusb backend, we can leave that in for a while if its helpful for troubleshooting, but generally systems should not write to log in public builds for anything that is within normal operation.

src/Cafe/OS/libs/nsyshid/BackendLibusb.cpp
deReeperJosh: Yes you can - this was a leftover of me debugging bits locally, have now removed

deReeperJosh:
I have restored the debug logs in the libusb backend - had changed them to normal logs while testing locally

PR #1478, opened 2025-01-21
nsyshid: Play Emulated Portal Audio via Mono Audio

deReeperJosh:
In the Skylanders games, the portal that is currently emulated is the Trap Team portal. Normally it writes requests via the SetReport method, but the trap team portal also accepts audio requests via the Write method. The portal attached to the Trap Team portal is a Mono audio, 8000hz speaker, and audio requests are 64 bytes, which is 32 16 bit PCM audio samples.

I have also made the portal audio a setting similar to gamepad/tv audio, as well as changing some methods to allow for more than a binary choice when playing audio output.

### DIFF
--- a/src/audio/IAudioAPI.cpp
+++ b/src/audio/IAudioAPI.cpp
@@ -20,7 +20,7 @@ uint32 IAudioAPI::s_audioDelay = 2;
 std::array<bool, IAudioAPI::AudioAPIEnd> IAudioAPI::s_availableApis{};
 
 IAudioAPI::IAudioAPI(uint32 samplerate, uint32 channels, uint32 samples_per_block, uint32 bits_per_sample)
-	: m_samplerate(samplerate), m_channels(channels), m_samplesPerBlock(samples_per_block), m_bitsPerSample(bits_per_sample) 
+	: m_samplerate(samplerate), m_channels(channels), m_samplesPerBlock(samples_per_block), m_bitsPerSample(bits_per_sample)
 {
 	m_bytesPerBlock = samples_per_block * channels * (bits_per_sample / 8);
 	InitWFX(m_samplerate, m_channels, m_bitsPerSample);
@@ -81,7 +81,7 @@ void IAudioAPI::InitializeStatic()
 #if BOOST_OS_WINDOWS
 	s_availableApis[DirectSound] = true;
 	s_availableApis[XAudio2] = XAudio2API::InitializeStatic();
-	if(!s_availableApis[XAudio2]) // don't try to initialize the older lib if the newer version is available
+	if (!s_availableApis[XAudio2]) // don't try to initialize the older lib if the newer version is available
 		s_availableApis[XAudio27] = XAudio27API::InitializeStatic();
 #endif
 #if HAS_CUBEB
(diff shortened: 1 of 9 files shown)


src/config/CemuConfig.cpp
```diff
@@ -508,12 +520,15 @@ void CemuConfig::Save(XMLConfigParser& parser)
 	audio.set("TVChannels", tv_channels);
 	audio.set("PadChannels", pad_channels);
 	audio.set("InputChannels", input_channels);
+	audio.set("InputChannels", portal_channels);
```
Exzap, line 523: Should be `"PortalChannels"`

deReeperJosh: good spot, thanks

src/audio/IAudioAPI.cpp
```diff
@@ -97,30 +98,29 @@ bool IAudioAPI::IsAudioAPIAvailable(AudioAPI api)
...
+AudioAPIPtr IAudioAPI::CreateDeviceFromConfig(AudioType type, sint32 rate, sint32 samples_per_block, sint32 bits_per_sample)
 {
-	auto& config = GetConfig();
-	sint32 channels = CemuConfig::AudioChannelsToNChannels(TV ? config.tv_channels : config.pad_channels);
+	sint32 channels = CemuConfig::AudioChannelsToNChannels(AudioTypeToChannels(type));
 	return CreateDeviceFromConfig(TV, rate, channels, samples_per_block, bits_per_sample);
 }
 
-AudioAPIPtr IAudioAPI::CreateDeviceFromConfig(bool TV, sint32 rate, sint32 channels, sint32 samples_per_block, sint32 bits_per_sample)
+AudioAPIPtr IAudioAPI::CreateDeviceFromConfig(AudioType type, sint32 rate, sint32 channels, sint32 samples_per_block, sint32 bits_per_sample)
```
goeiecool9999, line 107: GH won't let me comment on the actual line of code but near the end of this function it sets the volune of the created audio API. Currently still uses ternary statement to pick between pad and tv volume values. Should be set to portal audio volume instead.

src/config/CemuConfig.h
```diff
@@ -479,9 +479,9 @@ struct CemuConfig
 	// audio
 	sint32 audio_api = 0;
 	sint32 audio_delay = 2;
-	AudioChannels tv_channels = kStereo, pad_channels = kStereo, input_channels = kMono;
-	sint32 tv_volume = 50, pad_volume = 0, input_volume = 50;
-	std::wstring tv_device{ L"default" }, pad_device, input_device;
+	AudioChannels tv_channels = kStereo, pad_channels = kStereo, input_channels = kMono, portal_channels = kMono;
```
goeiecool9999, line 482: Should the portal have a setting to adjust the number of channels? The portal always plays mono sound right? This value controls the amount of channels the source is able to have, not the amount of channels on which playback happens.

goeiecool9999 (changes requested):

src/config/CemuConfig.h
deReeperJosh: Yeah this is a good point - I will remove this as a setting

deReeperJosh:
Discovered a strange bug where every re-launch of Cemu would require the settings window to be opened at least once to play audio, will re-publish when I have found a fix

deReeperJosh:
Found the issue with Audio at startup, there was still a reference to a ternery operator, and the AudioType wasn't being passed through correctly in IAudioAPI

Exzap:
Looks good. Thanks!

Jack54guythecoder:
I know this is CEMU, but is there a chance this would be implemented in RPCS3?

PR #1488, opened 2025-02-05
SoundAPI: OpenAL backend

LAGonauta:
This PR implements an OpenAL backend to Cemu. I've tested it on Linux and on Windows, but other platforms need testing.

## Reasoning

Whenever surround sound is available, I like to enable headphone 3D audio spatialization, however I've been using Moonlight on Android and in my experience surround sound has extreme audio lag in that platform. Each input has a 1 second audio delay or so, which is quite bad. Stereo works wonderfully, though.

Through OpenAL I can enable 3D audio spatialization (HRTF) support in OpenAL Soft's config and transmit the audio in Stereo, ensuring low latency.

When playing locally, audio latency with OpenAL Soft is comparable to Cubeb, and I added a concept of "Target Latency" that automatically resamples the audio to maintain one to three blocks buffered at all times. Unlike XAudio2 which allows it to grow to 24 blocks and then start dropping them to maintain acceptable audio lag.

As a bonus, if you are one of the lucky few to have access to a sound card with hardware OpenAL support you should expect undetectable audio lag.

## Notes

Instead of downloading the OpenAL headers through vcpkg I included them in the branch. I can't say I know what is best. I can change it to vcpkg if you feel it is better to do it that way.

Additionally, in this change set I load the library dynamically to not require OpenAL to be instead in the user's system even when support is compiled in, it is an optional dependency. I think it is sane to do this as other backends are always available.

I am available for any questions and I will maintain this backend for the long term, if not here in my branch.


LuismaSP89:
I'm curious about this PR. One of my WiiU's favorite games (Zombi U) has audio issues because Cemu is capable only to play this game in stereo, and there's a lot of missing sounds ingame. 

In the original WiiU you need to change the output in the system for this game, from 2.0 to 5.1 in order to play it with all the sounds, but if I do this in the system (booting wiiU os in cemu) this makes no difference. Maybe this PR fix this, we will see : )

Exzap:
Can you help me understand how exactly OpenAL fixes your issue and why a solution can not be implemented on a lower level so that it works for the other backends. Asking because from what I understand, OpenAL (soft) nowadays is just done via translation to a lower audio API and I fail to understand how OpenAL can achieve something that those APIs can't. I am assuming you are not running a setup where there is actual hardware support for OpenAL, if thats even a thing anymore.

> I'm curious about this PR. One of my WiiU's favorite games (Zombi U) has audio issues because Cemu is capable only to play this game in stereo, and there's a lot of missing sounds ingame.

This PR doesn't fix it. The issue you describe is a problem in the HLE audio emulation core (sndcore) which is not touched by this PR.

PR #1494, opened 2025-02-10
gui: Add menu item for opening shader cache directory

capitalistspz:
Especially helpful for non-portable Linux users, who have `shaderCache` outside of the main Cemu directory


JaimieVandenbergh:
Good for Mac users too, since that's in a slightly surprising location (~/Library/Caches/Cemu/)

HimanshuBansal112:
I think that a way to open shader cache for each game could also be a good idea for this or future pull request.

PR #1496, opened 2025-02-13
Vulkan: Match Vulkan Device Version and Shaders Targets

Xphalnos:
Instead of using Vulkan 1.1 and Spir-V 1.3 targets by default, It detect the Vulkan Version of the device and and matches them to the glslang targets.
To see which Spir-V version matches with which Vulkan version, I've used `glslang --help` command.

glslang

PS: Vulkan 1.4 devices uses Vulkan 1.3 and Spir-V 1.6 targets.

Tested on: Zelda Breath of the Wild, Mario Kart 8, Splatoon, Zelda Wind Waker HD, Zelda Twilight Princess HD, Super Mario 3D World and Wii U Menu
Tested with a Vulkan 1.4 device and no issues are visible.
Ready for modifications!

### DIFF
--- a/src/Cafe/HW/Latte/Renderer/Vulkan/RendererShaderVk.cpp
+++ b/src/Cafe/HW/Latte/Renderer/Vulkan/RendererShaderVk.cpp
@@ -318,9 +318,25 @@ void RendererShaderVk::CompileInternal(bool isRenderThread)
 	const char* cstr = m_glslCode.c_str();
 	Shader.setStrings(&cstr, 1);
 	Shader.setEnvInput(glslang::EShSourceGlsl, state, glslang::EShClientVulkan, 100);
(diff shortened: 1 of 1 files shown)


amayra:
look forward to dropping GLSL 

and support spir-v directly

Exzap:
SPIR-V binaries get cached so its possible for someone to copy over the cache to a machine which has a lower Vulkan version, which then wouldn't be able to compile the shaders. So adjustments would have to be made to how caching works. I am also not super happy with how verbose the code is and I think a lookup table would be more compact. If there ever is a Vulkan 2.x version then this code would stop calling `Shader.setEnvClient`/`Shader.setEnvTarget` at all.

Newer SPIR-V features are mostly irrelevant to us and I prefer the simplicity of the original code. If there is an actual improvement to FPS then feel free to reopen.

PR #1514, opened 2025-03-15
input: timer shift factor hotkey

Anime37:
pressing space toggles timer shift factor between 1x and 4x.

poor man's fast-forward toggle hotkey.


Anime37:
might work on a hotkey config while i have free time

Anime37:
almost done with configurable hotkey impl, so closing this one

PR #1515, opened 2025-03-15
cubeb: Allow to select the default device even if cubeb_enumerate_devices fails

mitoposter:
Cemu would not let you select even cubeb's default device if cubeb_enumerate_devices failed,
however, there are cases where it will fail but you can still get sound if you use cubeb's default device.
(For example, if you use ALSA. )

### DIFF
--- a/src/audio/CubebAPI.cpp
+++ b/src/audio/CubebAPI.cpp
@@ -183,16 +183,16 @@ void CubebAPI::Destroy()
 
 std::vector<IAudioAPI::DeviceDescriptionPtr> CubebAPI::GetDevices()
 {
-	cubeb_device_collection devices;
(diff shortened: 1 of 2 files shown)


Squall-Leonhart:
The solution is to not use alsa, Mozilla doesn't support it beyond 2ch and certainly not on multiple output devices like the Wii Gamepad audio requires.

goeiecool9999 (approved):

mitoposter:
&gt;the solution is to not use alsa<br/>Show me on the doll where ALSA touched you.<br/>&gt;mozilla doesn't support it beyond&hellip;<br/>Yes, I could tell that the &ldquo;support&rdquo; is just the very bare minimum.

Squall-Leonhart:
> Yes, I could tell that the “support” is just the very bare minimum.

wait till https://github.com/mozilla/cubeb/issues/705 and kills alsa entirely.

PR #1519, opened 2025-03-19
Configurable hotkeys + fastforward hotkey as example

Anime37:
Implemented configurable hotkeys and tried to make it as easy as possible to add new hotkeys.  
Second commit is an example of how to add a new hotkey and it's also a pretty useful hotkey in my opinion.

im bad at design but it works:  
image  
image


Anime37:
fixed modifiers not saving to config + cut down hotkey values to 2bytes instead of 4bytes.

chriztr:
Question;
Could a hotkey for exiting cemu be added?
And can these be mapped to your controller?

Anime37:
> Question; Could a hotkey for exiting cemu be added? And can these be mapped to your controller?
if this gets approved, either me or anyone else could add a hotkey for exiting the emulator.
currently this is only capturing keyboard events, so can't bind controller buttons. that's the next thing i'll try to implement.

src/gui/input/HotkeySettings.cpp
```diff
@@ -0,0 +1,148 @@
...
+extern WindowInfo g_window_info;
+
+static auto& cfg_hotkeys = GetConfig().hotkeys;
+static std::unordered_map<uHotkey*, std::function<void(wxFrame&)>> cfg_hotkey_to_func_map{
+	{&cfg_hotkeys.toggle_fullscreen, [](wxFrame& frame) {frame.ShowFullScreen(!frame.IsFullScreen());}},
+	{&cfg_hotkeys.toggle_fullscreen_alt, [](wxFrame& frame) {frame.ShowFullScreen(!frame.IsFullScreen());}},
+	{&cfg_hotkeys.exit_fullscreen, [](wxFrame& frame) {frame.ShowFullScreen(false);}},
+	{&cfg_hotkeys.take_screenshot, [](wxFrame&) {g_window_info.has_screenshot_request = true;}},
+	{&cfg_hotkeys.toggle_fastforward, [](wxFrame& frame) {ActiveSettings::SetTimerShiftFactor((ActiveSettings::GetTimerShiftFactor() < 3) ? 3 : 1);}},
+};
```
goeiecool9999, line 15: It seems arbitrary to have the MainWindow frame passed as an argument for all hotkey functions when not all of them use it. Could you call the frame functions via some other means?

src/gui/input/HotkeySettings.h
```diff
@@ -0,0 +1,32 @@
...
+
+	HotkeySettings(wxWindow* parent);
+	~HotkeySettings();
+
+  private:
+	wxPanel* m_panel;
+	wxFlexGridSizer* m_sizer;
+	std::vector<HotkeyEntry> m_hotkeys;
+	std::unordered_map<wxButton*, uHotkey*> m_input_to_cfg_map;
+	bool need_to_save = false;
```
goeiecool9999, line 22: General note: Please conform to the [style guidelines](https://github.com/cemu-project/Cemu/blob/main/CODING_STYLE.md#names-for-variables-functions-and-classes). Admittedly some of Cemu's code already strays from it, but at the very least need_to_save should have an ``m_`` prefix for consistency. I don't know how strictly we should enforce the style guidelines, but I think it's best to adhere to it for new code. I'll defer to @Exzap on that.

src/gui/input/HotkeySettings.cpp
```diff
@@ -0,0 +1,148 @@
...
+		new_hotkey.ctrl = event.ControlDown();
+		new_hotkey.shift = event.ShiftDown();
+		if (new_hotkey.raw != old_hotkey.raw)
+		{
+			auto& cfg_hotkey = *m_input_to_cfg_map[obj];
+			need_to_save |= (new_hotkey.raw != old_hotkey.raw);
+			cfg_hotkey = new_hotkey;
+			hotkey_to_func_map.erase(old_hotkey.raw);
+			hotkey_to_func_map[cfg_hotkey.raw] = cfg_hotkey_to_func_map[&cfg_hotkey];
+		}
```
goeiecool9999, line 108: m_input_to_cfg_map can be eliminated. wxButton inherits from wxEvtHandler which means you can call SetClientData() to store any arbitrary pointer, which can later be retrieved with GetClientData(). As wxWidgets' documentation points out ``Normally, any extra data the programmer wishes to associate with the object should be made available by deriving a new class with new data members.`` but I think in this case defining a whole new button class is more trouble than it's worth.

src/gui/input/HotkeySettings.cpp
```diff
@@ -0,0 +1,148 @@
...
+		new_hotkey.alt = event.AltDown();
+		new_hotkey.ctrl = event.ControlDown();
+		new_hotkey.shift = event.ShiftDown();
+		if (new_hotkey.raw != old_hotkey.raw)
+		{
+			auto& cfg_hotkey = *m_input_to_cfg_map[obj];
+			need_to_save |= (new_hotkey.raw != old_hotkey.raw);
+			cfg_hotkey = new_hotkey;
+			hotkey_to_func_map.erase(old_hotkey.raw);
+			hotkey_to_func_map[cfg_hotkey.raw] = cfg_hotkey_to_func_map[&cfg_hotkey];
```
goeiecool9999, line 107: Does this account for an edge case where a user selects the same key combination for multiple hotkeys? Should that be allowed?

src/gui/input/HotkeySettings.cpp
Anime37: oh, that's exactly what i was looking for and was surprised when i wasn't able to find. thanks.

src/gui/input/HotkeySettings.cpp
Anime37: good point, it doesn't. probably will block duplicate setups (to avoid having to parse vectors during key press captures).

src/gui/input/HotkeySettings.h
Anime37: that's my style too, somehow missed this one though

src/gui/input/HotkeySettings.cpp
Anime37: hm ya, mainly did this for fullscreen hotkeys. will see if i can store a static pointer to mainwindow on init or something.

Anime37: switched to storing pointer to main window on init.

src/gui/input/HotkeySettings.h
Anime37: fixed

src/gui/input/HotkeySettings.cpp
Anime37: removed `m_input_to_cfg_map` and done everything with `SetClientData` and `GetClientData` as was suggested.

src/gui/input/HotkeySettings.cpp
```diff
@@ -0,0 +1,151 @@
...
+wxFrame* s_main_window = nullptr;
+
+static auto& cfg_hotkeys = GetConfig().hotkeys;
+static std::unordered_map<uHotkey*, std::function<void(void)>> cfg_hotkey_to_func_map{
+	{&cfg_hotkeys.toggle_fullscreen, [](void) {s_main_window->ShowFullScreen(!s_main_window->IsFullScreen());}},
+	{&cfg_hotkeys.toggle_fullscreen_alt, [](void) {s_main_window->ShowFullScreen(!s_main_window->IsFullScreen());}},
+	{&cfg_hotkeys.exit_fullscreen, [](void) {s_main_window->ShowFullScreen(false);}},
+	{&cfg_hotkeys.take_screenshot, [](void) {g_window_info.has_screenshot_request = true;}},
+	{&cfg_hotkeys.toggle_fastforward, [](void) {ActiveSettings::SetTimerShiftFactor((ActiveSettings::GetTimerShiftFactor() < 3) ? 3 : 1);}},
+};
```
goeiecool9999, line 16: Since the number of distinct hotkey bindings is known at compile time, wouldn't a fixed size array with enumerator indices suffice? That probably applies to all currently used hashmaps.

Anime37: hmmm, sounds like a bit of an extra step when adding new hotkeys though. maybe could get away with making this map `const`?

as for the only other map left - `hotkey_to_func_map` - i pretty much go from input to func via hash, so i really don't want to change that one especially.

Anime37: ya, pointers as keys are scary, but since it's pointers to static data, i guess it's a edge-case where it stops being scary 😆

Anime37: made the map `const` anyways, maybe some of unordered_map data will go to .rodata, who knows.

Anime37:
fixed issue where you pressing another hotkey input button wouldn't restore the previously clicked hotkey input button.

Anime37:
cleanup up some formatting issues after setting up vscode environment with project's coding style rules in mind.

src/gui/input/HotkeySettings.cpp
```diff
@@ -0,0 +1,161 @@
...
+}
+
+void HotkeySettings::on_hotkey_input_click(wxCommandEvent& event)
+{
+	if (active_hotkey_input_btn)
+	{
+		wxKeyEvent revert_event{};
+		revert_event.SetEventObject(active_hotkey_input_btn);
+		on_key_up(revert_event);
+	}
```
goeiecool9999, line 92: Instead of invoking the event handler, perhaps it would be better to separate out the part of on_key_up that resets the button into a separate method and call that from both on_key_up and on_hotkey_input_click?

src/gui/input/HotkeySettings.h
```diff
@@ -0,0 +1,32 @@
...
+	inline static std::unordered_map<int, std::function<void(void)>> hotkey_to_func_map{};
+
+	HotkeySettings(wxWindow* parent);
+	~HotkeySettings();
+
+private:
+	wxPanel* m_panel;
+	wxFlexGridSizer* m_sizer;
+	std::vector<HotkeyEntry> m_hotkeys;
+	wxButton* active_hotkey_input_btn = nullptr;
```
goeiecool9999, line 21: Missing m_

Anime37: added missing prefix

src/gui/input/HotkeySettings.cpp
Anime37: should be better now.

Anime37:
will adapt coding style to conform with repo's coding style rules. putting on draft till that is done.

Anime37:
fixed all occurrences (i hope atleast) of code style being broken.

Anime37:
moved globals into class scope (i think looks cleaner)

Anime37:
disabled default hotkey for newly added fast-forward to not mess up anyone's current setups with unexpected behaviors when pressing SPACE.

Anime37:
added an ability to cancel hotkey input and clear configured hotkey with a right mouse button click

Anime37:
@goeiecool9999 since many things changed after adding controller hotkey support in https://github.com/cemu-project/Cemu/pull/1523 (mainly due to `uHotkey` becoming `sHotkeyCfg`), should port the new structure to this MR but omit all the controller related things. would make the controller hotkey support PR easier to review.

goeiecool9999:
I think there's no need for that. We'll have to review everything anyway. It looks like the sHotkeyCfg struct is specifically intended to support binding both keyboard and controller inputs. I don't think it makes much sense to backport that to a PR that only supports keyboard bindings. But to be honest I don't really even see the point in having a separate PR to add the controller support to the hotkeys. I think we should merge it all at once.

Anime37:
> I think there's no need for that. We'll have to review everything anyway. It looks like the sHotkeyCfg struct is specifically intended to support binding both keyboard and controller inputs. I don't think it makes much sense to backport that to a PR that only supports keyboard bindings. But to be honest I don't really even see the point in having a separate PR to add the controller support to the hotkeys. I think we should merge it all at once.

aight, i'll make it so this PR has both keyboard and controller support then.

Anime37:
restructured PR so that it would include the new structure + controller hotkey support.
split keyboard and controller into separate commits.

Anime37:
fixed some code styling issues (snake notation stuff)

Anime37:
Hm, guess i didn't commit something when it comes to the icon. Will check tomorrow.

Anime37:
seems like vcpkg issue, sync fork and rebased to latest main. hopefully this works.

Anime37:
included resources.h for Linux and MacOS (monkey see, monkey do)

Anime37:
made sure everything builds on my fork repo, so ready to be reviewed.

Exzap:
The UI needs a bit more work. There should be spacing between text labels and the window border. You can look at the code of other windows for common values that we use. The buttons should be centered on the same height as the labels next to them, right now they look misaligned.

src/gui/input/HotkeySettings.cpp
```diff
@@ -0,0 +1,363 @@
...
+}
+
+wxString HotkeySettings::To_wxString(uKeyboardHotkey hotkey)
+{
+	wxString ret{};
+	if (hotkey.raw)
+	{
+		if (hotkey.alt)
+		{
+			ret.append("ALT + ");
```
Exzap, line 340: To make these strings localizable wrap it inside `_("string")`. E.g. the whole line becomes:
`ret.append(_("ALT + "));`
Also a nitpick: I think only capitalizing the first letter is what is more commonly done and at least to me looks better. So `Ctrl` instead of `CTRL`

src/gui/input/HotkeySettings.cpp
```diff
@@ -0,0 +1,363 @@
...
+		if (hotkey.shift)
+		{
+			ret.append("SHIFT + ");
+		}
+		ret.append(wxAcceleratorEntry(0, hotkey.key).ToString());
+	}
+	return ret;
+}
+
+wxString HotkeySettings::To_wxString(ControllerHotkey_t hotkey)
```
Exzap, line 355: To match guidelines: `ToWxString`

src/input/api/Controller.cpp
```diff
@@ -67,6 +69,22 @@ const ControllerState& ControllerBase::update_state()
 
 #undef APPLY_AXIS_BUTTON
 
+	/* hotkey capturing */
+	const auto& hotkey_mod = HotkeySettings::s_cfgHotkeys.modifiers.controller;
```
Exzap, line 73: Working with internal state of `HotkeySettings` here. Preferably you would encapsulate this into function(s) of HotkeySettings. Right now the logic of Hotkey tracking leaks into other classes and thats something that should be avoided.

src/gui/input/HotkeySettings.cpp
```diff
@@ -0,0 +1,363 @@
...
+	{
+		keyInput->SetClientData(&hotkey);
+		keyInput->SetId(static_cast<wxWindowID>(InputButtonType::Keyboard));
+		controllerInput->SetClientData(&hotkey);
+		controllerInput->SetId(static_cast<wxWindowID>(InputButtonType::Controller));
+	}
+};
+
+HotkeySettings::HotkeySettings(wxWindow* parent)
+	: wxFrame(parent, wxID_ANY, "Hotkey Settings")
```
Exzap, line 43: Make the strings in this function localizable. E.g. `_("Hotkey Settings")`, same for the hotkey row labels

src/gui/input/HotkeySettings.cpp
```diff
@@ -0,0 +1,363 @@
...
+	s_keyboardHotkeyToFuncMap.reserve(s_cfgHotkeyToFuncMap.size());
+	for (const auto& [cfgHotkey, func] : s_cfgHotkeyToFuncMap)
+	{
+		auto keyboardHotkey = cfgHotkey->keyboard.raw;
+		if (keyboardHotkey > 0)
+		{
+			s_keyboardHotkeyToFuncMap[keyboardHotkey] = func;
+		}
+		auto controllerHotkey = cfgHotkey->controller;
+		if (controllerHotkey > 0)
```
Exzap, line 94: Shouldn't this be `>= 0` because -1 is the value for disabled controller hotkeys and not 0?

src/gui/input/HotkeySettings.cpp
Anime37: made strings localizable and made modifier's names like suggested

src/input/api/Controller.cpp
Anime37: encapsulated logic in `CaptureInput` methods for both keyboard and controller hotkeys

src/gui/input/HotkeySettings.cpp
Anime37: oops, fixed. also defined `None` values in `sHotkeyCfg` struct. should look better now

Anime37:
> The UI needs a bit more work. There should be spacing between text labels and the window border. You can look at the code of other windows for common values that we use. The buttons should be centered on the same height as the labels next to them, right now they look misaligned.

will address this later, since it will take more time to analyze the code.

Anime37:
reworked the UI.

had to rework the key/controller hotkey differentiation logic in order to change button colors to be the same as in `Input Settings`.

took some inspiration from Elden Ring for displaying unbound/edit-mode hotkeys 😄 

image

Exzap:
Took a while to review, apologies for that. It looks good with only minor nitpicks where some strings are not translate-able (missing the `_("") around the string`) but since we are planning changes in that regard anyway it's actually for the better.

Thanks for your contribution! I am sure this will be appreciated by many users.

PR #1546, opened 2025-04-27
Fix opening of save directory on macOS

neebyA:
On macOS, trying to open save directory locations through the context menu usually results in errors like:
```
Failed to open URL "file:/Users/xxxxx/Library/Application Support/Cemu/mlc01/usr/save/00050000/xxxxxxxx" in default browser. (error 0: Undefined error: 0)
```
This is due to the default MLC path on macOS being located at `~/Library/Application\ Support/Cemu/mlc01`, which contains a space. I tried escaping with a `\` or surrounding the path with `""`, but it appears that since `wxLaunchDefaultBrowser` is being used then URI encoding is required. ~I couldn't find a URI encoding function in either `<wx/uri.h>` or `<wx/url.h>`, so I just did a manual string replacement of all spaces.~

~Alternatively,~ it's possible to just use `wxLaunchDefaultApplication` instead like this: 
https://github.com/cemu-project/Cemu/blob/b089ae5b32f5f2bc901b9ff2db2424cbf3c1186b/src/gui/MainWindow.cpp#L683-L684

I have not tested this on Linux or Windows, though this change doesn't seem like it would break anything in either

### DIFF
--- a/src/gui/components/wxGameList.cpp
+++ b/src/gui/components/wxGameList.cpp
@@ -668,7 +668,7 @@ void wxGameList::OnContextMenuSelected(wxCommandEvent& event)
 				{
 				fs::path path(gameInfo.GetBase().GetPath());
 				_stripPathFilename(path);
-				wxLaunchDefaultBrowser(wxHelper::FromUtf8(fmt::format("file:{}", _pathToUtf8(path))));
+				wxLaunchDefaultApplication(wxHelper::FromPath(path));
 				break;
 				}
 			case kWikiPage:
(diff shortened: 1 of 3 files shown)


neebyA:
Confirmed that this also works:
```
wxLaunchDefaultApplication(wxHelper::FromPath(gameInfo.GetSaveFolder()));
```
Perhaps this is preferred as it doesn't require any additional string editing?

src/gui/components/wxGameList.cpp
```diff
@@ -689,7 +689,9 @@ void wxGameList::OnContextMenuSelected(wxCommandEvent& event)
 
 			case kContextMenuSaveFolder:
 			{
-				wxLaunchDefaultBrowser(wxHelper::FromUtf8(fmt::format("file:{}", _pathToUtf8(gameInfo.GetSaveFolder()))));
+				// escape any spaces in file path
+				std::string saveFolder = boost::replace_all_copy(_pathToUtf8(gameInfo.GetSaveFolder()), " ", "%20");
+				wxLaunchDefaultBrowser(wxHelper::FromUtf8(fmt::format("file:{}", saveFolder)));
```
neebyA, line 694:
```suggestion
				// handles any spaces in file path
				wxLaunchDefaultApplication(wxHelper::FromPath(gameInfo.GetSaveFolder()));
```
Confirmed that this also works. Perhaps it should be preferred as it does not require any additional string editing?

JaimieVandenbergh:
Thanks for this - it's been a longstanding minor annoyance :) 
Note that when you right-click a game and choose "Game directory" "Update directory" "DLC directory" these all suffer the same issue as "Save directory" and would probably be fixable the same way.

src/gui/components/wxGameList.cpp
emiyl: Probably best to use `wxHelper::FromPath`, as it can then handle _any_ special characters in the string.

neebyA: Updated

neebyA:
Opening directories from the title manager also saw this issue so I updated that in the latest commit, as well as the for the game, update, and DLC directories, as trying to open them would also lead to errors if the directory paths included any spaces.

Exzap:
Thanks!

PR #1559, opened 2025-05-10
CI: Bump MoltenVK to v1.3.0

shinra-electric:
This bumps the MVK version to 1.3.0 in the CI. 

Not sure if this is the only place it needs to be changed, but I wasn't able to find another reference to it. 

MVK 1.3.0 adds full Vulkan 1.3 support and many bug fixes.

Full changelog is [here](https://github.com/KhronosGroup/MoltenVK/releases/tag/v1.3.0).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -202,7 +202,7 @@ jobs:
 
     - name: "Install molten-vk"
       run: |
-        curl -L -O https://github.com/KhronosGroup/MoltenVK/releases/download/v1.2.9/MoltenVK-macos.tar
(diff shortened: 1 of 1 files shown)


Exzap:
Would be nice to get some feedback on whether there are any new issues. We had to [downgrade](https://github.com/cemu-project/Cemu/commit/593da5ed79cab9ca175391d0ba6666a1f8a52500) from 1.12.10 so I wonder if those problems have been fixed.

shinra-electric:
From what I saw on Discord, I think 1.2.10 crashed on launch? 

With this PR there is no crash: 
Screenshot 2025-05-10 at 12 03 34

I'll ask on Discord for other people to test it with a few games looking for regressions.

Exzap:
Good enough for me, worst case we can just revert again. Thanks for taking the time

PR #1564, opened 2025-05-12
dmae: Implement 16bit endian swap for DMAECopyMem

Mefiresu:
Simply adapted from the 32bit behavior above.
Needed it for an homebrew app.

### DIFF
--- a/src/Cafe/OS/libs/dmae/dmae.cpp
+++ b/src/Cafe/OS/libs/dmae/dmae.cpp
@@ -36,6 +36,16 @@ void dmaeExport_DMAECopyMem(PPCInterpreter_t* hCPU)
 			dstBuffer[i] = _swapEndianU32(srcBuffer[i]);
 		}
 	}
+	else if( hCPU->gpr[6] == DMAE_ENDIAN_16 )
+	{
+		// swap per uint16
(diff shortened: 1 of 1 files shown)


src/Cafe/OS/libs/dmae/dmae.cpp
```diff
@@ -36,6 +36,16 @@ void dmaeExport_DMAECopyMem(PPCInterpreter_t* hCPU)
 			dstBuffer[i] = _swapEndianU32(srcBuffer[i]);
 		}
 	}
+	else if( hCPU->gpr[6] == DMAE_ENDIAN_16 )
+	{
+		// swap per uint16
+		uint16* srcBuffer = (uint16*)memory_getPointerFromVirtualOffset(hCPU->gpr[4]);
+		uint16* dstBuffer = (uint16*)memory_getPointerFromVirtualOffset(hCPU->gpr[3]);
+		for(uint32 i=0; i<hCPU->gpr[5]; i++)
```
Exzap, line 44: I am unsure but it might be size * 2 here because it counts in multiple of 4 bytes. Do you know what behavior the homebrew expects?

Mefiresu: You are definitely right, I didn't take into account the size * 2 in the homebrew code and it ended up working anyway...

Exzap:
thanks!

Issue #592, opened 2022-12-27
CTRL + Q and/or Gamepad Hotkeys for closing Cemu?


sonicnkt:
Just played around with the latest linux build (2.0-22) and it works pretty great on my system. 
I added a few games to steam so i can also launch them from my tv in another room using a gamepad.

This worked great as well but sadly i couldn't figure out a way to close the emulator without a keyboard/mouse as the steam overlay isn't working correctly

Some emulators support gamepad hotkeys (like yuzu for example) and for some other applications i use antimicrox  but even the last option didn't work for me. 

 (Almost) all desktop apps can be closed using CTRL + Q and with a tool like antrimicro i can easily add this as a button combo to a profile which is loaded before the emulator is started and removed afterwards. The mentioned key combo isn't working tho in the latest appimage.   

Adding Gamepad hotkeys is definitely a bigger implementation but maybe support for CTRL+Q (or another shortcut) could be added for working around this issue.

jn64:
While I'm not against the suggestion, I've never noticed the lack of a close shortcut in any app, because practically every DE/WM has a global shortcut for closing windows. GNOME, KDE, and Xfce (xfwm) all have it as Alt-F4 by default (which I always change to Super-q).

Having "Quit Cemu" as a bindable option would be useful though, so you don't have to use an external tool to bind it to controller.

sonicnkt:
Yeah on gnome it defaults to SUPER + Q which i am used to by now. Many Apps also support CTRL + Q tho. 
I am using antimicrox (flatpak), but the SUPER + Q combination isn't working here for some reason,  other button combinations (which don't use SUPER) work fine tho so this might be a gnome limitation. 
I discovered the CTRL-Q thing by accident and it worked in any other app (besides terminals). 

My solution for now is to keep using antimicrox with a more complicated setup:
Use GUIDE Button (Hold) to switch to another profile and here i mapped one button to ALT-ENTER and another button which switches to a third profile which enables mouse input so i can close the app manually. A bit complicated but nothing that can be accidentally triggered. 

Custom Hotkeys which support Gamepad input would be a great solution tho in the future.

jn64:
For your workaround, surely you could do something simpler with xdotool or wmctrl to close the window?

```sh
# Filter by both classname and name to avoid closing the wrong window,
# like a browser window with "Cemu" in the title
xdotool search --classname 'Cemu' search --name 'Cemu 2.0' windowquit

[...]

PR #1565, opened 2025-05-12
Disable Ctrl+Q shortcut on non-macOS platforms to prevent accidental exits during gameplay

gamersalpha:
This PR updates the `MainWindow::OnKeyDown` logic to prevent `Ctrl+Q` from closing the application on Windows and Linux.

### Why this change?

On non-macOS platforms, `Ctrl+Q` is not a standard quit shortcut. However, it was previously hardcoded in Cemu to close the window, using `event.CmdDown() && event.GetKeyCode() == 'Q'`, which maps to `Ctrl+Q` on Windows/Linux and `Cmd+Q` on macOS.

For users who map `Ctrl+Q` (or press it accidentally) during gameplay—especially those playing with keyboard controls—this could result in an unexpected app exit and potential progress loss.

### What this PR changes

- The quit shortcut `Cmd+Q` now works **only** on macOS.
- On Windows/Linux, only `Alt+F4` triggers a clean shutdown.
- `Ctrl+Q` is ignored on non-macOS systems to avoid unintended closures.

### Tested:
✅ Windows 11, built with MSVC  
✅ Cemu launches and closes as expected using `Alt+F4`

### DIFF
--- a/src/gui/MainWindow.cpp
+++ b/src/gui/MainWindow.cpp
@@ -1448,15 +1448,23 @@ void MainWindow::OnKeyUp(wxKeyEvent& event)
 
 void MainWindow::OnKeyDown(wxKeyEvent& event)
 {
-	if ((event.AltDown() && event.GetKeyCode() == WXK_F4) || 
-		(event.CmdDown() && event.GetKeyCode() == 'Q'))
-	{
-		Close(true);
-	}
-	else
-	{
-		event.Skip();
-	}
+#if defined(__APPLE__)
+       // On macOS, allow Cmd+Q to quit the application
+    if (event.CmdDown() && event.GetKeyCode() == 'Q')
+    {
+        Close(true);
+    }
+#else
+     // On Windows/Linux, only Alt+F4 is allowed for quittinger
+    if (event.AltDown() && event.GetKeyCode() == WXK_F4)
+    {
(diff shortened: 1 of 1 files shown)


src/gui/MainWindow.cpp
```diff
@@ -1448,15 +1448,23 @@ void MainWindow::OnKeyUp(wxKeyEvent& event)
...
-		event.Skip();
-	}
+#if defined(__APPLE__)
+       // On macOS, allow Cmd+Q to quit the application
+    if (event.CmdDown() && event.GetKeyCode() == 'Q')
+    {
+        Close(true);
+    }
+#else
+     // On Windows/Linux, only Alt+F4 is allowed for quittinger
```
Exzap, line 1458:
```suggestion
     // On Windows/Linux, only Alt+F4 is allowed for quitting
```

Exzap:
Thanks!

Squall-Leonhart:
Ctrl+Q is a shortcut known to close applications on a number of linux apps on several distros (Ubuntu, Debian, Fedora) and has frequently been requested ([was requested of cemu infact](https://github.com/cemu-project/Cemu/issues/592)) to be used as such, using the control modifier within the mapped controller seems like user error if anything.

PR #1571, opened 2025-05-18
GameList: Allow sorting by more columns

goeiecool9999:
Implement sorting by Play Time, Last started, Region and Title ID

Q: How should this interact with Favorites. Should they still float to the top of the list when sorting by something other than name?


Exzap:
I think the sorting takes precedence over favorites. It would be weird to sort by play time and then have favorites sorted independently of non-favorites. So I would say just ignore the favorite flag for anything but the default.

PR #1576, opened 2025-05-26
Specify the size of the memset

FireBurn:


Squall-Leonhart:
the correct fix is `memset((void*)ivOut, 0, sizeof(ivOut));` not hardcoding the value and overriding a safeguard for is the array size were to change.

goeiecool9999:
> the correct ~fix~ _change_ is `memset((void*)ivOut, 0, sizeof(ivOut));` not hardcoding the value and overriding a safeguard for if the array size were to change, which is what your change does.
> 
> ivOut is a fixed-size array and sizeof(ivOut) correctly evaluates to 16, the warning is a false alarm.

It doesn't evaluate to the correct value due to array-to-pointer decay
https://godbolt.org/z/8qMsab49E

Anyhow this PR has been succeeded by Exzap's 7168d20

FireBurn:
Thanks for the propper fix

PR #1585, opened 2025-05-31
build: Fix undefined SPIRV symbols on linux

### LINKED ISSUE #1544: Cannot build CEMU since the upgrade to Fedora 42.
### Current Behavior

Hello, I want to report a problem that I can't seem to fix... It is now impossible for me to build CEMU since I upgraded from Fedora 41 to Fedora 42, Fedora 41 had GCC 14 and Clang 19, while Fedora 42 has Clang 20 and GCC 15.

Basically, when compiling with VCPKG enabled it will fails at the SDL2 package, the error has been fixed in upstream VCPKG for a long time now, the problem was caused by PipeWire 1.4(Fedora 41 had PipeWire 1.2).
If you choose to build without VCPKG then it'll compile fine but it'll fail at link time because of multiple undefined reference to glslang/spirv-tools functions(happens with GCC/Clang ld/lld).
(Fedora 42 has GLSLANG 15.1)

### Expected Behavior

Successful build of CEMU on Fedora 42.

### Steps to Reproduce

Install Fedora 42 in a VM.
Try to build CEMU, both with and without VCPKG.
See the build fails.

### System Info (Optional)

OS: Fedora Linux 42 KDE Plasma Edition
GPU: RTX 4060 8GB Laptop GPU
CPU: Ryzen 7 7435HS 8C/16T 4.55GHz
RAM: 40GB DDR5 4800MHz

_No response_

### Logs (Optional)

_No response_

ColinKinloch:
For some reason on Fedora 42 linking glsl doesn't link it's SPIRV-Tools dependencies.

At the linking stage cmake spits out `undefined reference to` symbols with the prefix `spv` or in the namespace `spvtools`.

Fixes #1544


goeiecool9999:
I see that the issue this is meant to solve was fixed by #1622. Can this be closed?

PR #1586, opened 2025-06-01
UI: wxCAPTION flag on input API dialog to fix kwin

### LINKED ISSUE #867: DSUController IP configuration is missing on Wayland
Building through the Arch Linux AUR [package](https://aur.archlinux.org/packages/cemu).

> cemu --version
> 2.0-353 (experimental)

The DSUController configuration screen is supposed to show a dialogue with an IP address. On Wayland it doesn't show up.

image

On X11 (through XWayland) it does show up.

image

ColinKinloch:
The IP and port input boxes are not shown when the DSUController API is first selected when using wayland with kwin.

Setting the dialog as wxCAPTION causes kwin to add decorations to the window which for whatever reason causes the toplevel to resize correctly.

This may be a bug in wxWidgets or kwin.

With dialog style set to 0:

https://github.com/user-attachments/assets/42d15f24-f402-4217-bf85-594db658e063

With dialog style set to wxCAPTION:

https://github.com/user-attachments/assets/a05d0860-308d-4f31-9501-4df4f7f3bd77

Fixes #867

### DIFF
--- a/src/gui/input/InputAPIAddWindow.cpp
+++ b/src/gui/input/InputAPIAddWindow.cpp
@@ -23,7 +23,7 @@ using wxControllerData = wxCustomData<ControllerPtr>;
 
 InputAPIAddWindow::InputAPIAddWindow(wxWindow* parent, const wxPoint& position,
                                      const std::vector<ControllerPtr>& controllers)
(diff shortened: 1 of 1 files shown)


Exzap:
The line between what is a wxWidgets quirk and what is a bug is blurry. But either way I think it looks better with a titlebar anyway. Thanks!

PR #1587, opened 2025-06-01
"verbose" command line argument to log to stdout

ColinKinloch:
I find the way Cemu logs intuitive.

When I try to track down an issue too much of my time is spent trying to figure out where my log lines are going.

What I want is to type `Cemu --help`, see that there's an option to print to stdout and use that.


Exzap:
Can you make this option be compiled conditionally only for non-Windows platforms? On Windows, GUI applications have a nul device attached to stdout. We use a very hacky workaround to make some of the commands (like --help) work, but it isn't suitable for long term logging. Easier to just not provide this on Windows for now

PR #1592, opened 2025-06-07
Add FSTimeToCalendarTime function

Nintendocustom:
This adds FSTimeToCalendarTime to coreinit_Time.cpp, which is commonly used in Lego games to convert a given timestamp.
grafik
grafik


Exzap:
How accurate is this? Was this tested with anything other than Lego games?

Nintendocustom:
It should be pretty accurate. However, I don’t know of any other games that use it. In Lego City Undercover, Cemu currently passes zero as parameter, so it defaults to 1/1/1980. You can leave the PR open until some other games that use it are found.

PR #1593, opened 2025-06-08
boss: Use HTTP/1.1 instead of default

Wiichele:
The Pretendo BOSS server does not accept HTTP/2 connections, even when the request is correctly formed (it assumes you're using a web browser). This, from what I've seen online, is not a problem because libcurl defaults to HTTP/1.1. Except on the Flatpak build for some reason. This fixes Splatoon for me.


Exzap:
Good catch!

emansom:
Ideally the backend should support this? @jonbarrow

Exzap:
Wii U doesn't use it and it's a good security practice to disable anything that isn't needed.

PR #1599, opened 2025-06-16
use Microsoft::WRL::ComPtr

oltolm:
Use `ComPtr` where possible. It's a smart pointer that does reference counting for you automatically.


src/audio/DirectSoundAPI.cpp
```diff
@@ -1,9 +1,11 @@
 #include "DirectSoundAPI.h"
 
+#include "XAudio2API.h"
```
Exzap, line 3: Out of curiosity why are you including the XAudio header here?

Squall-Leonhart:
Xaudio 2.8+ doesn't actually need COM Initialization in this way at all, 

Instead of using CoCreateInstance, you now call XAudio2Create directly, which returns a raw pointer to the IXAudio2 interface. This means you can manage it with a smart pointer like std::unique_ptr without needing Microsoft::WRL::ComPtr

doing this on Xaudio27 would have made sense though.

PR #1603, opened 2025-06-18
Fix deprecation warnings in curl and pugixml

oltolm:
I updated the code to use the new APIs.

### DIFF
--- a/src/Cafe/OS/libs/nlibcurl/nlibcurl.cpp
+++ b/src/Cafe/OS/libs/nlibcurl/nlibcurl.cpp
@@ -1098,7 +1098,7 @@ size_t read_callback(char* buffer, size_t size, size_t nitems, void* instream)
 }
 
 
-int progress_callback(void* clientp, double dltotal, double dlnow, double ultotal, double ulnow)
+int progress_callback(void* clientp, curl_off_t dltotal, curl_off_t dlnow, curl_off_t ultotal, curl_off_t ulnow)
 {
 	//peterBreak();
 	CURL_t* curl = (CURL_t*)clientp;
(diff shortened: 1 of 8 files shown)


src/Cafe/OS/libs/nlibcurl/nlibcurl.cpp
```diff
@@ -1380,11 +1380,11 @@ void export_curl_easy_getinfo(PPCInterpreter_t* hCPU)
...
-		case CURLINFO_SIZE_DOWNLOAD:
-		case CURLINFO_SPEED_DOWNLOAD:
-		case CURLINFO_SIZE_UPLOAD:
-		case CURLINFO_SPEED_UPLOAD:
-		case CURLINFO_CONTENT_LENGTH_DOWNLOAD:
+		case CURLINFO_SIZE_DOWNLOAD_T:
+		case CURLINFO_SPEED_DOWNLOAD_T:
+		case CURLINFO_SIZE_UPLOAD_T:
+		case CURLINFO_SPEED_UPLOAD_T:
+		case CURLINFO_CONTENT_LENGTH_DOWNLOAD_T:
```
Exzap, line 1387: `export_curl_easy_getinfo` runs in the context of the emulated process, so the parameters that are passed to it come from the game. We have to handle the non-`_T` version of the parameters here since thats what games use. (nlibcurl is based on libcurl from around 2012-2013, the `_T` parameters were added much later)

The solution here is generally to write glue code that redefines the outdated constants (if they are no longer available or marked as deprecated) and convert it to whatever the equivalent in the modern version of the lib is. Similar to what our Wii U socket implementation code does (as an example see [here](https://github.com/cemu-project/Cemu/blob/4f4412b3341d4ac02377e5aa66eddbcfe2072c5c/src/Cafe/OS/libs/nsysnet/nsysnet.cpp#L38) and [setsockopt](https://github.com/cemu-project/Cemu/blob/4f4412b3341d4ac02377e5aa66eddbcfe2072c5c/src/Cafe/OS/libs/nsysnet/nsysnet.cpp#L489)).

oltolm: Thanks for the explanation. I fixed the code, I think.

Exzap:
I have been thinking about how to handle these refactor PRs that touch sensitive code. I am stretched thin between other projects and Cemu work, so my capacity to test and double check everything is close to zero. But also I believe that if someone submits a refactor PR it should be error free except for genuine, but hopefully rare, mistakes. That means that changes have to be tested *well* and if you cannot do that you should not submit it.

The changes in this PR are pretty harmless for the most part but it's also easy to slip up somewhere since curl API isn't type safe and this can then lead to issues that can be very hard to track down later. And it's almost certainly not you who is going to have to deal with it. Games usually don't give us nice error logging, a failed download can manifest as some weird issue minutes after the bug and the association may not be obvious. 5 minutes for a single change can translate to 5 hours of debugging time. So maybe that explains a bit why I am hesitant to merge pure refactor PRs.

I see you have loads of other refactor commits queued on your fork. I think it's great but if it's purely cosmetic then it doesn't really move the project forward. If you really want to help us out, while doing refactor work beyond just the UI, you could look into setting up the Wii U homebrew environment [WUT](https://github.com/devkitPro/wut) and accompany your changes with unit tests when it makes sense, and for changes that don't touch any of the HLE libs you could still do C++ unit tests when applicable. It is more work, but we really desperately lack unit tests in general and it's the root issue why we can't just easily merge changes like this. But it's up to you if you are interested in that.

oltolm:
I get it, then I'm closing the ticket. But UI PRs would be OK I guess?

Issue #1591, opened 2025-06-07
Unable to recognize Wiimote extension (3rd-party Classic Controller)


clamcee:
### Current Behavior

Cemu not recognizing a 3rd-party Classic Controller connected to my Wiimote in the input settings

### Expected Behavior

Cemu recognizing the Classic Controller in the input settings and the inputs from that controller

### Steps to Reproduce

1. Open Cemu
2. Open Dolphin(for connecting Wiimote, skip to 7)
3. Go to Controller Settings
4. Turn off then on the "Continuous Scanning" option
5. Press SYNC button on Wiimote (connects)
6. Close Dolphin
7. Open Cemu window
8. Go to Options -> Input settings
9. Click on drop-down menu of "Controller"
10. API: Wiimote
11. Controller: Controller 1
12. Press Add
13. Press Settings
14. Plug Classic Controller into Wiimote
15. Notice Cemu not recognizing Classic Controller

### System Info (Optional)

OS: 
GPU:

Squall-Leonhart:
what model wii-mote is it?

OS required too

capitalistspz:
If it's a Classic Controller Pro, it won't be detected

clamcee:
> If it's a Classic Controller Pro, it won't be detected

what's a classic controller pro

clamcee:
> what model wii-mote is it?

https://www.google.com/imgres?imgurl=https%3A%2F%2Fm.media-amazon.com%2Fimages%2FI%2F41vCe7eWTxL.jpg&tbnid=Bh_LE0pDSYLPzM&vet=1&imgrefurl=https%3A%2F%2Fwww.amazon.com%2FWii-Nunchuck-Controllers-Certified-Refurbished-Nintendo%2Fdp%2FB07N4D6MWL&docid=t09CA3A5dxrzbM&w=479&h=1005&hl=vi-VN&source=sh%2Fx%2Fim%2Fm1%2F4&kgs=8f6c7551368edb20&shem=isstc
> 
> OS required too

Windows 10

Squall-Leonhart:
> what's a classic controller pro

![Image](https://github.com/user-attachments/assets/4973c826-78fe-407b-846b-530ed03abecf)

[...]

Issue #1301, opened 2024-08-18
Linux / Wiimote API: Classic Controller Pro is not detected


git-developer:
### Current Behavior

When a [Classic Controller Pro](https://en.wikipedia.org/wiki/Classic_Controller#Classic_Controller_Pro) is used as extension on a Wiimote, it is not detected ("Connected extension: None").

Additional Info:
- An original [Classic Controller](https://en.wikipedia.org/wiki/Classic_Controller) is detected
- In `xwiishow` and `evtest`, the Classic Controller Pro is detected and working correctly.
- The uevent data is identical for Classic Controller and Classic Controller Pro, and a [comment in the linux driver](https://github.com/xwiimote/xwiimote/blob/4df713d9037d814cc0c64197f69e5c78d55caaf1/lib/xwiimote.h#L283) states:
    > The Classic Controller Pro is also reported via this interface, but cannot be distinguished from a normal classic controller extension.

### Expected Behavior

The Classic Controller Pro should be detected just as an original Classic Controller.

### Steps to Reproduce

1. Connect a Wiimote
1. Connect a Classic Controller Pro extension
1. Start Cemu and go to Input settings, enter Settings and look at the label that shows the extension

### System Info (Optional)

Cemu 2.0-91 Flatpak build on Linux (batocera v40)

git-developer:
Detected in https://github.com/batocera-linux/batocera.linux/pull/12291

kevincrans:
Possible duplicate of my 2022 question #200

Issue #200, opened 2022-09-07
Please add native Wii Classic Controller Pro support


kevincrans:
For all versions at least up to 2.0-1
If I connect a wii-remote using Dolphin like always, everything works perfectly on all versions now.
However, while the nunchuck extension gets recognized, the classic controller makes the extension field stay None.

This is my favourite deprecated controller, so I really love to see it work one time.

There are no IR pointer and motion mappings, meaning the input settings window should grow anyways.

I recommend a radio button to swap the nunchuck mapping fields with the classic controller instead of a checkbox to remain compact.

Do you think I need a log file?

kevincrans:
I have to add, I don't know what an U classic controller is, but if it is a misname I don't assume, this is about the physical controller, not just the mapping API.

Squall-Leonhart:
the classic control is supported by SDL2

kevincrans:
That's not fair, that the nunchuck cán connect without SDL, while the classic controller can't.

Cemu should support the whole wii-remote with classic controller without needing to install tools seperately, no matter how, just like Dolphin.

It's so wastefull, since it's as easy to add as the nunchuck.

Squall-Leonhart:
the wiiu games that can use the classic control can be counted on a single hand, these already work via the wiimote api.

Usage with anything else requires mapping the control to the wii u input devices, which SDL2 provides for, nothing further is required from Cemu's end.

kevincrans:
Not true, the cemu wiimote api is incomplete, else the classic controller would show up and pressing the buttons would map/do anything.

Squall-Leonhart:
Can be closed with the merging of https://github.com/cemu-project/Cemu/pull/1851

Squall-Leonhart:
> the cemu wiimote api is incomplete

not exactly, the classic controls are implemented, via the "Wii U Classic Controller" option, you can then map a classic controller extension via the wii mote api.

There was a bug in the logic that prevented the buttons from all assigning properly, this will be fixed with https://github.com/cemu-project/Cemu/pull/1851

PR #1614, opened 2025-06-26
input/wiimote: Detect Classic Controller Pro as Classic Controller

capitalistspz:
According to Wiibrew, they have the same packet format


capitalistspz:
May solve 
https://github.com/cemu-project/Cemu/issues/200
https://github.com/cemu-project/Cemu/issues/1301
https://github.com/cemu-project/Cemu/issues/1591

Issue #1618, opened 2025-06-27
Failed to build git on ARM64 Linux


salvalie:
### Current Behavior

Hi! I understand that this isn't fully supported, yet, I would like to point out the problems I faced trying to compile it with clang-15 as the wiki asks... I am not sure if utilizing clang-15 was a good idea to begin with. anyway, that wasn't the problem. 

console log at the bottom. 

### Expected Behavior

To compile, while not yet officially supported

### Steps to Reproduce

following the doc compilation process. 

### System Info (Optional)

```
salva@rock-5-itx:~/dev/Cemu$ fastfetch 
                ..                     salva@rock-5-itx
            `:]x**j-,'                 ----------------
       .,+t***********z\<"             OS: Armbian 25.8.0-trunk.279 noble aarch64
       ?******************;            Host: Radxa ROCK 5 ITX
      '*n` .'`^,;;,^`'. ,cc.           Kernel: Linux 6.1.115-vendor-rk35xx
      -<.                .[l           Uptime: 21 mins
     //     ^^      ^^    \\           Packages: 2423 (dpkg)
     !^         ^^         ":          Shell: bash 5.2.21
    'tt}`     !~]rj_     ")t/.         Display (22MP55): 1920x1080 @ 60 Hz in 22" [External]
    Itttt?'   ~~]rr]   `{tttt,         DE: GNOME 46.0
    \tttttt!""I_]r("""~tttttt1         WM: Mutter (Wayland)
  '_tttttttttttt)ftttttttttttti.       WM Theme: Adwaita
 \*ztttttttttttttttttttttttttf**[      Theme: Adwaita [GTK2/3/4]
l**c)tttttttttttttttttttttttt(z**,     Icons: Numix-Circle [GTK2/3/4]
.z*x.`tttttttttttttttttttttttt.`u*n    Font: Cantarell (11pt) [GTK2/3/4]
>`   (tttttttttttttttttttttt]   "I     Cursor: DMZ-White (24px)
     ,tttttttttttttttttttttt`          Terminal: GNOME Terminal 3.52.0
     ./tttttfttttttttfttttt(           Terminal Font: Monospace (11pt)
      'I)))(\()(tt))|\()({;'           CPU: rk3588 (8) @ 2.40 GHz
        .~~~~~~~|)~~~~~~~<             GPU: Mali-G610 [Integrated]
        '[)))))1|()))))))?             Memory: 2.85 GiB / 15.58 GiB (18%)
[...]

PR #1630, opened 2025-07-04
fix syntax of conditional assembly directives for aarch64

qurious-pixel:
fixes https://github.com/cemu-project/Cemu/issues/1618

### DIFF
--- a/dependencies/ih264d/CMakeLists.txt
+++ b/dependencies/ih264d/CMakeLists.txt
@@ -182,7 +182,7 @@ target_sources(ih264d PRIVATE
 "decoder/arm/ih264d_function_selector_av8.c"
 "decoder/arm/ih264d_function_selector.c"
 )
-target_compile_options(ih264d PRIVATE -DARMV8)
+target_compile_options(ih264d PRIVATE -DARMV8 $<$<COMPILE_LANGUAGE:ASM,Clang>:-Wno-unused-command-line-argument>)
 if(APPLE)
     target_sources(ih264d PRIVATE "common/armv8/macos_arm_symbol_aliases.s")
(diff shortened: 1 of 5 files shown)


dependencies/ih264d/CMakeLists.txt
```diff
@@ -182,7 +182,7 @@ target_sources(ih264d PRIVATE
 "decoder/arm/ih264d_function_selector_av8.c"
 "decoder/arm/ih264d_function_selector.c"
 )
-target_compile_options(ih264d PRIVATE -DARMV8)
+target_compile_options(ih264d PRIVATE -DARMV8 $<$<COMPILE_LANGUAGE:ASM,Clang>:-Wno-unused-command-line-argument>)
```
SSimco, line 185: I think the issue on aarch64 Linux is that clang doesn't run the C preprocessor on the assembly files.
According to https://developer.arm.com/documentation/100066/0611/assembling-assembly-code/preprocessing-assembly-code you could either replace the assembly code source file suffix from `.s` to `.S` or use the `-x assembler-with-cpp` option. Also, revert the assembly files back to use the `#ifdef` preprocessor directive.
```suggestion
target_compile_options(ih264d PRIVATE -DARMV8 $<$<COMPILE_LANGUAGE:ASM,Clang>:-Wno-unused-command-line-argument>)

if(NOT MSVC)
    set(CMAKE_ASM_FLAGS "${CFLAGS} -x assembler-with-cpp")
endif()
```

qurious-pixel: Yes, this is the right way. closing this PR in favor of @SSimco 's suggestion.

PR #1634, opened 2025-07-08
Windows: use modern API to set thread name

oltolm:
This uses the new API introduced in Windows 10 to set the thread name. It has advantages compared to the old method.

### DIFF
--- a/src/util/helpers/helpers.cpp
+++ b/src/util/helpers/helpers.cpp
@@ -116,6 +116,19 @@ typedef struct tagTHREADNAME_INFO
 void SetThreadName(const char* name)
 {
 #if BOOST_OS_WINDOWS
+	using SetThreadDescription_t = HRESULT (*)(HANDLE hThread, PCWSTR lpThreadDescription);
+	static SetThreadDescription_t pSetThreadDescription = nullptr;
+	if (!pSetThreadDescription)
+		pSetThreadDescription = (SetThreadDescription_t)GetProcAddress(LoadLibraryW(L"Kernel32.dll"), "SetThreadDescription");
+	if (pSetThreadDescription)
+	{
+		size_t len = strlen(name) * 2 + 1;
+		auto threadDescription = new wchar_t[len];
+		if (boost::nowide::widen(threadDescription, len, name))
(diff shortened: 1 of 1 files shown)


RedBlackAka:
This would drop support for W7, W8, W8.1 and W10 up to version 1511 entirely. Windows 10 1607 would be the new minimum. Would strongly recommend against this

oltolm:
> This would drop support for W7, W8, W8.1 and W10 up to version 1511 entirely. Windows 10 1607 would be the new minimum. Would strongly recommend against this

You didn't even look at the code. It is only used when available. Also it does not even build and I am not sure what to do about it.

oltolm:
I think I fixed it.

Exzap:
You should use `boost::nowide::widen` instead of `MultiByteToWideChar`

oltolm:
done

Exzap:
You can avoid the manual memory management and shorten it further to:
```
	auto nameWide = boost::nowide::widen(name);
	pSetThreadDescription(GetCurrentThread(), nameWide.c_str());
```
Technically could even avoid the temporary and make it one line but I think that sacrifices readability too much.
It's safe to assume that in practice `widen` will never fail/throw an exception here and `name` will never have a zero length, so any extra error handling for those cases is unnecessary.

oltolm:
My first version looked like this, but it didn't compile with MSVC. I got this error https://learn.microsoft.com/en-us/cpp/error-messages/compiler-errors-2/compiler-error-c2712?view=msvc-170. Basically it's not possbile to have an object with a destructor in this function because it requires unwinding. That's why I can't use `std::wstring` and have to do manual memory management.

goeiecool9999:
> Basically it's not possbile to have an object with a destructor in this function because it requires unwinding. That's why I can't use `std::wstring` and have to do manual memory management.

I read a suggestion on stack overflow to just move the __try block into it's own function. Have you tried that?

oltolm:
No I have not, is this not good enough?

Exzap:
I think keeping it all in one function is preferable, no need to nitpick this further

PR #1642, opened 2025-07-13
CI: Fix deploy release workflow for multiarch macos builds

ZoweZilsio:


Crementif:
Could you maybe explain what this change does? Does this make it so that our CI will now also add an arm build to the artifacts?

RedBlackAka:
@Crementif While commit #1255 introduced support for the coexistence of x64 and ARM64 builds, it missed updating `deploy_release.yml`, causing the macOS x64 artifact to not being able to be collected (CI deploy currently fails when trying to build a full release). This PR fixes that and makes sure both x64 and ARM64 builds are included in releases.

ZoweZilsio:
> Could you maybe explain what this change does? Does this make it so that our CI will now also add an arm build to the artifacts?

Yes, it will also add the arm builds to the releases. And as @RedBlackAka said, it also fixes the deploy release for x64 macos builds since the artifacts name was changed from `cemu-bin-macos-x64` to `cemu-bin-macos-x86_64`

PR #1643, opened 2025-07-13
UI: fix static initialization

oltolm:
#1519 has problems with static initialization. It happens to work sometimes with some compilers, but my builds simply crash at start.

I fixed it by doing static initialization on demand.

### DIFF
--- a/src/gui/wxgui/input/HotkeySettings.cpp
+++ b/src/gui/wxgui/input/HotkeySettings.cpp
@@ -112,13 +112,7 @@ std::optional<std::string> SaveScreenshot(std::vector<uint8> data, int width, in
 }
 
 extern WindowSystem::WindowInfo g_window_info;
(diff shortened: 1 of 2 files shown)


Crementif:
Just curious, but what was the compiler and compiler version you used to compile Cemu with? Is it a memory-usage related issue?

oltolm:
I used Clang 20 for Windows. I think the problem is https://isocpp.org/wiki/faq/ctors#static-init-order. Basically the order of static initialization is undefined if it happens in multiple compilation units.

Exzap:
Preferable solution is to initialize both maps at runtime in `HotkeySettings::Init`, it's more compact, doesn't hurt compile time or performance due to all the encapsulation and static accessors (even if the cost is negligible), is less LoC and it seems to be more in line with what the original author intended (since `s_keyboardHotkeyToFuncMap` is already initialized in `Init()`).

oltolm:
done

Exzap:
thanks!

Issue #1491, opened 2025-02-06
Add Windows installer


RedBlackAka:
### Your suggestion

Add an installer for Windows, providing a more integrated feel and having a proper place for the application to be!

Exzap:
There are no plans to offer an installer version of Cemu

PR #1645, opened 2025-07-15
Add NSIS Windows installer

RedBlackAka:
Addition of an [NSIS](https://nsis.sourceforge.io/Main_Page) Windows installer to Cemu's build and distribution. Addresses #1491

Installers are more familiar to many Windows users and have the following benefits:
- Automatically creates shortcuts on the Desktop and in the Start Menu
- Appearance in Apps and Programs, and uninstallation through it
  - Cleans registry leftovers
  - User files in `AppData\Roaming` remain untouched
- Gives the application a dedicated default path in `AppData\Local`, rather than just leaving it float in Downloads or Documents, which is true for most users
  - For inexperienced users, this can avoid problems caused by OneDrive syncing
- Automatically assigns Rom files (`.wud, .wux, .wua`) to be opened with Cemu
- Portable single-file installer for easy distribution and offline updates

The `.nsi` file is based upon the one from [Azahar](https://github.com/azahar-emu/azahar) (originally taken from Dolphin), which I have touched-up recently.

Tested and verified as working.

I have also adjusted spacing for consistency in the build files.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -67,14 +67,14 @@ jobs:
     - name: "cmake"
       run: |
         cmake -S . -B build ${{ env.BUILD_FLAGS }} -DCMAKE_BUILD_TYPE=${{ env.BUILD_MODE }} -DCMAKE_C_COMPILER=/usr/bin/clang-15 -DCMAKE_CXX_COMPILER=/usr/bin/clang++-15 -G Ninja -DCMAKE_MAKE_PROGRAM=/usr/bin/ninja
-        
+
     - name: "Build Cemu"
       run: |
         cmake --build build
-        
+
     - name: Prepare artifact
       run: mv bin/Cemu_release bin/Cemu
-              
+
     - name: Upload artifact
       uses: actions/upload-artifact@v4
       with:
@@ -138,7 +138,7 @@ jobs:
     - name: "Bootstrap vcpkg"
       run: |
         ./dependencies/vcpkg/bootstrap-vcpkg.bat
-        
+
     - name: 'Setup NuGet Credentials for vcpkg'
       shell: 'bash'
       run: |
@@ -152,7 +152,7 @@ jobs:
         `./dependencies/vcpkg/vcpkg.exe fetch nuget | tail -n 1` \
         setapikey "${{ secrets.GITHUB_TOKEN }}" \
         -source "https://nuget.pkg.github.com/${{ github.repository_owner }}/index.json"
-        
+
     - name: "cmake"
       run: |
         mkdir -p build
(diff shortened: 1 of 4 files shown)


Exzap:
One question, I assume this causes problems with the auto-updater in Cemu since by default there is no write access to `C:\Program Files`? How would we handle that?

> The .nsi file is based upon the one from [Azahar](https://github.com/azahar-emu/azahar) (originally taken from Dolphin), which I have touched-up recently.

The original file is GPLv2 licensed which is incompatible with our MPL-2.0 license. I assume you are not the original author (or at least not the only one), so there are multiple ways to solve this:
- Rewrite the file from scratch without using GPL licensed code
- Ask the original author(s) to grant us a special exemption
- Ask the original author(s) to dual license the file under GPLv2 and MPL-2.0 (or anything less restrictive and non copyleft would also be fine, e.g. MIT or BSD)

RedBlackAka:
> One question, I assume this causes problems with the auto-updater in Cemu since by default there is no write access to `C:\Program Files`? How would we handle that?

Overlooked that, that is indeed an issue and updates do not progress without opening as administrator. The already running process cannot be elevated, so to do it smoothly, having it in Program Files would require splitting out the updater into a separate exe that would be requested to run as administrator, or making the updater detect if the installed version is used and instead download and run the newest installer. The installer gives the choice of both installing system-wide (`Program Files`) or just for the user (`AppData\Local`).  So the simplest solution for now would be to just cut system-wide. Would potentially see how to solve system-wide later then (if you permit larger changes to the updater).

In regard to license, I have now asked in Dolphin and Azahar for permission of use.
However noticing this:
> Cemu is licensed under [Mozilla Public License 2.0](https://github.com/cemu-project/Cemu/blob/main/LICENSE.txt). Exempt from this are all files in the dependencies directory for which the licenses of the original code apply as well as **some individual files in the src folder, as specified in those file headers respectively**.

Can I simply use the original GPL-2 license for this file?

Exzap:
> Can I simply use the original GPL-2 license for this file?

Short answer is no, we cannot include a GPL-2 source file. Some other licenses are compatible (like BSD and MIT) hence the paragraph you quoted, but GPL is very strict in it's copyleft requirements.

RedBlackAka:
I have contacted the earliest documented author `delroth`, who gave me full permission to use their parts under any open license, plus made further changes that make me responsible for most of the modern file (I give permission for licensing under MPL-2.0 lol). So licensing should be fixed now.

Installer is changed to be user-only, and I now actually don't see any need for a system-wide one in this case. Also made several improvements and cleanups, fixed things I missed, plus mentioned in the README that it is available as both installer and portable.

Did some more throughout testing with it and would say that it is in a mature state!
Try it for yourself here:
[cemu-installer-windows-x64.zip](https://github.com/user-attachments/files/21308072/cemu-installer-windows-x64.zip)

RedBlackAka:
Bump @Exzap ?

Exzap:
Sorry for the delay, pretty swamped the last few weeks so Cemu work had to take a backseat. But this PR is fine anyway so lets just merge it. This is a great addition so thanks for the contribution.

PR #1647, opened 2025-07-20
UI: Upgrade to wxWidgets 3.3.0 and add dark mode support for Windows

Crementif:
**wxWidgets 3.3.0 recently released, and with it came the long-awaited dark mode support for Microsoft's native controls that wxWidgets uses to render most of Cemu's GUI on Windows bar some custom controls. This updates us to that version, adds a new option to the general settings menu and fixes many other dark mode related bugs.**

### Technical Changes
This PR changes the following:
* Updates Cemu to wxWidgets 3.3.0 by introducing a new vcpkg overlay port while it gets an official port which is likely to happen when 3.3.0 is considered stable enough to replace the wxWidgets 3.2.7 port.
* Fixes for any of the breaking changes that came with wxWidgets 3.3.0
* Improves a lot of other windows in Cemu that never had dark mode added to them.
  * Hardcoded-colors were with the OS-provided colors whenever its close enough, but some cases required a custom color palette.
  * You can see the list of commits to see the various controls that were improved specifically.
* Old legacy code was removed in favor of built-in support. For example, various windows now render native checkboxes instead of relying on packaged non-themed icon images, removing the custom CheckedListCtrl Cemu shipped with in favor of wxListCtrl, and various other bugs and improvements.
* There's some quirks in wxWidgets' ListCtrl that made the game list icons offset. This is due to a mysterious [+6 pixel offset that was added/changed in 3.3.0](https://github.com/wxWidgets/wxWidgets/commit/0ca1c14df0f81814615b425a61301e0c88d307e6), which I've yet to report upstream. However, in the meantime I've made a patch for our vcpkg port overlay https://github.com/cemu-project/Cemu/commit/40302a706b1387fa293fa18f2a9f51111f87d455 that patches this cleanly enough to be shippable in Cemu at least. Here's a before photo (it also affected the light mode theme):
image

---
### Screenshots
Here's also some new screenshots of the Windows UI in dark mode.

image

image

image

image

image


goeiecool9999:
This is everything I've found on linux so far:
There are some visible differences on my arch installation. I use the adwaita dark theme which is probably pretty common. 

# about window
As you can see there's a ton of space in between the parentheses for the links and they're not at the same height.
| before             |  after |
|---------------|----------------|
|image|image|

# main window
I didn't like the added vertical lines at first but this may just take some getting used to.
## before:
image

## after:
Screenshot_2025-07-20_11-40-43

# F11 bug

When launching a game in fullscreen and pressing F11 the menu bar does not reappear in the main window.
When launching a game in windowed mode if you then activate fullscreen mode from the "Options" dropdown menu the checkbox stays checked when you exit fullscreen with F11. Meaning you have to click it twice to re-enter fullscreen.

goeiecool9999:
The texture cache window and the newly merged hotkey settings window both do not have proper theming.
image
image

PR #1649, opened 2025-07-23
UI: replace wxCheckTree with wxTreeListViewCtrl

oltolm:
I wanted to get rid of the non-standard `wxCheckTree` in the "Graphic packs" window. I replaced it with a `wxTreeListViewCtrl`. I think that it looks better because you now have three states checkboxes and you can see where there are active options. The only thing that was lost are colors.

graphic packs

### DIFF
--- a/src/gui/wxgui/GraphicPacksWindow2.cpp
+++ b/src/gui/wxgui/GraphicPacksWindow2.cpp
@@ -3,7 +3,6 @@
 #include "wxgui/DownloadGraphicPacksWindow.h"
 #include "Cafe/GraphicPack/GraphicPack2.h"
 #include "config/CemuConfig.h"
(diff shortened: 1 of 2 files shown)


Crementif:
I like the idea to remove some of the antiquated custom controls with new features that are provided by wxWidgets, but I think not having the colors is a pretty big downgrade in terms of usability and clarity. I don't think the middle state checkboxes really add anything, and I don't really think we should have users be able to mass-enable ALL graphic packs for a game without knowing what they're enabling. Unless it's possible to have the colors too, I don't think it's worth replacing our custom solution. But up to Exzap whether he agrees with me.

oltolm:
You can not mass-enable options. All inner nodes are read-only. You can only change leafs.

PR #1700, opened 2025-10-08
Enable use of SSE4.2 instructions

RedBlackAka:
Enables the use of SSE4.2 instructions for the entire project, enabling the compiler to make optimizations with them wherever it sees fit and bringing potential performance improvements for x86_64 builds. Also adds an option to turn this optimization off.

Makes processors with the SSE4.2 extension the official minimum requirement, which I consider reasonable. Looking through the assembly, it currently only makes further use of up to SSE4.1 instructions, but MSVC only offers a flag for 4.2.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -240,6 +240,7 @@ Example usage: `cmake -S . -B build -DCMAKE_BUILD_TYPE=release -DENABLE_SDL=ON -
 | ENABLE_VCPKG       |   | Use VCPKG package manager to obtain dependencies                            | ON      |                    |
 | ENABLE_VULKAN      |   | Enable the Vulkan graphics backend                                          | ON      |                    |
 | ENABLE_WXWIDGETS   |   | Enable wxWidgets UI                                                         | ON      | Currently required |
+| ENABLE_SSE42       |   | Enable SSE4.2 optimizations                                                 | ON      |                    |
 
 ### Windows
 | Flag               | Description                       | Default | Note               |


Exzap:
Any benchmarks?

amayra:
i think AVX support should be a minimum requirement for fewer performance issues reports

Ammar-Sadaoui:
i don't think support 20 years old hardware is a good idea 

plus Windows 11 24H2 requires the CPU to support SSE4.2. Otherwise, the Windows kernel is unbootable

how about making Bulldozer and sandy Bridge as minimum requirement?

RedBlackAka:
> Any benchmarks?

Only did some quick testing, but there seem to be no noticeable improvements during gameplay on high-end hardware. Improvements could show up under specific scenarios or on weaker hardware. I can do some more testing though, if you have any recommendations for intensive games/scenarios?

Looking at the assembly and logs though, auto-vectorization does seem to be very conservative here... But this was a low-hanging fruit and opens up for manual intrinsics writing using those instructions in the future.

> i think AVX support should be a minimum requirement for fewer performance issues reports

AVX is a bit high for a minimum, especially if not specifically used. Intel produced some CPUs without it until 2020. In addition, on AMD processors preceding Zen 2, AVX is implemented via a 128-bit width, tending to actually decrease performance when used.

Exzap:
Makes sense that you don't see a lot (or any) improvement. Most of the code that Cemu runs is JIT translated and thus unaffected by compiler flags and on the GPU thread we already use [handwritten assembly](https://github.com/cemu-project/Cemu/blob/d54fb0ba78ce2ffac3caa399414af19ec1016f05/src/Cafe/HW/Latte/Core/LatteIndices.cpp#L805) where performance matters. In my experience auto-vectorization really only seems to provide benefits when the code isn't well optimized to begin with.

I'm going to close this because I think it's not worth locking out users (even if only a few) for no tangible benefit. Feel free to reopen a new PR if you happen to find any substantial gains.

PR #1706, opened 2025-10-15
Add End emulation and better Recent files menu

RedBlackAka:
Adds (or rather enables) an End emulation button under the File menu, making it possible to exit games without entirely closing Cemu, and made it assignable to a hotkey (set to F5 by default).

Also improves the the recent files list by making it a submenu instead, now being able to hold up to 10 entries.

Small improvements that are more in line with other emulators.

### DIFF
--- a/src/gui/wxgui/MainWindow.cpp
+++ b/src/gui/wxgui/MainWindow.cpp
@@ -677,13 +677,7 @@ void MainWindow::OnFileMenu(wxCommandEvent& event)
 	}
 	else if (menuId == MAINFRAME_MENU_ID_FILE_END_EMULATION)
 	{
-        CafeSystem::ShutdownTitle();
-		DestroyCanvas();
-		m_game_launched = false;
-		RecreateMenu();
-        CreateGameListAndStatusBar();
-        DoLayout();
-		UpdateChildWindowTitleRunningState();
+		EndEmulation();
 	}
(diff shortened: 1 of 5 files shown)


Kiyuja:
fantastic addition. I wanted this for so long, it always felt awkward to restart the entire thing if you just wanted to test something. If this works without any issues it's a massive thumbs up from me!

Exzap:
What you did is add a hotkey for a feature that exists as a skeleton in the code but doesn't work properly...

The fact that the `Stop emulation` menu option is not exposed in release builds should have been a clue that maybe it just doesn't work right. Why would you not question that instead of just blindly enabling it. I really don't get it.

The end emulation feature, as it exists in the code currently and with this PR, will at best work *sometimes* and at worst it can corrupt saves. Even if it doesn't straight out crash, it leaks memory quite heavily, doesn't properly reset a lot of systems and causes lingering bugs down the line. It's just nowhere ready to make it generally available. This PR does nothing to fix any of that.

I will be closing this PR because it is just frankly misleading and will get people's hopes up for a feature that is just not gonna be ready for a while. Feel free to isolate the change to improve the recent files change and submit that as a separate PR.

RedBlackAka:
Ahh I see, did not encounter any crashes or other issues during testing. Will attempt to look more into that

PR #1726, opened 2025-11-19
cmake: fix usages of target_link_libraries

oltolm:
The usage of `target_link_libraries` is a mess. I cleaned it up.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -138,7 +138,6 @@ target_link_libraries(CemuBin PRIVATE
 	CemuGui
 	CemuInput
 	CemuUtil
-	OpenGL::GL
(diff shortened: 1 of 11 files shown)


abouvier:
`zstd::zstd` should still be linked to `CemuCafe` target (`zstd` is used in `src/Cafe/HW/Latte/Common/RegisterSerializer.cpp` and `src/Cafe/HW/Latte/Common/ShaderSerializer.cpp`).

oltolm:
Why does it build without `zstd::zstd`?

abouvier:
It doesn't when you use system libraries everywhere. `zstd::zstd` is probably transitively included by another library when only in-tree libraries are used.

Issue #1672, opened 2025-08-18
Wx 3.3+ behavioral and performance regressions


Squall-Leonhart:
### Current Behavior

The WxGamelist.h specifies hardcoded kListIconWidth and kIconWidth values which previously behaved fine under Wx 3.2.x but cause a high delay (cpu choke) on loading the emulator at high DPI values when built against 3.3+, a secondary consequence of this change in scaling behavior is that the icon overlaps the name text on dark mode, and is only partially displayed on light mode.

In testing fixed values, i found that 26 and 51 (down from 64 and 128) achieved the expected icon size at 250% dpi, resolving both of these issues, this is obviously unfeasible as the icon width at 100% dpi will be undersized.

With the default values, at 250% dpi, cemu will take its sweet time iterating through the game icons before first displaying them no matter what mode you have the list set to, this can delay readiness by minutes for larger lists as the bitmaps are prepared for display at the scaled size

In starting to look into this, i noticed several other issues.

1. the icon does not resize with the column, like in other apps that have a similar icon displaying field.
2. the column can be resized by double clicking the seperator so it adjusts the icon size, but can not be dragged, and does not persist in the cemu configuration.
3. one of the icon modes can hang permanently because cemu somehow displays clickable blank placeholders with names, before eventually showing the icons finally
4. Small Icon mode is the same size as Icon mode.

Other Wx 3.3 changes include the oversizing of several UI panels, resulting in extensive unused space on them, and List Scrolling has gone to shit (though this one [is a reported, unfixed Wx bug](https://github.com/wxWidgets/wxWidgets/issues/24011))

### Expected Behavior

*

### Steps to Reproduce

*

### System Info (Optional)

Windows 10 / 11, at high dpi's.

PR #1654, opened 2025-07-25
UI: Fix minor dark mode issues

Crementif:
This fixes some very minor issues:
- The placeholders for the game list icons are now black instead of white when using dark mode.
- If you disable an activated graphic pack it'll no longer change the text to black when using dark mode.
- Some other minor fixes for things that already seemed to work in dark mode.

PR #1728, opened 2025-11-21
UI: Fix (MSW-only) dark mode wxGameList issues with upstreamed wxWidgets fix

Crementif:
There were some issues with my initial patch #1654, which did partially solve some issue. However, a larger rework of the msw/listctrl.cpp file was needed so I did I PR'ed https://github.com/wxWidgets/wxWidgets/pull/25950. This makes wxWidgets render the dark mode (which used custom rendering for the most part, unlike the white mode that was drawn by Windows) match the white mode closer, which for Cemu just means that:
* No overdrawn text or game icons onto the next column in that row
* Zero-width rows are hidden, which Cemu uses for hiding columns
* The row sizes match the column widths, specifically for the game icon

It also brings back the old way of setting icons in the game list, so it should be much more performant.

This does mean that we're using a slightly experimental version of wxWidgets. Testing is appreciated if you're using Mac or Linux, in case there's any issues.

Fixes https://github.com/cemu-project/Cemu/issues/1672.

### DIFF
--- a/dependencies/vcpkg_overlay_ports/wxwidgets/portfile.cmake
+++ b/dependencies/vcpkg_overlay_ports/wxwidgets/portfile.cmake
@@ -1,8 +1,8 @@
 vcpkg_from_github(
     OUT_SOURCE_PATH SOURCE_PATH
     REPO wxWidgets/wxWidgets
-    REF "f3da561"
-    SHA512 04e273b83ec683ed18ff5278371441b5ecf5897d34a896307cb59c58fb7adf68a70f6af3b3193bb1c63b9985edd5e6807202c05af3202a54c4142518ed15a71b
+    REF "bfd436b"
+    SHA512 bd3fd6d0d0db3b6fa34eceae1119e21ffd2f62221dcd249f8b8b82a6e65d83a05101e4e1e4ca9b9c4d7937add73b113bb029b03b05d2c3d87d17c1922d800a24
     HEAD_REF master
     PATCHES
         install-layout.patch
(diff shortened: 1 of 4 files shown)


Squall-Leonhart:
theres still drag scrolling stutters in dark mode, but the rest of the issues i listed have been addressed

the alternating colored rows in the dark mode are an inconsistency vs the light mode view though, where there is no apparent alternation between rows (there were in 2.6)

image

image



it may actually be incorporating this alternating colors that makes scrolling sluggish, i do notice if you minimize the window, light mode returns the list on restoring the window almost immediately while dark mode has a measurable delay with a solid grey color
the solid grey is not either of the shades in the dark mode list once it comes up.


Confirmed it, its the alternation that affects the scrolling and initial paint performance where, you can replicate it without fully restarting cemu by opening the Manager viewers - column bar paints white until the rows finish too.

Perhaps the same thing as was done in https://github.com/cemu-project/Cemu/pull/261/commits/26586d00adcb7fdddbdc1ec637111c6786b5de42 needs to be done here.

Crementif:
I've reverted to an earlier version of wxWidgets to prevent the regression you mentioned for the white mode. 

I'm fairly sure that https://github.com/wxWidgets/wxWidgets/pull/25962 has caused a minor regression in this white mode, however I'm unable to replicate it with the sample. I might look into it more later to see if I can find a cause and upstream a fix.

As for the scrolling performance, maybe @Exzap can replicate it being slower with his large game list? But I doubt its slower then Cemu 2.6, since both alternating backgrounds are drawn the same as far as I can tell.

Squall-Leonhart:
its just dark mode thats hurting for scrolling and ui paint performance, open the title manager and you'll see what i mean

Exzap:
Looks like good improvements to me. Thanks! We'll solve the remaining issues as we go

PR #1729, opened 2025-11-22
UI - fix warnings

oltolm:
I fixed compiler warnings and did some trivial refactorings like replacing `Connect` with `Bind` and `list` with `vector`.

### DIFF
--- a/src/gui/wxgui/wxcomponents/checktree.cpp
+++ b/src/gui/wxgui/wxcomponents/checktree.cpp
@@ -121,49 +121,49 @@ void wxCheckTree::Init()
 
 	// Unchecked
 	renderer_dc.SelectObject(unchecked_bmp);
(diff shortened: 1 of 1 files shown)


capitalistspz:
why replace `list` with `vector`?

oltolm:
Efficiency, although it probably does not matter in this case. I can revert.

Exzap:
Looks good. Thanks.

> why replace `list` with `vector`?

Generally speaking, `std::list` is a really inefficient container due to allocating each item individually. The only reason to use it is that it's maybe a tiny bit more convenient to use than a `std::vector` if you have queue-like logic. But it's not [worth the overhead](https://danglingpointers.com/post/dont-use-std-list) and vector performs better most of the time despite having to shift all the elements around for insert/erase. And for larger number of elements you want to use (unordered) map or set anyway.

So personally I just made it a habit to not use it. That said, the code that is changed in this PR is not performance relevant at all so it really makes no difference. The work has already been done so I'll just merge it

PR #1731, opened 2025-11-22
gui: ENABLE_BLUEZ=OFF: bluetooth/bluetooth.h uneeded

mazes-80:
Use case: packaging cemu for gentoo users (WIP)

ENABLE_BLUEZ=OFF should prevent using bluetooth under linux.
However wxgui still expose some:
`#include <bluetooth/bluetooth.h>`

To be able to compile I just used
```
sed -i -e '/input\/PairingDialog/d' src/gui/wxgui/CMakeLists.txt
sed -i -e '/auto\* pairingDialog/,/sizer->Add(pairingDialog/d;' \
       -e '/include.*PairingDialog/d' src/gui/wxgui/input/InputSettings2.cpp
```

This early PR based on local work I did when upgrading package but uses build system logic

### DIFF
--- a/src/gui/wxgui/CMakeLists.txt
+++ b/src/gui/wxgui/CMakeLists.txt
@@ -76,8 +76,6 @@ add_library(CemuWxGui STATIC
   input/InputAPIAddWindow.h
   input/InputSettings2.cpp
   input/InputSettings2.h
-  input/PairingDialog.cpp
-  input/PairingDialog.h
   input/panels/ClassicControllerInputPanel.cpp
   input/panels/ClassicControllerInputPanel.h
   input/panels/InputPanel.cpp
@@ -117,6 +115,13 @@ add_library(CemuWxGui STATIC
   wxgui.h
   wxHelper.h
 )
+
+if (ENABLE_BLUEZ)
+  add_library(CemuWxGui STATIC
+    input/PairingDialog.cpp
+    input/PairingDialog.h
+  )
+endif()
 
 set_property(TARGET CemuWxGui PROPERTY MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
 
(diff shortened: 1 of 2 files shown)


src/gui/wxgui/CMakeLists.txt
```diff
@@ -118,6 +116,13 @@ add_library(CemuWxGui STATIC
   wxHelper.h
 )
 
+if (ENABLE_BLUEZ)
+  add_library(CemuWxGui STATIC
```
Exzap: This should be `target_sources(CemuWxGui PRIVATE`
Currently it breaks the build because add_library tries to create a new library with the same name

Squall-Leonhart:
This caused the button to disappear on windows builds.

> ENABLE_BLUEZ=OFF should prevent using bluetooth under linux.

no it shouldn't, all it should determine is if the pairing process runs(with bluez) or if the user is told to use their system bluetooth manager.

Instead you broke the button entirely on all supported platforms, including those without bluez from doing its alternative intended job, which is raise a message saying "Please use your system's Bluetooth manager instead."

Exzap:
I had to undo these changes for a future commit because of the reasons Squall mentioned. The pairing dialog should be available on Windows but there is also already fallback code in the dialog for when bluez is not available, so excluding it from build is the wrong path. Most likely there is just an issue where some specific part of the pairing dialog isn't properly gated behind `#ifdef HAS_BLUEZ` but I would need to know the actual errors you ran into to be able to fix it

mazes-80:
> Most likely there is just an issue where some specific part of the pairing dialog isn't properly gated behind `#ifdef HAS_BLUEZ` but I would need to know the actual errors you ran into to be able to fix it

This header was still included in "wxgui/input/PairingDialog.h" even ENABLE_BLUEZ is off:
`#include <bluetooth/bluetooth.h>` 

So I removed the pairing dialog from gui.
However it was not intended to be removed from every build but only when under linux and bluez not available.
I assumed wxWidgets was just used under linux so only bluez test was required

So the test I meant would rather be something like:
#if not defined(__linux__) or (defined(__linux) and HAS_BLUEZ)

My distro still does not provide wxwidget-3.3, so I'm stuck at the commit before requiring it.
Until wxwidget 3.3 available on my system I will not test  latest sources.

PR #1741, opened 2025-12-01
docs: Follow standard markdow formating.

LukeFrankio:


Exzap:
Making the table of contents larger is probably not the right direction. It's already hard to parse

PR #1747, opened 2025-12-18
Functioning Miiverse applet

ariankordi:
I've added the missing functions from nn_olv that Miiverse needs to boot (the "PortalAppParam" methods specific to the app). First commits are borrowed from @jonbarrow['s branch](https://github.com/PretendoNetwork/Cemu/commits/main/).

That lets the app boot without crashing, but I also added two code patches needed for Miiverse clones to work:
* Bypass the internal allowlist in wave.rpx

This is why the applet will not even attempt to load from non-Nintendo domains, staying stuck on the loading screen.
Inkay [gets around this](https://github.com/PretendoNetwork/Inkay/blob/b7c86c9edcf99ab6a8e1fc16fc2770a973b68fa6/src/patches/olv_applet.cpp#L33-L52) by replacing the .nintendo.net entry in the allowlist. But, I was able to find a function responsible for validating this and patched it from Cemu to completely bypass it.
Only tested with the USA titles, though. I noticed eShop shared the same function code, so it should work there too, and so will Account Settings if patched manually. I put a note on where to find the function in GamePatches.cpp.

* Pass CA cert validation

Otherwise, the browser shows a generic error message. I added the same patch done by [PretendoNetwork/Meowth](https://github.com/PretendoNetwork/Meowth/blob/dbcb813ca7277b6b226d38fbeef31dd784ff3437/src/patcher/patches/webkit_applets.cpp#L24-L35) as well, applied only if SSL verification is off.
(Notably, *.pretendo.cc or any other Cloudflare domain will work without this, as they use a GTS CA that's already trusted on the Wii U.)

Now, Miiverse successfully loads on Pretendo without nn_olv.rpl or any other patching :)

image

Feel free to suggest feedback for any part of this, or if we should split the patches out of Cemu or anything like that. Thanks.

### DIFF
--- a/src/Cafe/GamePatch.cpp
+++ b/src/Cafe/GamePatch.cpp
@@ -2,6 +2,7 @@
 #include "Cafe/OS/RPL/rpl.h"
 #include "Cafe/HW/Espresso/Interpreter/PPCInterpreterInternal.h"
 #include "CafeSystem.h"
+#include "config/ActiveSettings.h" // Selectively add some patches based on network settings.
 
 void hleExport_breathOfTheWild_busyLoop(PPCInterpreter_t* hCPU)
 {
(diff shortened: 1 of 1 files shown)


jonbarrow:
🙏 I've been wanting to take a stab at this again at some point but never found the time to sit back down with it, thank you for finishing up my experiment 🙌

Exzap:
Using GamePatches based on the network config is a sensible approach for this. Great contribution, thank you!

cool-guy-awesome:
LETS GO

PR #1754, opened 2025-12-23
UI: Adjust title & description and cleanups, end emulation

RedBlackAka:
This includes small miscellaneous adjustments aimed at cleaning up both files and UI presentation, in addition to handling more stuff for end emulation.

- Simplify window title by removing the title ID, I feel like this is unnecessary there and that the title was getting a bit too long, and add Metal to API display
Screenshot 2025-12-23 at 04 39 43

- Add "Cemu" to the Windows .exe file description, as this is used for Task Manager and currently just displays Cemu as "Wii U emulator"

- Some file and spacing cleanups

- Handle more things around end emulation and add shortcut, not user-facing. Did a bit of a bit of experimentation, but it still has issues. I am not familiar enough with the renderer and resource management to complete it and have given up for now

I am also considering doing other changes, such as also consolidating the NFC menu into another submenu and other UI cleanups. Would love some input, but am done for now

### DIFF
--- a/src/gui/wxgui/wxWindowSystem.cpp
+++ b/src/gui/wxgui/wxWindowSystem.cpp
@@ -121,7 +121,8 @@ void WindowSystem::UpdateWindowTitles(bool isIdle, bool isLoading, double fps)
 	else if (LatteGPUState.glVendor == GLVENDOR_APPLE)
 		graphicMode = "[Apple GPU]";
 
-	windowText.append(fmt::format(" - FPS: {:.2f} {} {}", (double)fps, renderer, graphicMode));
+	const uint64 titleId = CafeSystem::GetForegroundTitleId();
+	windowText.append(fmt::format(" - FPS: {:.2f} {} {} [TitleId: {:08x}-{:08x}]", (double)fps, renderer, graphicMode, (uint32)(titleId >> 32), (uint32)(titleId & 0xFFFFFFFF)));
 
 	if (ActiveSettings::IsOnlineEnabled())
 	{


src/gui/wxgui/MainWindow.cpp
```diff
@@ -1771,6 +1760,42 @@ void MainWindow::SetFullScreen(bool state)
...
+	// close memory searcher if created
+	if (m_toolWindow)
+	{
+		m_toolWindow->Close();
+		m_toolWindow = nullptr;
+		m_memorySearcherMenuItem->Enable(false);
+	}
+
+	// unsure about this, but seems to free up additional 10 MB memory, see LatteThread.cpp
+	if (g_renderer)
```
Exzap: Cleaning up the renderer is already handled by `LatteThread_Exit()` which in turn is triggered by `Latte_Stop()` in `ShutdownTitle()`

Exzap:
Please don't move around UI options unless it's a significant and objective improvement. It leads to "Where did the CPU timer go" bug reports because people expect it to be in the place it always has been in. Dropping the GPU suffix is also not really an improvement. `[Intel]` for example is very much ambiguous.

Rest of PR looks good if you also change the renderer cleanup mentioned in my other comment.

oltolm:
```
[build] C:/src/Cemu/src/gui/wxgui/MainWindow.cpp:2253:55: error: expected ';' after expression
[build]  2253 |                 m_fileMenuSeparator1 = m_fileMenu->AppendSeparator()
[build]       |                                                                     ^
[build]       |                                                                     ;
[build] 1 error generated.
```

PR #1775, opened 2026-01-09
nim: Fix memory leak

goeiecool9999:
Store NIMPackage structs in vector instead of pointers on the heap (which weren't freed before)


Exzap:
Looks good to me. CI is failing due to an unrelated problem.

PR #1780, opened 2026-01-12
[draft] Refactor AllocateMemory and FreeMemory functions

rcaridade145:
This adds code to make sure size is also page aligned. Free Memory refactored also to take into consideration the possibly different start that could be calculated in AllocateMemory.

### DIFF
--- a/src/util/MemMapper/MemMapperUnix.cpp
+++ b/src/util/MemMapper/MemMapperUnix.cpp
@@ -40,31 +40,57 @@ namespace MemMapper
 		munmap(baseAddr, size);
 	}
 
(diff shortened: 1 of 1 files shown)


Exzap:
Since MemMapper's purpose is to allocate pages directly, like when setting up Wii U memory space. It would be better if it asserts on unaligned base or size instead of silently trying to adjust those values. If any calling code passes invalid size we should fix that instead.

rcaridade145:
The adapter to the base was added on previous PR to add M1 support by @exverge-0 .
This adds adapt to size and adds the same logic of adapt to base to free

Exzap:
It was merged as part of larger PR and it already was a concern back then. But since this is a refactor PR it should not double down on workarounds.

rcaridade145:
No really improvement found. Closing it.

PR #1781, opened 2026-01-13
Backend: Exclude more Metal only logic on non-Apple platforms and cleanups

RedBlackAka:
Exclude even more Metal only logic on non-Apple platforms, building on #1753 where I missed some. Plus some cleanups. Reduces e.g. Windows binary size by 8 KB! (lol)

Tested on all platforms to verify that nothing accidentally broke.


Exzap:
I am not sure if this level of separation is worth it. `#if` have the downside of making the code harder to read. Some conditionals where it checks the API at runtime are on the hot path but the performance impact is still negligible and as you said it only saves 8KB of binary size. So there are no real gains here but it makes the code a lot less readable.

I would be interested in @SamoZ256's opinion

SamoZ256:
> I am not sure if this level of separation is worth it. `#if` have the downside of making the code harder to read. Some conditionals where it checks the API at runtime are on the hot path but the performance impact is still negligible and as you said it only saves 8KB of binary size. So there are no real gains here but it makes the code a lot less readable.
> 
> 
> 
> I would be interested in @SamoZ256's opinion

I was also unsure where to use #if and where not. Checking for both if metal is enabled and if it is actually used in the same place is pretty unreadable, and even more so if other APIs don't get the same treatment. Imo I'd just leave it without those #ifs whenever possible, especially since there is no significant disadvantage to it.

RedBlackAka:
> > I am not sure if this level of separation is worth it. `#if` have the downside of making the code harder to read. Some conditionals where it checks the API at runtime are on the hot path but the performance impact is still negligible and as you said it only saves 8KB of binary size. So there are no real gains here but it makes the code a lot less readable.
> > I would be interested in @SamoZ256's opinion
> 
> I was also unsure where to use #if and where not. Checking for both if metal is enabled and if it is actually used in the same place is pretty unreadable, and even more so if other APIs don't get the same treatment. Imo I'd just leave it without those #ifs whenever possible, especially since there is no significant disadvantage to it.

Yeah I get that. Am very centered about not shipping unnecessary things, so would at least like to keep it for the includes and the hot code paths. Will look into getting rid of it for the insignificant things

RedBlackAka:
Apologies for the spam, a lot of dependencies to take into account. But this should be better now

Exzap:
No worries about the spamming. We squash PRs before merging so you can experiment as much as you want during PR creation.

I think it's still a little too many checks. Especially in places where `#if` and runtime `if` come together to both check for Metal. But rather than nitpicking this to death lets just merge and next time someone works on the code they can refine it.

PR #1783, opened 2026-01-14
Metal: Create temporary autorelease pools

SamoZ256:
Some Metal objects (like command buffers and encoders) are (annoyingly) autoreleased. This PR wraps such objects in temporary autorelease pools to prevent crashes when the global pool is drained.


Squall-Leonhart:
we found this out the hard way when a metal renderer was added to VBAM, started getting users joining our discord posting about the emulator using GB's of memory, where anywhere else it was less than a GB.

SamoZ256:
> we found this out the hard way when a metal renderer was added to VBAM, started getting users joining our discord posting about the emulator using GB's of memory, where anywhere else it was less than a GB.

And it's extremely hard to debug as well, since it's technically not a memory leak. Also, the memory usage grows a lot faster when Metal frame capture is enabled (either manually or when running through Xcode).

PR #1784, opened 2026-01-14
[Test] CI: Codesign macOS app

shinra-electric:
Currently the macOS app bundles are not codesigned. This is not an issue for the x86 build, as all x86 apps can launch fine without a code signature. 

However, the Arm builds do require a signature. The linker signs it at compilation time, but the changes made to the app bundle afterwards invalidates it. 

This PR will add a signature step at the end of setting up the app bundle and before it is packaged in a `.dmg`. 

Testing: 

- [ ] Ensure the Arm build launches without requiring `xattr`
- [ ] Ensure the x86 build still launches fine (no changes expected)


shinra-electric:
Closing this for now because although the codesigning worked, there is still some other issue that prevents the app from launching properly.

PR #1795, opened 2026-01-27
CI: Fix macOS Arm builds not being permitted to open

shinra-electric:
On macOS all Arm apps that are quarantined (i.e. downloaded from the internet) need at least a basic ad-hoc signature, or else the OS will refuse to open it. (This is not required for x86 apps, and can be worked around by using `xattr -cr ./Cemu.app`). 

While adding an ad-hoc signature I found that a bug when setting the rpath for `libusb` was causing the app to be reported as broken even if the signature was fine. With the signature and this fix, the Arm app bundle now launches correctly when download from the internet.

### Main: 
 
Screenshot 2026-01-27 at 13 05 00

### PR: 
Screenshot 2026-01-27 at 13 10 58

Note: I get the option to open because I have set Gatekeeper to allow apps from anywhere. Regular users will still have to go to the security settings to allow the app to run. This occurs on first run only. 

### Changes that fix the issue: 
- Adding a codesigning step
- Fix the rpath setting for `libusb`

### Other changes made: 
- Set the output name to `Cemu` instead of `Cemu_release` only when the build type is Release. This prevents lots of renaming and modifying the info.plist after building, which invalidates the signature. (Affects other platforms, doesn't affect debug builds)
- Change bundle name in info.plist to output name instead of hard-coding to Cemu
- Move the chmod permissions step for the update script to be done by cmake
- Add entitlements file. Not strictly required, but it is standard practice for modern macOS apps
- Remove ninja as a dependency since it it already installed. Silences a warning
- Remove renaming steps for Linux and Windows as well
- Dynamically add the current year to the copyright field in info.plist

### Testing: 
- Ensure the GitHub Actions builds launch on Apple Silicon Macs (without using `xattr`)
- Since touching the output name will affect other platforms, ensure they are building correctly


JaimieVandenbergh:
Just as a background FYI: libusb path has been fixed at least four times that I know of in the life of the macOS port :) Thanks for taking it on again!

RedBlackAka:
While true for some apps, I have never experienced Cemu being reported as broken, regardless of x86 or ARM. Also the builds produced by this PR still do not allow me to directly open the application, still needing to go through the security step. The only difference is that it now says that the app needs an update, instead of it being unable to be checked for malware (does not affect the end result). Ventura 13.7.8

shinra-electric:
> While true for some apps, I have never experienced Cemu being reported as broken, regardless of x86 or ARM. 

For Arm apps that are quarantined, the a hash of the app bundle is compared with the signature when first launched (an ad-hoc signature is basically just a hash). If the signature doesn't match, it will say the app has been damaged. Signatures are not required for x86 apps, so the signing step would not make any difference to them. 

However the changing of the path for `libusb` should affect x86 apps as well. 

> Also the builds produced by this PR still do not allow me to directly open the application, still needing to go through the security step.

This is normal and will need to be done by most users on first launch anyway. The only way to avoid this is to get the app notarised by Apple, which required a paid developer account. It is not worth paying for just to remove the inconvenience of approving the app in the security settings. (You can also set Gatekeeper to [allow apps from anywhere](https://macreports.com/how-to-disable-gatekeeper-on-mac-and-enable-the-anywhere-option-for-installing-any-software/), which is what I have done).

>  The only difference is that it now says that the app needs an update, instead of it being unable to be checked for malware (does not affect the end result). Ventura 13.7.8

This is unusual. That type of message usually shows when trying to run a 32-bit x86 app on Catalina or later that doesn't support them. Can you provide further details? I have tested on both Arm and Intel Macs but haven't seen this message.

RedBlackAka:
Huh. This is the exact message with the changes:
Screenshot 2026-01-29 at 13 00 23
This does not actually have any effect on behavior as said.

Also, even for the ARM builds, the broken message has never been an issue for me here on Cemu. None of the builds have ever needed xattr

Setup:
MacBook Air M1
Ventura 13.7.8
Cemu aa37d0b / ARM64
Compared to Cemu 2.6 and latest main

shinra-electric:
Will close this and separate out the changes into new PRs

PR #1805, opened 2026-02-07
GraphicPack: Support universal patches

Luminyx1:
Implemented new features for graphic packs to enable creation of game-agnostic packs:

- New `*` wildcard value for rules.txt `titleIds` -> Effectively disables title ID matching for that graphic pack and makes it check module hashes on every title
- New `rpx` wildcard value for `moduleMatches` in graphicpacks assembly files -> Patch will apply to the main RPX of whatever titles are matched by rules.txt titleIds
- Combining both of the above you can have a universal graphic pack which applies to any game, but only the game itself, without clobbering system RPLs, this can be used for example by dynamic mod loaders.
- New `.callback entry <symbol>` directive in graphicpacks assembly files, allowing graphicpacks to register labels/functions to be called before the launched game starts. (Only in games which that graphicpack matched to be active)
  - Analogous to `ON_APPLICATION_START` in Aroma plugins

### DIFF
--- a/src/Cafe/GraphicPack/GraphicPack2.cpp
+++ b/src/Cafe/GraphicPack/GraphicPack2.cpp
@@ -1199,15 +1199,14 @@ std::vector<uint64> GraphicPack2::ParseTitleIds(IniParser& rules, const char* op
 	if (!option_text)
 		return result;
 
+	if (*option_text == "*")
+	{
+		m_universal = true;
+		return result;
+	}
+
 	for (auto& token : TokenizeView(*option_text, ','))
 	{
-		if (token == "*")
-		{
-			m_universal = true;
-			result.clear();
-			break;
-		}
-
 		try
 		{
 			result.emplace_back(ConvertString<uint64>(token, 16));
--- a/src/Cafe/GraphicPack/GraphicPack2Patches.cpp
+++ b/src/Cafe/GraphicPack/GraphicPack2Patches.cpp
@@ -172,7 +172,7 @@ void GraphicPack2::ApplyPatchesForModule(const RPLModule* rpl)
 	std::vector<PatchGroup*> list_groups;
 	for (auto itr : list_patchGroups)
 	{
-		if (itr->matchesCRC(rpl->patchCRC) || (itr->m_isRpxOnlyTarget && rpl->fileInfo.flags & 2))
+		if (itr->matchesCRC(rpl->patchCRC) || (itr->m_isRpxOnlyTarget && rpl->IsRPX()))
 			list_groups.emplace_back(itr);
 	}
 	// apply all groups at once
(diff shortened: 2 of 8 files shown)


mrdc:
I hope it will enable more games to have increased resolutions selectable.

src/Cafe/GraphicPack/GraphicPack2.cpp
```diff
@@ -1198,6 +1201,13 @@ std::vector<uint64> GraphicPack2::ParseTitleIds(IniParser& rules, const char* op
 
 	for (auto& token : TokenizeView(*option_text, ','))
 	{
+		if (token == "*")
```
Exzap: I recommend moving the check out of the loop. Right now it allows expressions like `0005000010143400,*` which doesn't make a lot of sense. It should allow either a list of titleIds or just `*`.

src/Cafe/GraphicPack/GraphicPack2Patches.cpp
```diff
@@ -188,7 +188,7 @@ void GraphicPack2::RevertPatchesForModule(const RPLModule* rpl)
 	std::vector<PatchGroup*> list_groups;
 	for (auto itr : list_patchGroups)
 	{
-		if (itr->matchesCRC(rpl->patchCRC))
+		if (itr->matchesCRC(rpl->patchCRC) || (itr->m_isRpxOnlyTarget && rpl->fileInfo.flags & 2))
```
Exzap: same here

src/Cafe/GraphicPack/GraphicPack2PatchesParser.cpp
```diff
@@ -425,7 +431,23 @@ bool GraphicPack2::ParseCemuPatchesTxtInternal(MemStreamReader& patchesStream)
 			}
 			continue;
 		}
-		
+		else if (parser.matchWordI(".callback"))
```
Exzap: .callback is not really specific enough. What if we want to add other callbacks in the future? If you are up for it you could make this a command with a type field, as an example: `.callback entrypoint <symbol>`
Otherwise the simpler solution is to rename it to a more descriptive `.entrypointcallback` (not very concise, feel free to think of something better if you choose this route)

Similarly this applies to `list_callbacks`.

src/Cafe/GraphicPack/GraphicPack2Patches.cpp
```diff
@@ -172,7 +172,7 @@ void GraphicPack2::ApplyPatchesForModule(const RPLModule* rpl)
...
```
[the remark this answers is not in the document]
Luminyx1: Added method `IsRPX()`

Exzap:
Looks good to me now. Thanks!

PR #1808, opened 2026-02-09
Changing StartupWMClass

Kyuyrii:
This is to use the correct StartupWMClass.


Exzap:
Our WM_CLASS should be `Cemu` since that's what we are setting via wxWidgets. Is this an issue with flatpak builds specifically?

Kyuyrii:
It seems to be because of Wayland, and regarding Flatpak, the defined WMClass is info.cemu.Cemu.

Kyuyrii:
I tested the AppImage version; it uses the WMClass 'Cemu' instead of the WMClass 'info.cemu.Cemu' from the Flatpak version.

PR #1813, opened 2026-02-16
Few small fixes for Graphics packs, Game Profiles and Logging

Squall-Leonhart:
Wrap graphics pack text such as descriptions, updating the text wrap as dragged.
use DIP to make the window sizes  work properly at high DPI, the current behavior at DPI's above 100% requires that the user manually resize them to show all controls.

Before @ 250%

image

After

image
image
image

Wrapping works on long description text, somewhat on drop lists when filtered, or graphics packs with long option value friendly names aren't present
image


emiyl:
macOS UI doesn't look quite right :(

The top bar and pane on the right are not the right colour

image

Text is cut off past the first line,

image

Until I resize the window

image

PR #1815, opened 2026-02-22
Camera

capitalistspz:
Couldn't merge properly, so here's a new branch.
Mostly the same as the other one.


Exzap (changes requested):
Using std mutex will clash with Cafe OS primitives.
Any function that can block or resume a thread (OSJoinThread, OSSignalEvent, OSResumeThread, OSSleepTicks, OSLockMutex..) can yield the current fiber and start running other PPC threads.

Take this example:
```
// part of a HLE function
OSSleepTicks(ONE_SECOND);
```
Instead of blocking the host thread for one second, coreinit threading will instead suspend the host fiber (or ucontext in posix terms) assigned to the current PPC thread. Then the PPC scheduler picks the next PPC thread to run and continues doing work. Eventually after the sleep timeout occurs the scheduler will switch back and resume the fiber from within OSSleepTicks. From the code's perspective the 1 second delay happened, but the actual CPU emulation thread never actually slept. In multicore emulation it can even be a different host thread that returned from `OSSleepTicks`.

I hope this explanation makes sense. What currently happens is that you carry the std mutex into other PPC threads because you are calling OS functions which can context switch. It can easily lead to deadlocks or race conditions.

The solution is to use `OSMutex` over `std::mutex` in HLE code. `OSMutex` cannot be used outside of PPC contexts, so in that case you have to use `std::mutex` but need to make sure that you are not holding the lock while calling any OS* functions.

Exzap:
~~Add:
`cemu_use_precompiled_header(CemuCamera)`
After add_library~~

Ah nvm you are behind too many commits for that. But this should work:
`target_precompile_headers(CemuCamera PRIVATE precompiled.h)`

capitalistspz:
> ~Add: `cemu_use_precompiled_header(CemuCamera)` After add_library~
> 
> Ah nvm you are behind too many commits for that. But this should work: `target_precompile_headers(CemuCamera PRIVATE precompiled.h)`

Maybe I should rebase then

capitalistspz:
Because of a bug in SDL (https://github.com/libsdl-org/SDL/issues/15003), cameras created on Linux via `v4l2loopback` (e.g. OBS Virtual Camera) with `exclusive_caps` will not be discovered

capitalistspz:
> Because of a bug in SDL ([libsdl-org/SDL#15003](https://github.com/libsdl-org/SDL/issues/15003)), cameras created on Linux via `v4l2loopback` (e.g. OBS Virtual Camera) with `exclusive_caps` will not be discovered

Should be fixed in SDL 3.4.16

PR #1832, opened 2026-03-13
Fix fmt format spec crash in WPADGetInfoAsync

Squall-Leonhart:
Fix crash in WPADGetInfoAsync where applying {:08x} format spec to a void const* returned by fmt::ptr() causes fmt to throw an invalid format spec exception


Exzap:
Thanks. Printing the raw 64bit host pointer is not that useful, so generally what we try to do for API logging is print the pointer relative to the PowerPC address space.
Can you try this:
Keep the  `{:08x}` and instead change `fmt::ptr(wpadInfo)` to `memory_getVirtualOffsetFromPointer(wpadInfo)`

Squall-Leonhart:
Ah right, forgot about that from the last time i tripped over an fmt issue.

Squall-Leonhart:
Done!

Issue #1382, opened 2024-10-16
Twilight princess HD mini map cut in half


A-randomNerd:
### Current Behavior

When playing Twilight Princess HD the mini-map is cut in half and the main map only half-loads.
<img width="1440" alt="Screenshot 2024-10-16 at 21 34 31" src="https://github.com/user-attachments/assets/0f9da09e-13d2-4a7a-9a40-b81a46823645">
<img width="1440" alt="Screenshot 2024-10-16 at 21 33 39" src="https://github.com/user-attachments/assets/47334abc-88b8-47c4-a0e8-41643f6bd579">

I assumed there was an issue with my setup but I saw someone else had this issue and using different graphics packs/ restarting my device didn't work.
.

### Expected Behavior

the map should appear as normal?. (sorry I'm not sure how to word this correctly)

### Steps to Reproduce

As far as I can tell it happens by just running the game.

### System Info (Optional)

OS: macOS
GPU: Apple M1 GPU (8 core i think)

JaimieVandenbergh:
Known issue; it is due to MoltenVK not supporting geometry shaders yet. There’s no workaround, except to play as-is or not on macOS.

galad87:
https://github.com/cemu-project/Cemu/pull/1287 should already supports geometry shaders, so it will be probably be fixed when and if it's merged.

UsRnM44:
It was actually surprising for me that 2.1 release took care of:

Fixed incorrect decoding of R4G4 texture format when VK_FORMAT_R4G4_UNORM_PACK8 is not supported. Fixes invisible rupee counter in Twilight Princess HD'

but this one left behind . . .

WillsitoGG:
I have the same issue too in latest version of Cemu and Android 13.

indianajson:
Same issue with latest verison of Cemu on MacOS.

CallowayCrusader08:
same issue but im on a m5 pro

5HT2:
Is this fixed if you compile MoltenVK from source?

domitakes:
For anyone landing here: this is caused by MoltenVK not supporting geometry shaders. Cemu emulates the Wii U RECTS primitive (3 vertices, the 4th one is derived) with a geometry shader, so on macOS/Vulkan only one triangle of each rect gets drawn, hence the minimap and the map screen being cut diagonally.

[...]

PR #1838, opened 2026-03-15
Metal: fix broken triangle fan primitive and uninitialized render targets

fabioarnold:
**Heads up**: I used AI/Claude to generate these changes. I have only shallow knowledge of this code base. I do have some experience with graphics APIs though and read over these changes. Feel free to close if AI is not acceptable in this project.

I was excited to try Twilight Princess HD on main with the new Metal backend since Vulkan on macOS has other issues (#1382). And I faced the following glitches mainly during scene transitions:

metal-glitch

map-glitch

I described these issues to Claude and let it fix it for me. With these changes applied I did not experience the glitches anymore. Occasionally there are still geometry glitches only on the title screen. Claude says this is due to async shader compilation.

### DIFF
--- a/src/Cafe/HW/Latte/Renderer/Metal/LatteTextureMtl.h
+++ b/src/Cafe/HW/Latte/Renderer/Metal/LatteTextureMtl.h
@@ -19,6 +19,8 @@ public:
 
 	void AllocateOnHost() override;
 
+	bool m_isInitialized = false;
+
 protected:
 	LatteTextureView* CreateView(Latte::E_DIM dim, Latte::E_GX2SURFFMT format, sint32 firstMip, sint32 mipCount, sint32 firstSlice, sint32 sliceCount) override;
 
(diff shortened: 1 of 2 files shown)


Exzap:
The problem with submitting LLM authored code like this is that you are offloading the bulk of the work (verifying & cleaning up the code) to the maintainers. It's doing maybe 5% of the labor involved in actually contributing fixes. I know there are some developers out there that merge everything on the basis of "It works so it's fine", but code can be fragile, too verbose or just incorrect. Merging bad code always comes back to bite you eventually in the long term.

Just on a quick glance there are some weird and unnecessary changes in this PR. Feel free to clean it up and resubmit it. I think it's fine to use claude to help you author a PR, but more in the sense of asking it questions about the code base and help you understand the problem and how to best fix it. And if you just let it write the code, consider that a first draft and don't submit it as-is. The idea is that you should understand the code your are submitting and the reasoning behind it. Otherwise it's not really of help and we either have to merge a black-box solution (which I personally will never do) or do the bulk of the reviewing+clean up work ourselves which kinda defeats the purpose of PRs.

fabioarnold:
Thanks for writing this up. That's very understandable.

Depending on how bad AI-generated PRs and issues might get it might be worth adding a sentence to the Contributing section of the README. I tried to find information on how this project handles LLM authored code but couldn't find anything.

PR #1839, opened 2026-03-17
nn_ac: Add local IP getter implementation for Linux

ashquarky:
Adds implementation of `_GetLocalIPAndSubnetMask` for Linux. This allows LAN play in NEX games to work correctly. Previously, the fallback address was always used, which caused connection instability.

### DIFF
--- a/src/Cafe/OS/libs/nn_ac/nn_ac.cpp
+++ b/src/Cafe/OS/libs/nn_ac/nn_ac.cpp
@@ -4,6 +4,9 @@
 
 #if BOOST_OS_WINDOWS
 #include <iphlpapi.h>
+#elif BOOST_OS_LINUX
+#include <ifaddrs.h>
+#include <net/if.h>
 #endif
 
 // AC lib (manages internet connection)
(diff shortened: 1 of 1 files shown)


goeiecool9999:
This is good.
I have one note. What if there are multiple IP addresses associated with the host? For example if the user is running virtual machines with libvirt it creates a bridge interface with a valid IP address. And when a VPN is running it also creates a virtual interface with an IP address associated with it. Often these interfaces will have a RFC 1918 address so we can't filter by that either.
It's possible that these virtual interfaces will appear later in the linked list because they're usually created later than the interfaces for the physical hardware, but I don't think the API explicitly guarantees that order and if so we probably shouldn't rely on it(?)
At least virtual VPN connections can be filtered by checking if the interface has the IFF_POINTOPOINT flag set though.

ashquarky:
There isn't a great way to detect unusual/bridge interfaces like those made by virtual machines. One could look at the default route, but then it gets caught up by VPNs when we want specifically the outside LAN. It might be possible to eliminate bridges entirely, but some folks (including me!) deliberately have their LAN on a bridge interface so it can be easily attached to virtual things.

I've added a filter for point to point, as well as some common bridge names (br- for libvirt, docker) as a rough heuristic.

Exzap:
I think its fine. Thanks!

PR #1842, opened 2026-03-18
Fix issues found by UBSan and ASan

fabioarnold:
These are various issues I encountered while starting up Zelda TPHD in a debug build of Cemu with UBSan and ASan enabled. Each commit fixes one issue.

Most of these are over-/underflows with signed integer arithmetic which is undefined behavior in C++. I fixed this using unsigned ints where over-/underflows are defined as wrapping.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,6 +76,14 @@ set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 set(CMAKE_INTERPROCEDURAL_OPTIMIZATION_RELEASE ON)
 set(CMAKE_INTERPROCEDURAL_OPTIMIZATION_RELWITHDEBINFO ON)
 
+option(ENABLE_SANITIZERS "Enable AddressSanitizer and UndefinedBehaviorSanitizer" OFF)
+if (ENABLE_SANITIZERS)
+	add_compile_options(-fsanitize=address,undefined -fno-sanitize=vptr -fno-omit-frame-pointer)
+	add_link_options(-fsanitize=address,undefined -fno-sanitize=vptr)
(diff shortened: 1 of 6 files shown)


src/Cafe/HW/Espresso/Interpreter/PPCInterpreterImpl.cpp
```diff
@@ -11,7 +11,9 @@ class PPCItpCafeOSUsermode
 
 	inline static uint32 memory_readCodeU32(PPCInterpreter_t* hCPU, uint32 address)
 	{
-		return _swapEndianU32(*(uint32*)(memory_base + address));
+		uint32 v;
+		memcpy(&v, memory_base + address, sizeof(uint32));
```
Exzap: Yeah unfortunately we can't actually do this. In debug builds this will compile down to a function call plus some overhead and since the memory_* and ppcMem_* functions are all critical for performance (literally called many millions of times per frame) it would hurt performance too much. I further suspect that in release mode it will also not nicely compile down to a single instruction (which the old code does) since the compiler doesn't know that address is guaranteed to be aligned. So it will have to emit a byte-wise copy.

We can probably tell the compiler that the pointer is aligned via some [assume](https://en.cppreference.com/w/cpp/language/attributes/assume) construct but I wouldn't trust them to always optimize down to a dword fetch (it certainly wont optimize in debug builds).

C++23 has a solution for this: `std::start_lifetime_as`. So I think best course of action is to just wait until thats broadly available.

fabioarnold:
Apart from unaligned reads and writes in PPCInterpreterImpl.cpp with these changes I can load and run TPHD without trapping due to UBSand/ASan now.

PR #1847, opened 2026-03-24
Migrate from SDL2 to SDL3 3.4.2

lijunyu-cn:
Also fixed crashes if game controllers disconnect/reconnect more than 8 times. Tested under Windows, testing under Linux and Mac needed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -139,7 +139,7 @@ option(ENABLE_CUBEB "Enabled cubeb backend" ON)
 option(ENABLE_WXWIDGETS "Build with wxWidgets UI (Currently required)" ON)
 
 find_package(Threads REQUIRED)
-find_package(SDL2 REQUIRED)
+find_package(SDL3 REQUIRED CONFIG REQUIRED COMPONENTS SDL3)
 find_package(CURL REQUIRED)
 find_package(pugixml REQUIRED)
 find_package(RapidJSON REQUIRED)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -202,7 +202,6 @@ if (ENABLE_DISCORD_RPC)
 endif()
 
 if (ENABLE_HIDAPI)
-	find_package(hidapi REQUIRED)
 	set(SUPPORTS_WIIMOTE ON)
 	add_compile_definitions(HAS_HIDAPI)
 endif ()
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -166,7 +166,7 @@ target_link_libraries(CemuBin PRIVATE
 	CemuGui
 	CemuInput
 	CemuUtil
-	SDL2::SDL2
+	SDL3::SDL3
 )
 
 if(UNIX AND NOT APPLE)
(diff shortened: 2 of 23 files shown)


Squall-Leonhart:
SDL_main is used incorrectly, it won't build on anything other than windows

lijunyu-cn:
> SDL_main is used incorrectly, it won't build on anything other than windows

Thanks, I will try fix this.

Exzap:
Current build failure could be due to caching or some dependency still pulling in SDL2. I'll look into it

Squall-Leonhart:
the logs say theres a failure to upload nupkgs to the cemu nuget repository?

lijunyu-cn:
It can be roughly determined that the Ubuntu build failure is due to a wayland conflict between wxWidgets and SDL3. Disable wayland feature in SDL3 can make a build success, but not sure what downside it is.

Squall-Leonhart:
quickfix, set the linker flag `-Wl,-z,muldefs` this is safe here because both symbols are built from the same wayland definition.

the build failure is because both SDL3 and WxGTK export a wayland symbol `wp_pointer_warp_v1_interface`
unless its safe to disable wayland in SDL3?

Exzap:
> It can be roughly determined that the Ubuntu build failure is due to a wayland conflict between wxWidgets and SDL3. Disable wayland feature in SDL3 can make a build success, but not sure what downside it is.

This should be fine. We only use the input part of SDL3.

Squall-Leonhart:
with the latest commit, Classic controllers are working, sorta.

image

lijunyu-cn:
**Remaining issues**

1.  SDL thread safety assertion failure in Debug builds. This issue has existed since SDL2, but it was not triggered because SDL2's assertions might have had logic loopholes.

2.  Destructors for `InputManager` and the classes it manages are never triggered, thus exiting without executing resource release operations like `SDL_Quit`. Additionally, one or more of these destructors cause a deadlock.

3.  SDL2 and SDL3 require that initialization, event handling, and quit functions should only be used on the "main thread". We currently only use its input part (additionally, some video functionality is used in the screensaver disabling module, and because initialization is not on the "main thread", this code causes crashes on macOS). For the sake of functional modularity, I think it is acceptable to keep SDL event handling in a thread created by `SDLControllerProvider` as it is currently done, but SDL initialization and quit functions must also be placed in this thread. This complies with the "main thread" requirements for Windows and Linux. macOS has different "main thread" requirements, but as long as SDL video functions are not used in this "main thread", the input functionality should be usable on macOS just like in SDL2.

Exzap:
The PR is good as-is and those problems are largely a separate issue. I am currently finishing up some other commits and after that I will look into merging this PR and addressing the remaining issues you listed.

Exzap:
I added a codepath for macOS so that SDL functions will be called from the main thread there (I checked and the main thread requirements seemed to be pretty strict for SDL3 and include event pumping). Using SDL_WaitEvent has lower latency than polling so I decided to keep that path for the other platforms. I also tweaked DSUControllerProvider and removed unnecessary checks and changed the mechanism by which the reader thread is woken up. I think the deadlock you mentioned came from the writer thread not being woken up but either way it should be fixed.

Anyway, I am going to merge this once the current CI run finishes, thanks for the PR!

PR #1852, opened 2026-03-28
revert #1731 and gate the bluez headers properly.

Squall-Leonhart:
#1731 took a sledge hammer to the pairing button and skipped it entirely by build gating it behind linux only libraries, when those linux only libraries exist in an attempt to prevent builds failing on distro's lacking the bluetooth headers provided by bluez.

This reverts that and properly gates the headers behind the HAS_BLUEZ definition.

second commit improves initial pairing with RVL-CNT-01-TR controllers, starts hid if authenticated(implies remembered), drops the unpairs and re-pairs the wiimote if remembered but not authenticated.

adds a retry with an alternative wiimote side pin (can help with 1+2 pairing)
and closes handles that were previously being leaked.

Fixes the wiimote settings gui (#1854)

### DIFF
--- a/src/gui/wxgui/CMakeLists.txt
+++ b/src/gui/wxgui/CMakeLists.txt
@@ -129,6 +129,10 @@ if (ENABLE_BLUEZ)
   target_compile_definitions(CemuWxGui PRIVATE HAS_BLUEZ)
 endif()
 
+if (SUPPORTS_WIIMOTE)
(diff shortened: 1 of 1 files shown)


Squall-Leonhart:
while im here in these files i'll also test skipping authentication for already paired controllers, it seems to be the reason for instant skip to failure on windows and push that along with this.

PR #1853, opened 2026-03-28
making wii motes less crashy

Squall-Leonhart:
probe devices outside the exclusive lock to avoid blocking reader/writer threads during I/O

Replace newly opened handles with existing ones for already-active devices to avoid closing and reopening handles every poll cycle (handle shenigans can be racy)

improve device connect/disconnect safety.

remove early is_connected return that prevented cemu from setting the LED on re-associated wiimotes on hot-swap/reconnect and on start if a mote is already associated and attached.

### DIFF
--- a/src/input/api/Wiimote/WiimoteControllerProvider.cpp
+++ b/src/input/api/Wiimote/WiimoteControllerProvider.cpp
@@ -1,6 +1,7 @@
 #include "input/api/Wiimote/WiimoteControllerProvider.h"
 #include "input/api/Wiimote/NativeWiimoteController.h"
 #include "input/api/Wiimote/WiimoteMessages.h"
+#include "input/InputManager.h"
 
 #ifdef HAS_HIDAPI
 #include "input/api/Wiimote/hidapi/HidapiWiimote.h"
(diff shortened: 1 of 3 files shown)


Squall-Leonhart:
review please @capitalistspz

Squall-Leonhart:
so far i've been able to bind the same remote to every tab, remove it, and bind it again without a crash, usually it'd happen by the 4th or 5th.

src/input/api/Wiimote/WiimoteControllerProvider.cpp
```diff
@@ -37,38 +38,51 @@ std::vector<std::shared_ptr<ControllerBase>> WiimoteControllerProvider::get_cont
 	auto devices = m_connectedDevices;
 	m_connectedDeviceMutex.unlock();
 
-	std::scoped_lock lock(m_device_mutex);
-
+	// probe devices outside the exclusive lock to avoid blocking reader/writer threads during I/O
+	std::vector<WiimoteDevicePtr> reachableDevices;
 	for (auto& device : devices)
 	{
 		const auto writeable = device->write_data({kStatusRequest, 0x00});
```
capitalistspz, line 45: `HidapiWiimote::write_data` calls `hid_write`, which [isn't thread safe](https://github.com/libusb/hidapi/wiki/Multi%E2%80%90threading-Notes#multiple-devices)

PR #1857, opened 2026-03-30
Vulkan: Use LLVM shader compiler for BOTW on Linux + Mesa

goeiecool9999:
Workaround for GPU crashes when using certain runes in BOTW
Can be overridden by setting RADV_DEBUG=aco

### DIFF
--- a/src/gui/wxgui/MainWindow.cpp
+++ b/src/gui/wxgui/MainWindow.cpp
@@ -623,6 +623,35 @@ bool MainWindow::FileLoad(const fs::path launchPath, wxLaunchGameEvent::INITIATE
     }
 #endif
 
+	// Workaround for BOTW + Mesa. Runes like Magnesis and the camera cause GPU crashes.
+	// Using the LLVM shader compiler prevents crashes which points to an issue with the default ACO compiler.
+	// Either that or Cemu is violating a specification (GLSL?) causing the shaders to be broken after compilation.
+#if BOOST_OS_LINUX
+	uint64 currentTitleId = CafeSystem::GetForegroundTitleId();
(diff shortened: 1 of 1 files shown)


Squall-Leonhart:
considering even web browsers are broken on ACO, its unlikely cemu's fault.

Exzap:
Can you trigger release on your fork so that we can send that to people for testing

goeiecool9999:
> Can you trigger release on your fork so that we can send that to people for testing

Done https://github.com/goeiecool9999/Cemu/releases/tag/v2.7

emansom:
> considering even web browsers are broken on ACO, its unlikely cemu's fault.

**Wild claim. Can you substantiate?** I've been using ACO for the past seven years, on a RX 580 and later 9070 XT. Never encountered it crashing my web browser.

And **if** that is **happening** to you (web browsers crashing, being "broken"), please **report** to the **upstream** Mesa project so they can fix it. Or **replace** your graphics card, if the cause is **bad/corrupt memory**. Or **revert** your **overclock**, if that's the **cause**.

`RADV_DEBUG=llvm` is a **fallback** for **debugging** purposes, it even says so in the environment variable name: `RADV_DEBUG`. It triggers the LLVM compiler to act as shader compiler and is **not supposed to be used by programs, as it is very slow**.

**Enforcing** the **LLVM** shader compiler for BotW in Cemu, **now ensures everyone has to wait a god awful long time for shader compilation**. Not a great end user experience! And probably the opposite of what Cemu is striving for!

<sup><sub>_Bolds so it is readable for zero attention span, zero critical thought fast readers._</sup></sub>

PR #1861, opened 2026-04-05
Revert #1857

pixelcluster:
#1857 started manipulating RADV_DEBUG variables to force LLVM as a compiler backend. **This is really broken**, for multiple reasons.

First of all, the LLVM backend is completely unmaintained. Nobody is looking at issues for it, much less ever testing it. This has been the case for years.

Should any issues arise, modifying such environment variables under the user's belt is very misleading. Developers will not suspect the LLVM backend to be used at all, and unless this behavior is known by whoever is debugging the issue, it will probably cause lots of headaches.

Also, I'm not sure if this ends up applying to Cemu, but there are a ton of features that just plainly aren't there with the LLVM backend. Performance of generated shader ISA is worse, time required to compile shaders is also worse.

Worst of all, perhaps, is that RADV has no hard requirement on LLVM at all. If RADV is not compiled with LLVM (some distributions might disable this) and RADV_DEBUG=llvm is set, the driver will crash instantly.

All of this is to say that RADV_DEBUG options are strictly reserved for users debugging a specific issue and apps should never set them themselves, especially not RADV_DEBUG=llvm.

The issue this is working around has since been fixed in upstream Mesa, so the workaround is not required for a sufficiently new Mesa, either. The next stable release should pick this up, and for the time being mesa-git has the fix as well.

### DIFF
--- a/src/gui/wxgui/MainWindow.cpp
+++ b/src/gui/wxgui/MainWindow.cpp
@@ -623,35 +623,6 @@ bool MainWindow::FileLoad(const fs::path launchPath, wxLaunchGameEvent::INITIATE
     }
 #endif
 
-	// Workaround for BOTW + Mesa. Runes like Magnesis and the camera cause GPU crashes.
-	// Using the LLVM shader compiler prevents crashes which points to an issue with the default ACO compiler.
-	// Either that or Cemu is violating a specification (GLSL?) causing the shaders to be broken after compilation.
-#if BOOST_OS_LINUX
-	uint64 currentTitleId = CafeSystem::GetForegroundTitleId();
-	if (currentTitleId == 0x00050000101c9500 || currentTitleId == 0x00050000101c9400 || currentTitleId == 0x00050000101c9300)
-	{
-		// if the variable is empty set it to llvm, otherwise check if it contains llvm/aco and if not append it
-		if (const char* value; (value = getenv("RADV_DEBUG")) != NULL && strlen(value) != 0)
-		{
-			std::string valueStr{value};
-			// only append ,llvm when llvm or aco are not already passed as flags.
-			// "aco" is not a mesa flag (anymore, it used to be when llvm was default)
-			// but it will provide users with a way to override this workaround by setting RADV_DEBUG=aco
-			// should parse the flag list but there are currently no other flags containing llvm or aco as a substring
-			if (valueStr.find("llvm") == std::string::npos && valueStr.find("aco") == std::string::npos)
-			{
-				valueStr.append(",llvm");
-				setenv("RADV_DEBUG", valueStr.c_str(), 1);
-			}
-		}
-		else
-		{
-			setenv("RADV_DEBUG", "llvm", 1);
-		}
-	}
-
-#endif
-
 	CreateCanvas();
 	CafeSystem::LaunchForegroundTitle();
 	RecreateMenu();


Exzap:
There are some misleading/exaggerated statements (like saying the LLVM backend is completely unmaintained which is not [the case](https://cgit.freedesktop.org/mesa/mesa/log/?q=ac/llvm&qt=grep)), although fundamentally I agree that it's not a great workaround. But that doesn't change the practical nature of the issue: A crashing game is not a playable game. So no matter how ugly the fix, going from crashing to not crashing is an improvement. And for what it's worth no other downsides have been reported so maybe the downsides just don't affect us or it wasn't tested well enough.

I think a better solution than just straight up reverting would be to narrow the conditions on when the workaround is applied. For example checking for the mesa version range which is affected by the bug.

goeiecool9999:
When I proposed this workaround I didn't know how soon the upstream issue would be resolved after I figured out how to easily bisect RADV and I actually did that. Like Exzap said, with the mesa bug present the game was unplayable, and setting environment variables in general is confusing for users. I even provided an override by reintroducing RADV_DEBUG=aco for those that know what they're doing. But yes, as soon as the fix hits the distros that stay reasonably up to date this workaround should be removed (or narrowed like Exzap suggested and removed farther in the future)

pixelcluster:
> saying the LLVM backend is completely unmaintained which is not [the case](https://cgit.freedesktop.org/mesa/mesa/log/?q=ac/llvm&qt=grep)

I'm not trying to mislead anyone here, and I'm not sure why you'd accuse me of acting in bad faith. To explain the commit log you posted: None of these commits are for RADV's LLVM compilation path. RADV uses the NIR->LLVM translation utilities that are, at this point, primarily used by radeonsi. RADV's LLVM backend is effectively unmaintained, ac/llvm receiving commits because of other users does not change that.

> And for what it's worth no other downsides have been reported so maybe the downsides just don't affect us or it wasn't tested well enough.

I would suspect the latter is the case. Like I said, there are distros on which this commit will cause instant crashes. You have no way to detect whether it will crash or not. This happens independently of Mesa versions, and it depends on a build-time flag where you don't know if it was set or not. Arguably, this is even a regression for these distros.

> I even provided an override by reintroducing RADV_DEBUG=aco for those that know what they're doing.

That is meaningless in the context of the problems I raised. To even get the idea of trying out RADV_DEBUG=aco, you need to be a) aware of the fact Cemu sets RADV_DEBUG=llvm in the first place and b) have read either the code or the discussion here or in #1857. I think almost everyone who would meet both conditions has commented in this PR by now as well.

>  So no matter how ugly the fix, going from crashing to not crashing is an improvement.

Sure, but what I'm questioning is why Cemu is even trying to fix it in the first place. As per https://docs.mesa3d.org/release-calendar.html, the next stable Mesa release (which will contain this fix) is April 15. Why jump the gun here? Is there even going to be a Cemu release before that point? Why not wait a few more days until the proper fix arrives for everyone? The Mesa report has been open for a month now, 10 more days aren't going to end the world. However, like I said above, messing with debug options that you're not supposed to be messing with causes issues.

Exzap:
> Sure, but what I'm questioning is why Cemu is even trying to fix it in the first place. As per https://docs.mesa3d.org/release-calendar.html, the next stable Mesa release (which will contain this fix) is April 15. Why jump the gun here? Is there even going to be a Cemu release before that point? Why not wait a few more days until the proper fix arrives for everyone?

We are doing this because not everyone updates Mesa immediately. Many users are on distributions with slow update cycles or are stuck on an affected version of Mesa for other reasons. When the emulator crashes, users also don't know it's Mesa's fault. Instead they blame us and that creates a lot of troubleshooting burden that we really don't want and we are already low on capacity.

I am closing this PR because reverting without an alternative solution is just not acceptable for all the reasons I mentioned. You have more insight in Mesa so maybe you can think of a workaround with less or no downsides. Otherwise, we will move forward with refining the current workaround to specifically target the affected Mesa versions, assuming thats possible.

pixelcluster:
> We are doing this because not everyone updates Mesa immediately.

Since this is a Mesa regression, I'm not sure these users would even be affected by the Mesa bugs this PR is working around in the first place. People that update frequently enough to be affected will also receive the fix resolving the issues for them very soon. (Also, running bleeding-edge Cemu but ancient Mesa? interesting, but eh, I guess someone would try.) What people *will* be affected by, though, is this PR crashing the emulator on startup when run on systems that compile RADV without LLVM. Like I said before, this is not fixable by improving the workaround.

> Instead they blame us and that creates a lot of troubleshooting burden that we really don't want and we are already low on capacity.

I truly can empathize with that, but what I'm trying to tell you is that your workaround is not viable. The only proper solution is waiting for the Mesa bugfix to trickle down.

Squall-Leonhart:
> The issue this is working around has since been fixed in upstream Mesa, so the workaround is not required for a sufficiently new Mesa, either. The next stable release should pick this up, and for the time being mesa-git has the fix as well.

no, it hasn't.

Or steamos would have finally released a 6.18+ based beta.

pixelcluster:
I'm not sure why you think the kernel version is relevant here at all. Good that you mention SteamOS though, because it's one of the distros that ship RADV without LLVM. This workaround should therefore cause instant crashes on all SteamOS-based devices.

R1chterScale:
> > The issue this is working around has since been fixed in upstream Mesa, so the workaround is not required for a sufficiently new Mesa, either. The next stable release should pick this up, and for the time being mesa-git has the fix as well.
> 
> no, it hasn't.
> 
> Or steamos would have finally released a 6.18+ based beta.

Yes it has:

https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/40770

Just a thought, maybe the actual RADV dev might know a tad bit better about that state of RADV :)

Squall-Leonhart:
the radv dev is ignoring the other issues with amdgpu and the linux kernel and is speaking out of turn.

especially given the false information about steam os's llvm capability.

daniel-schuermann:
> the radv dev is ignoring the other issues with amdgpu and the linux kernel and is speaking out of turn.

Why would user-space driver dev need to take care of kernel driver issues? We are also not asking you to fix mesa issues. In fact, we ask you to do the opposite and leave our debugging ENV vars untouched, so that issues don't get hidden away.

pixelcluster:
I mean, I actually do also take care of kernel issues sometimes. I just have no idea which kernel issues are being talked about here, or why they would be relevant to RADV debugging variables in the slightest.

Also, which information I provided was false? I'm happy to clarify if there was a misunderstanding.

Squall-Leonhart:
> Why would user-space driver dev need to take care of kernel driver issues? We are also not asking you to fix mesa issues. In fact, we ask you to do the opposite and leave our debugging ENV vars untouched, so that issues don't get hidden away.

because the UMD can antagonise or mitigate issues in the KMD by being more or less aggressive in its assembly generation.

Changing from aco to llvm has not been a consistent workaround throughout the life cycle of mesa 26, and changes to unreleased mesa 26 do not necessarily fix the problem, alleviation to some degree might be achieved but until the kernel gets its crap together, claiming "Fixed in mesa upstream" isn't reliably affirmable.

Squall-Leonhart:
> Also, which information I provided was false? I'm happy to clarify if there was a misunderstanding.

steamos still includes llvm for fallback

> I mean, I actually do also take care of kernel issues sometimes. I just have no idea which kernel issues are being talked about here, or why they would be relevant to RADV debugging variables in the slightest.

https://gitlab.freedesktop.org/drm/amd/-/work_items/2950#note_3279864

pixelcluster:
> steamos still includes llvm for fallback

Not sure which version you were looking at. If anything, try SteamOS Main. It might have still been enabled in some previous version, but I can absolutely 100% assure you it's on its way out.

> https://gitlab.freedesktop.org/drm/amd/-/work_items/2950#note_3279864

Completely, entirely unrelated. I can't even see anyone mentioning RADV_DEBUG ever? In any case, that issue has zero relevance to Cemu running into GPU hangs.

Squall-Leonhart:
> Completely, entirely unrelated. I can't even see anyone mentioning RADV_DEBUG ever? In any case, that issue has zero relevance to Cemu running into GPU hangs.

lol. ok.

except these radeon issues predate mesa 26.

pixelcluster:
> except these radeon issues predate mesa 26.

Correct, there have been bugs in the stack before Mesa 26 as well. Just like any other driver stack. The bug you linked is a kernel driver issue, so Mesa version most likely won't matter and it will happen with literally any Mesa version you might install. What's your point?

CounterPillow:
Chiming in to say that distro release cadence is a non-argument for keeping a workaround if it's been fixed upstream. Fixes will be picked into bugfix release branches of Mesa, and "stable" distributions that decide to stay on upstream-unsupported Mesa versions like the debioids aren't off the hook either. On those, users are supposed to file a bug against the package, and link to the patch they want backported with an explanation as to why.

The kernel side of the equation, amdgpu, being a buggy disaster is also unrelated to any discussion about Mesa. The kernel wouldn't even be able to know what shader compiler Mesa uses, because looking at the shader code is not something a kernel driver for a GPU does.

emansom:
This PR **should** be **re-opened**, and **merged** once it is **confirmed** the bug is **no longer present** in ACO/RADV and/or Cemu itself.

Userspace applications **should not workaround bugs** in the graphics driver stack, rather this **should be fixed upstream**.

@pixelcluster is **acting in good faith** here and wants the Linux experience for Cemu to be **performant and great all across the board**, enforcing LLVM as shader compiler acomplishes the opposite.

<sup><sub>_Bolds so it is readable for zero attention span, zero critical thought fast readers. Thanks for fucking up Cemu btw @goeiecool9999 and @Exzap. I am a less good faith person here, as I am a pissed off user that does not like random contributors introducing subtle performance reducing unrecommended patchy workarounds in software! AND IT GETTING MERGED?! HELLO?! IS THIS REAL SOFTWARE DEVELOPMENT OR JUST A HOBBY YOLO PROJECT, WHAT THE HELL IS GOING ON?!_</sup></sub>

emansom:
This PR **should** be **re-opened**, and **merged** once it is **confirmed** the bug is **no longer present** in ACO/RADV and/or Cemu itself.

Userspace applications **should not workaround bugs** in the graphics driver stack, rather this **should be fixed upstream**.

As I'm fairly certain Cemu intent is not to force users like me, to maintain seperate patches to get good performance in BoTW?

@pixelcluster is **acting in good faith** here and wants the Linux experience for Cemu to be **performant and great all across the board**, enforcing LLVM as shader compiler acomplishes the opposite.

<sup><sub>_Bolds so it is readable for zero attention span, zero critical thought fast readers._</sup></sub>

PR #1862, opened 2026-04-05
vulkan: Improve the Breath of the Wild RADV/LLVM workaround

goeiecool9999:
Use child process to detect the presence of the LLVM compiler
Only apply workaround on affected mesa versions.
Log the manipulation of RADV_DEBUG

### DIFF
--- a/src/Cafe/HW/Latte/Renderer/Vulkan/VulkanRenderer.cpp
+++ b/src/Cafe/HW/Latte/Renderer/Vulkan/VulkanRenderer.cpp
@@ -339,12 +339,12 @@ void VulkanRenderer::GetDeviceFeatures()
 #if BOOST_OS_LINUX
 #include <sys/wait.h>
 
-void WorkaroundChildAbortHandler(int unused)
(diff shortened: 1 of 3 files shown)


K0bin:
Honestly, why bother supporting old driver versions like that?

Proton doesn't work around old driver bugs so people who play games on Linux are strongly incentivized to keep their GPU driver up to date anyway.

Exzap:
@K0bin This is an emulator for a 14 year old console. The whole spirit of the project is about preservation and long-term support. But to give you a more tangible answer, the crash problem affects people *right now*. It's not an old Mesa version, it's the current release. The next Mesa release fixes it, but not everyone can or wants to update immediately. Ultimately, it's not my place to dictate which Mesa version a user needs to have installed and our software should just work without caveats. At least that is what we are striving towards.

I also want to mention that Cemu is officially distributed through auto-updating appimage and flatpak builds, so fixes usually reach users quite quickly. For critical issues we even provide pre-releases when the next release is too far off. In fact we did this for the issue addresses by this PR. So affected users can play **right now** instead of having to wait because of release schedules.

PR #1871, opened 2026-04-12
LatteBufferCache: fix interval lookups after range splits

oltolm:
In Devil's Third I had image corruption. I fixed it with the help of Copilot. I don't know why other people haven't reported it. Maybe because I am using latest version from `main` and I have a Radeon RX 580. The corruption was a heavy flicker that occurred with both OpenGL and Vulkan.

Screenshot 2026-04-12 125316

IntervalTree2 used std::map::find() for overlap-based queries, which can miss a valid covering node after range splits or resizes. Switch these lookups to a lower_bound-based first-overlap search so reserve and retrieve paths always find the correct cache range.

Also keep the original heap allocation offset separate from the visible cache offset and update page metadata correctly when shrinking a node. That keeps frees and deferred releases aligned with the underlying allocation and avoids stale page state after front trims.

### DIFF
--- a/src/Cafe/HW/Latte/Core/LatteBufferCache.cpp
+++ b/src/Cafe/HW/Latte/Core/LatteBufferCache.cpp
@@ -46,10 +46,18 @@ class IntervalTree2
 	std::map<InternalRange, TNodeObject*> m_map;
 	std::vector<TNodeObject*> m_tempObjectArray;
 
+	typename std::map<InternalRange, TNodeObject*>::iterator findFirstOverlapping(TRangeData rangeBegin, TRangeData rangeEnd)
+	{
+		auto itr = m_map.lower_bound(InternalRange(rangeBegin, rangeBegin + 1));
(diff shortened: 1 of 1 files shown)


Exzap:
The issues are real but it's quite surprising that this never manifested as problems on a larger scale. IntervalTree is extremely central to the renderer and gets hammered a lot. Vertex corruptions do get reported very rarely already and can almost always be attributed to something else. Specifically I have never seen these glitches in Devil's Third. Not saying that there isn't a serious problem in the code, but it's interesting that in practice we must have skirted around it by relying on std::map implementations working a certain way.

Since I had plans to rewrite the code anyway (there is headroom for optimization and the current code is a bit messy) I will see if I can do that soonish. Until then I will leave this PR open as a backup plan. Thanks for bringing this up.

Exzap:
Can you check if the issue is gone for you on latest main? c080b7385e20cdd9635f69ea29f00c4260bd57e4 + 58c25d5303d668a7882e7754cfd30087375e3e6f should have fixed it

oltolm:
It seems to be fixed.

PR #1872, opened 2026-04-12
Update macOS build instructions for ARM64 compilation

emiyl:
The build instructions are outdated and instruct the user to compile an x86 version of Cemu, as there was no ARM64 version at the time of writing. This PR removes the instructions to compile the x86 version so that the user will compile a version of Cemu specific to the architecture of their device.


Exzap:
thanks!

PR #1875, opened 2026-04-15
Build: Codesign macOS app bundle

shinra-electric:
This PR adds an entitlements file and uses it to codesign the app bundle in the Github Actions runs. 

I conferred with @SamoZ256 as to which entitlements would be appropriate for Cemu, and I believe these should be ok. 

It would be preferable to get CMake to perform the codesigning step, but further changes would have to be made. This could be done in a future PR.


Exzap:
thanks!

PR #1878, opened 2026-04-16
CI: Silence runner warnings as ninja is already installed

shinra-electric:
When building the macOS artifacts, there is no need to install ninja.

Silences these warnings:

Screenshot 2026-04-16 at 23 24 38

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -230,7 +230,7 @@ jobs:
     - name: "Install system dependencies"
       run: |
         brew update
-        brew install ninja nasm automake libtool
+        brew install nasm automake libtool
 
     - name: "Install molten-vk"
       run: |


emiyl:
Why not do something like this?

```
if ! command -v ninja; then
      brew install ninja
fi
```

I imagine it's cached or something, but it's good to still have the install instruction there.

shinra-electric:
> Why not do something like this?
> 
> ```
> if ! command -v ninja; then
>       brew install ninja
> fi
> ```
> 
> I imagine it's cached or something, but it's good to still have the install instruction there.

Well, we could do a check for it, but it is not cached - it comes pre-installed in the runner image so it is always expected to be there. If it is ever removed from the runners for whatever reason, the CI would fail and we could just add it back. 

See [this](https://github.com/actions/runner-images/issues/11391) for more details.

emiyl:
Ahh fair enough I hadn’t realised they’d added it to the runner

PR #1879, opened 2026-04-18
Log CPU and OS version on macOS

emiyl:
This PR brings the macOS logs in line with the Windows/Linux logs.

Before:
```
[16:32:27.393] RAM: 24576MB
[16:32:27.393] Platform: MacOS
```

After:
```
[16:36:26.981] CPU: Apple M5 Pro
[16:36:26.981] RAM: 24576MB
[16:36:26.981] Platform: macOS 26.4 (25E246)
```


emiyl:
hang on i should probably just chuck it into cpufeatures..

emiyl:
there we go that's better

PR #1883, opened 2026-04-19
macOS CMake fixes

emiyl:
While it _technically_ works, shell scripts like update.sh should not be in Contents/MacOS in an app bundle. The correct place to put them is in Contents/Resources. Contents/MacOS should be reserved only for executable binaries, and not for interpreted scripts, and it makes codesigning/notorization more fragile. The script has been moved now, and CemuUpdateWindow.cpp has been updated to use the new script location.

I've also swapped out the bash command (`COMMAND bash -c "install_name_tool ...`) for the proper CMake equivalent, and moved `chmod a+x` from the GitHub action over the CMake script, ~~so the updater works on home-compiled versions.~~ _Actually since it's being invoked from Cemu explicitly it doesn't need to be marked as executable. Still probably good to have this here._

As an added bonus this makes the project compilable with `-G Xcode`, so I've added two XCode lines in there to make building easier. This shouldn't affect Ninja compilation.

Not quite in scope of this PR put I also made ZArchive do `find_package(ZArchive QUIET)` because it was annoying me. Chucked it in here since idk if it's a big enough change to necessarilly need its own PR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,6 +78,7 @@ set(CMAKE_INTERPROCEDURAL_OPTIMIZATION_RELWITHDEBINFO ON)
 
 if (MSVC)
 	set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY VS_STARTUP_PROJECT CemuBin)
+	set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY XCODE_STARTUP_PROJECT CemuBin)
 	# floating point model: precise, fiber safe optimizations
 	add_compile_options(/EHsc /fp:precise)
 	if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,7 +78,6 @@ set(CMAKE_INTERPROCEDURAL_OPTIMIZATION_RELWITHDEBINFO ON)
 
 if (MSVC)
 	set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY VS_STARTUP_PROJECT CemuBin)
-	set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY XCODE_STARTUP_PROJECT CemuBin)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -78,6 +78,7 @@ set(CMAKE_INTERPROCEDURAL_OPTIMIZATION_RELWITHDEBINFO ON)
 
 if (MSVC)
 	set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY VS_STARTUP_PROJECT CemuBin)
+	set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY XCODE_STARTUP_PROJECT CemuBin)
```
goeiecool9999: This is inside `if (MSVC)` which checks if the microsoft compiler is used, yet it sets a property related to Xcode. Is that intentional?

shinra-electric:
I'm not sure if it's preferable to move the update script step from the Action build to CMake. 

Personally, when I build locally it's usually to test a PR or some specific code I've written, I don't want the updater to replace it with a new build.

I think the updater should only work on official actions builds

CMakeLists.txt
emiyl: whoops, nope

emiyl:
@shinra-electric this only applies if you're building an app bundle. In the current behaviour, you still get the update.sh, it's just not executable. If you're just testing, and you're not enabling the mac bundle, the update.sh won't be there. Also fwiw I've never had an update dialogue appear on a test build.

shinra-electric:
> @shinra-electric this only applies if you're building an app bundle. In the current behaviour, you still get the update.sh, it's just not executable. If you're just testing, and you're not enabling the mac bundle, the update.sh won't be there. 

I usually build an app bundle 😅

>Also fwiw I've never had an update dialogue appear on a test build.

Actually, that's a good point. Me neither. But if at some point in the future it does work, I still wouldn't want it replacing my local build

emiyl:
@shinra-electric well.. 
image

You can always press no.

Also I just tested it and the script still works even if it's not marked as executable so.. this actually doesn't change the behaviour _at all_. But still it's probably good to have it there.

PR #1886, opened 2026-04-20
Add ENABLE_LIBUSB option

emiyl:
Adds an option to build with libusb or not. Useful for building for platforms that don't support libusb such as iOS.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,9 +131,9 @@ option(ENABLE_SDL "Enables the SDLController backend" ON)
 option(ENABLE_LIBUSB "Build with libusb support" ON)
 
 if (ENABLE_LIBUSB)
-	add_compile_definitions(HAS_LIBUSB=1)
+	add_compile_definitions(HAS_LIBUSB)
 else()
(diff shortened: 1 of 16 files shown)


CMakeLists.txt
```diff
@@ -128,6 +128,13 @@ endif()
 
 option(ENABLE_HIDAPI "Build with HIDAPI" ON)
 option(ENABLE_SDL "Enables the SDLController backend" ON)
+option(ENABLE_LIBUSB "Build with libusb support" ON)
+
+if (ENABLE_LIBUSB)
+	add_compile_definitions(HAS_LIBUSB=1)
```
Exzap: To match the newer convention can you change this to:
```
if (ENABLE_LIBUSB)
	add_compile_definitions(HAS_LIBUSB)
endif()
```
With the check being `#if defined(HAS_LIBUSB)`

emiyl: 🫡

emiyl: all done

emiyl:
okay should be all ready now

PR #1887, opened 2026-04-20
Make ENABLE_OPENGL and ENABLE_VULKAN toggle associated code when compiling

emiyl:
This will enable building without OpenGL or Vulkan, which helps to build on devices without either graphic API, such as iOS.

## Changes

- Only compile OpenGL and Vulkan specific files when they are enabled
- Changed a load of if/else statements for graphic APIs to some lovely switch statements instead
- #ifdef'd all the switch statements and any OpenGL and Vulkan specific functions
- Created a "RendererCore.cpp" file for functions that all the renderers use.
- MainWindow is now more resilient if the config shows an unknown graphic API value
- Fixed an issue in LatteDecompilerAnalyzer where it would check if the API isn't Metal only when Metal was enabled
- GeneralSettings2 now has an m_api_map, so that graphic API values stay consistent with wxChoice box


src/Cafe/HW/Latte/Core/LatteShaderGL.cpp
```diff
@@ -26,7 +26,6 @@ bool gxShader_checkIfSuccessfullyLinked(GLuint glProgram)
 
 void LatteShader_prepareSeparableUniforms(LatteDecompilerShader* shader)
 {
-#ifdef ENABLE_OPENGL
```
emiyl: why has this been removed? this function has a lot of opengl stuff in it 🤔

goeiecool9999: Because the whole file is not compiled now when OpenGL is disabled. I added an ifdef at the call site.

goeiecool9999:
Just a heads up, I have moved all of the index cache stuff from RendererCore back to OpenGLRendererCore. It's only used by OpenGL. Vulkan and Metal have their own implementation

emiyl:
Oh sweet thank you - I wasn't too sure what was needed and what wasn't for those two functions so I kinda just shoved everything over until it built lol - this looks a lot cleaner though

src/Cafe/HW/Latte/Core/LatteShaderGL.cpp
emiyl: ..oh yeah

PR #1893, opened 2026-04-27
Add multiple account detection and implement GetMiiImageEx

### LINKED ISSUE #793: Minecraft Wii U cannot access system miis
It gets the name right for player one but when accessing splitscreen, you cannot see or choose any of the other miis on your system (meaning inventories do not get saved)
favmii
guestmii
Here you can see the guest miis working corectly and the favorites tab being empty

SirHrVedel:
This reworks the implementation of the account detection system in the iosu that exposes accounts to the emulated os. It allows titles and the Wii U system software to detect other users present in the MLC.
- Completely reworked the commented iosu-act multiple account detection implementation so it sorts them based on their persistantId.
- Get and use the correct slot number for the active account

- Implement GetMiiImageEx so Mii images are no longer blank or use a placeholder image in the Wii U's system software, if ofc the files are present in the accounts files.

### Aditionally
- Implement IsPasswordCacheEnabled, to get rid of the annoying password prompt for users on the Wii U Menu user select screen

Got some coding help from Claude making this.

Closes: #793 

### Images:
- Wii U Menu user select screen
Screenshot 2026-04-27 012022

- Mii Maker
Screenshot 2026-04-27 012220

- Minecraft WIi U edition
Screenshot 2026-04-27 012143

- Nintendo Land
Screenshot 2026-04-27 012431

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -250,6 +250,7 @@ Example usage: `cmake -S . -B build -DCMAKE_BUILD_TYPE=release -DENABLE_SDL=ON -
 | ENABLE_VCPKG       |   | Use VCPKG package manager to obtain dependencies                            | ON      |                    |
 | ENABLE_VULKAN      |   | Enable the Vulkan graphics backend                                          | ON      |                    |
 | ENABLE_WXWIDGETS   |   | Enable wxWidgets UI                                                         | ON      | Currently required |
+| ENABLE_LIBUSB      |   | Enable libusb                                                               | ON      |                    |
 
 ### Windows
 | Flag               | Description                       | Default | Note               |
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -132,6 +132,11 @@ endif()
 
 option(ENABLE_HIDAPI "Build with HIDAPI" ON)
 option(ENABLE_SDL "Enables the SDLController backend" ON)
+option(ENABLE_LIBUSB "Build with libusb support" ON)
+
+if (ENABLE_LIBUSB)
+	add_compile_definitions(HAS_LIBUSB)
+endif()
 
 # audio backends
 if (WIN32)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -119,10 +119,12 @@ if (MACOS_BUNDLE)
 			COMMAND ${CMAKE_COMMAND} ARGS -E copy_directory "${CMAKE_SOURCE_DIR}/bin/${folder}" "$<TARGET_BUNDLE_DIR:CemuBin>/Contents/SharedSupport/${folder}")
 	endforeach(folder)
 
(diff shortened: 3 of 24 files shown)


Exzap:
Can you factor out the `common.dat` logic (both for writing and reading) into a separate PR or just remove it for now. It needlessly complicates this PR while not actually contributing to the actual point of the PR. You can assume persistentId for account ordering, there is no need to make it more complicated than that for now.

SirHrVedel:
I'm totally fine with removing the common.dat code. The point in implementing it was to sort accounts similar to how a real Wii U does it, so it was mainly there for accuracy reasons. I do now realize my implementation of it is a bit rough.

SirHrVedel:
The build check might need to be rerun. GitHub was experiencing workflow issues when they initially ran causing the cmake error. They should be mitigated now (according to the GitHub status site). Other than that the PR should be good for further review.

SirHrVedel:
Im going to need to move the GetMiiImage code to iosu so this PR is not going to be ready for a little while yet. I also just stubbed IsPasswordCacheEnabled, as my implementation wasn't accurate. Stubbing it does essentially the same.

Exzap:
The PR looks good to me. Can you outline which parts were done by Claude exactly? There is a concern that AI might hallucinate parameters for e.g. GetMiiImageEx in a way that is good enough for most games, but not actually fully correct. HLE functions should be verified by a human.

SirHrVedel:
GetMiiImage is very much made by Claude. I only made a very few adjustments to it like making the code more readable. The signature looks correct to me, I verified it using ghidra and it looked good to me. 
GetCurrentAccountSlot i  iosu is also AI, I initially was super confused about why the first account always got loaded and it helped me fix that by reworking the old method for me (i was lazy).
The sorting method in account.cpp is ai too, after I prompted it to verify if the sorting was working correctly, because I was unsure that sorting would work on all platforms. It caught that on Linux filesystems (ext filesystems) that, account ordering might not work correctly, so it implemented the sort to fix it.

I have made sure to read the code thoroughly and verified it to the best of my abilities. For better accuracy i had Claude disassemble the rpl and fw.img it self and heavily prompted it if I found something out of the ordinary, or with lacking evidence to its claims, just so I was more confident that it was implementing the functions correctly.

Exzap:
It's probably fine, but subtle mistakes in HLE implementation can be incredibly hard to track down later. So I will at some point double check the implementation by hand just to be sure.

Unrelated to whether this was made with AI or not, with anything that touches account behavior there is always a (small) chance that it will cause save breakage. So I generally rather merge these kind of changes at the beginning of a release cycle since then there is enough time that someone would notice if there are unexpected side effects. So the plan is to keep this open until 2.7 releases and then after that I will merge it. From a technical standpoint the PR looks ok in the current state and you don't need to further tweak it.

PR #1894, opened 2026-04-28
Fixes to H264 processing

rcaridade145:
This fixes the NAL processing of h264 parser / decoder to consider 3 or 4 bytes.
I also commented a cemu_assert and change other into a log.
I used LLM in indentifying the underlying issue on NAL. 
This allows Cod:Ghosts to go into the main menu. 

Captura de ecrã 2026-04-28, às 16 04 40


SirHrVedel:
Wow, this also resolves crashes on game mode select in Wii Party U. Great work!

Crementif:
Just to check because this is a large change... but I'm having a hard time understanding what your changes do. Can you tell me what was exactly the thing that Cemu was doing but wasn't accurate.

I also see that you disabled some intentional error catching to catch an invalid input. Is that just causing the game to continue playing instead of crashing intentionally due to this error catching. Does the game show the video properly? Can you run the game without disabling the assert?

rcaridade145:
Will update the PR with more comments and a bit of a clean up.
I came into 3 issues developing. One was running in debug and hitting an assert when setting a variable. Then issues in GetImage… and used an LLM which pointed out that the NAL unit could start with 3 or 4 bytes - https://membrane.stream/learn/h264/3 this did allowed me to understand. 
With regards to the u4_num_bytes_consumed from I read the decoder may not need info from the whole buffer to decode and may stop before.

PR #1895, opened 2026-05-03
input+build: Make SDL optional

SSimco:


Exzap:
I don't think this is really a beneficial change for the desktop version of Cemu. SDL is our standard controller API across pretty much all platforms. Especially on Linux and MacOS where it's the only controller API (besides remote controller support via DSU). If this helps you with the Android port we can merge it but otherwise I prefer not having to maintain more options.

But also a heads up but this will probably have conflicts caused by #1847

SSimco:
> I don't think this is really a beneficial change for the desktop version of Cemu. SDL is our standard controller API across pretty much all platforms. Especially on Linux and MacOS where it's the only controller API (besides remote controller support via DSU). If this helps you with the Android port we can merge it but otherwise I prefer not having to maintain more options.
> 
> But also a heads up but this will probably have conflicts caused by #1847

It will help, since the port uses the Android controller APIs directly rather than SDL, and I encountered some errors the last time I tried to integrate SDL into the Android port.

I’ll resolve any merge conflicts that appear when #1847 is merged.

PR #1902, opened 2026-05-07
Stub MCP_DemoGetRemainder to 99

SirHrVedel:
This allows game demos to be launched when loading the drmapp.rpl file on the Wii U Menu.

### DIFF
--- a/src/Cafe/OS/libs/coreinit/coreinit_MCP.cpp
+++ b/src/Cafe/OS/libs/coreinit/coreinit_MCP.cpp
@@ -421,6 +421,15 @@ namespace coreinit
 		return 0;
 	}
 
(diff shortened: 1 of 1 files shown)


src/Cafe/OS/libs/coreinit/coreinit_MCP.cpp
```diff
@@ -421,6 +421,15 @@ namespace coreinit
 		return 0;
 	}
 
+	uint32 MCP_DemoLaunchGetRemainder(uint32 mcpHandle, uint32 ukn_r4, uint32 titleId_high, uint32 titleId_low, uint32be* outRemainder)
```
Exzap:
```suggestion
	uint32 MCP_DemoLaunchGetRemainder(uint32 mcpHandle, uint64 titleId, uint32be* outRemainder)
```
TitleId is passed as a uint64. This also explains why r4 is not used. Because the ABI treats registers like memory, meaning a uint64 can cause registers to be skipped for alignment padding (e.g. uint64 can only be stored in r3:r4, r5:r6, r7:r8, etc.)

PR #1910, opened 2026-05-11
WinArm64-MSVC

qurious-pixel:
Build Windows on Arm using MSVC


src/Cafe/HW/Espresso/Recompiler/IML/IMLInstruction.h
```diff
@@ -758,7 +758,7 @@ struct IMLInstruction
 	void make_fpr_compare(IMLReg regA, IMLReg regB, IMLReg regR, IMLCondition cond)
 	{
 		this->type = PPCREC_IML_TYPE_FPR_COMPARE;
-		this->operation = -999;
+		this->operation = 255;
```
Crementif: At first, I thought you made this change so that the value fits some uint8 type. But looking further, this type is larger then uint8... so why make this change? Is this to prevent a compiler warning? These random changes require a lot of decipher and double-checking, so I would just avoid them in the first place. Splitting up your PR into multiple changes would've also allowed to document these things.

src/Cafe/HW/Espresso/PPCTimer.cpp
```diff
@@ -33,39 +30,32 @@ uint64 muldiv64(uint64 a, uint64 b, uint64 d)
...
+#if defined(ARCH_X86_64)
 	if (!g_CPUFeatures.x86.invariant_tsc)
 		cemuLog_log(LogType::Force, "Invariant TSC not supported");
-    #endif
+#endif
 
-	_mm_mfence();
-	uint64 tscStart = __rdtsc();
+	BARRIER_FENCE();
+	uint64 tscStart = READ_TSC();
```
Crementif: __rdtsc was already being used our arm code. Why not reuse that instead of redefining this. Same with Multiply64to128 etc.

src/Cafe/HW/Espresso/PPCTimer.cpp
```diff
@@ -33,39 +30,32 @@ uint64 muldiv64(uint64 a, uint64 b, uint64 d)
...
 	unsigned int startTime = GetTickCount();
 	HRTick startTick = HighResolutionTimer::now().getTick();
-	// wait roughly 3 seconds
```
Crementif, line 45: Removed comments?

src/Cafe/HW/Espresso/PPCTimer.cpp
```diff
@@ -33,39 +30,32 @@ uint64 muldiv64(uint64 a, uint64 b, uint64 d)
...
 	uint64 tsc_freq = muldiv64(tsc_diff, hrtFreq, hrtDiff);
 
-	// uint64 freqMultiplier = tsc_freq / hrtFreq;
-	//cemuLog_log(LogType::Force, "RDTSC measurement test:");
-	//cemuLog_log(LogType::Force, "TSC-diff:   0x{:016x}", tsc_diff);
-	//cemuLog_log(LogType::Force, "TSC-freq:   0x{:016x}", tsc_freq);
-	//cemuLog_log(LogType::Force, "HPC-diff:   0x{:016x}", qpc_diff);
-	//cemuLog_log(LogType::Force, "HPC-freq:   0x{:016x}", (uint64)qpc_freq.QuadPart);
-	//cemuLog_log(LogType::Force, "Multiplier: 0x{:016x}", freqMultiplier);
-
```
Crementif, line 68: Ditto, comments removed.

EDIT: On hindsight. This is an issue everywhere. But I've already made a few comments on it so I'll just leave it.

src/Cafe/CMakeLists.txt
```diff
@@ -660,6 +664,10 @@ target_link_libraries(CemuCafe PRIVATE
   OpenSSL::SSL
 )
 
+if(MSVC AND CMAKE_SYSTEM_PROCESSOR STREQUAL "ARM64")
+    target_link_libraries(CemuCafe PRIVATE ntdll.lib)
```
Crementif: From my experience, linking .lib files directly is usually not necessary and often breaks things in the future. Is there another way to solve this?

src/Cafe/CMakeLists.txt
```diff
@@ -668,7 +676,9 @@ endif()
 if (ENABLE_LIBUSB)
   if (ENABLE_VCPKG)
     if(WIN32)
-      set(PKG_CONFIG_EXECUTABLE "${VCPKG_INSTALLED_DIR}/x64-windows/tools/pkgconf/pkgconf.exe")
+		if(DEFINED VCPKG_TARGET_TRIPLET)
+		    set(PKG_CONFIG_EXECUTABLE "${VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/tools/pkgconf/pkgconf.exe")
+		endif()
     endif()
```
Crementif: VCPKG_TARGET_TRIPLET can be assumed to always be available.

src/Common/precompiled.h
```diff
@@ -341,44 +341,62 @@ inline uint64 _udiv128(uint64 highDividend, uint64 lowDividend, uint64 divisor,
...
 
 inline unsigned char _addcarry_u64(unsigned char carry, unsigned long long a, unsigned long long b, unsigned long long *result)
 {
-    *result = a + b + (unsigned long long)carry;
-    if (*result < a)
-        return 1;
-    return 0;
+    unsigned __int128 res = (unsigned __int128)a + b + carry;
+    *result = (unsigned long long)res;
+    return (res >> 64) ? 1 : 0;
```
Crementif, line 413: Did you change this to use the __unsigned __int128 type, since it might be more explicit then long long. But then I saw that a line under it you still use long long. So this change seems superficial.

src/resource/installer.nsi
```diff
@@ -150,8 +158,6 @@ SectionEnd
 Section Uninstall
   Delete "$DESKTOP\$(^Name).lnk"
   Delete "$SMPROGRAMS\$(^Name).lnk"
-
-; Be a bit careful to not delete files a user may have put into the install directory
```
Crementif, line 154: This seems like a useful comment to keep for future developers, no?

src/resource/installer.nsi
```diff
@@ -162,9 +168,6 @@ Section Uninstall
   DeleteRegKey HKCU "Software\Classes\.wux"
   DeleteRegKey HKCU "Software\Classes\.wua"
   DeleteRegKey HKCU "Software\Classes\$(^Name)"
-
-  DeleteRegKey HKCU "Software\Classes\discord-460807638964371468"
```
Crementif, line 166: I'm not sure why we deleted the registry key for discord? Is this RPC related? IIRC, dolphin's installer also has a similar thing, which was what this was based off.

src/CMakeLists.txt
```diff
@@ -70,10 +70,15 @@ add_executable(CemuBin
...
-if(MSVC AND MSVC_VERSION EQUAL 1940)
-	# workaround for an msvc issue on VS 17.10 where generated ILK files are too large
-	# see https://developercommunity.visualstudio.com/t/After-updating-to-VS-1710-the-size-of-/10665511
-	set_target_properties(CemuBin PROPERTIES LINK_FLAGS "/INCREMENTAL:NO")
+if(MSVC)
+    set(CEMU_LINKER_FLAGS "/NODEFAULTLIB:MSVCRT")
+
+    # Check for the specific VS 17.10 version
+    if(MSVC_VERSION EQUAL 1940)
+        string(APPEND CEMU_LINKER_FLAGS " /INCREMENTAL:NO")
```
Crementif, line 79: Please keep the comment description. Also, put this in a new if statement and use https://cmake.org/cmake/help/latest/command/set_property.html instead with the append mode. It'll look better. Also, why do we need /NODEFAULTLIB:MSVCRT. Is this MSYS2 related? A short comment explaining that wouldn't hurt, I think.

CMakeLists.txt
```diff
@@ -261,6 +286,9 @@ else()
 endif()
 if(CEMU_ARCHITECTURE MATCHES "(aarch64)|(AARCH64)|(arm64)|(ARM64)")
 	add_subdirectory("dependencies/xbyak_aarch64" EXCLUDE_FROM_ALL)
+	if(MSVC)
+        set_target_properties(xbyak_aarch64 PROPERTIES COMPILE_FLAGS "/wd4245 /wd4458")
```
Crementif: set_target_properties is quite disruptive, since it doesn't append compilation flags but rather replaces them entirely. I think you meant to use set_property with the append option?

src/resource/installer.nsi
```diff
@@ -19,38 +23,33 @@ ManifestDPIAware true
...
 
-; License page
 !insertmacro MUI_PAGE_LICENSE "..\..\LICENSE.txt"
-; Desktop Shortcut page
 Page custom desktopShortcutPageCreate desktopShortcutPageLeave
-; Directory page
 !insertmacro MUI_PAGE_DIRECTORY
-; Instfiles page
 !insertmacro MUI_PAGE_INSTFILES
-; Finish page
```
Crementif, line 46: Just to repeat it, but its really better to just keep things as-is if you can. It just makes reviewing more tedious :)

src/Cafe/HW/Espresso/PPCTimer.cpp
```diff
@@ -123,40 +113,37 @@ void PPCTimer_waitForInit()
...
-	diff.low = _umul128(rdtscDif, Espresso::CORE_CLOCK, &diff.high);
+	diff.low = Multiply64to128(rdtscDif, Espresso::CORE_CLOCK, &diff.high);
 
 	if(rdtscCurrentMeasure > _rdtscLastMeasure)
-		_rdtscLastMeasure = rdtscCurrentMeasure; // only travel forward in time
+		_rdtscLastMeasure = rdtscCurrentMeasure; 
 
 	uint8 c = 0;
-	#if BOOST_OS_WINDOWS
+	#if defined(_MSC_VER)
```
Crementif: What is the difference between these two? Is this a MSYS2 thing?

src/Cafe/HW/Latte/Core/LatteTextureCache.cpp
```diff
@@ -146,7 +146,7 @@ uint32 LatteTexture_CalculateTextureDataHash(LatteTexture* hostTexture)
 		bool isCompressedFormat = hostTexture->IsCompressedFormat();
 		if( isCompressedFormat == false )
 		{
-#if BOOST_OS_WINDOWS
+#if BOOST_OS_WINDOWS && defined(_M_X64)
```
Crementif, line 149: I think the macro you used previously in this code, ARCH_X86_64, was clearer. Assuming they're the same? Is this a windows-only macro?

src/Cafe/OS/common/OSCommon.h
Crementif, line 1: We already had memcpy_dword implemented. Or maybe I'm missing something. But if its not supported on MSVC ARM, then I think its better to provide the definition of memcpy_dwords as-is, instead of making a new macro system. Just makes the code diff smaller and more readable.

dist/windows/Cemu.manifest
```diff
@@ -1,16 +1,29 @@
...
+            <requestedPrivileges>
+                <requestedExecutionLevel level="asInvoker" uiAccess="false"></requestedExecutionLevel>
+            </requestedPrivileges>
+        </security>
     </trustInfo>
     <application xmlns="urn:schemas-microsoft-com:asm.v3">
         <windowsSettings>
+            <!-- Modern DPI awareness setting (covers Windows 7 through 11) -->
             <dpiAware xmlns="http://schemas.microsoft.com/SMI/2005/WindowsSettings">True/PM</dpiAware>
+            <dpiAwareness xmlns="http://schemas.microsoft.com/SMI/2016/WindowsSettings">PerMonitorV2</dpiAwareness>
```
Crementif: Might be better in another PR, unless its not necessary.

dependencies/ih264d/CMakeLists.txt
Crementif, line 1: I think this whole file was rewritten by AI? A lot of changes, removal of comments and making things shorter? I'm not sure what you changed. It seems like a lot of files are now no longer compiled? I would prefer to keep the source code of dependencies the same, if that's possible.

dependencies/ih264d/common/armv8/ih264_platform_macros.h
Crementif, line 1: I wonder if there's some upstream changes that are merged since we forked it. Would be more ideal to not diverge from that source code.

Crementif, line 1: Did you check if ih264d had any upstream changes to make it compile-able on MSVC ARM?

src/Cafe/HW/Espresso/Debugger/Debugger.cpp
```diff
@@ -235,9 +235,51 @@ void debugger_updateMemoryBreakpoint(DebuggerBreakpoint* bp)
...
+            else
+            {
+                // Disable breakpoint slot 0
+                ctx.Bvr[0] = 0;
+                ctx.Bcr[0] = 0;
+            }
+            SetThreadContext(hThread, &ctx);
+        }
+        ResumeThread(hThread);
+    }
```
Crementif, line 279: You added support for setting the breakpoint registers for the debugger. But when I check the ExceptionHandler_win32.cpp code, I'm quite sure that this is not actually handled? I feel like this should've come up if you had tested it, which would've been nice. You can use the PPC debugger to make a write/read breakpoint in the PPC debugger at some memory address.

Crementif (changes requested):
NOTE: I see that this has now been marked as a draft PR. But since I started reviewing this before then. I hope my feedback will guide you through a compatible solution that can be merged, because I think as-is there's some concerns.

---

Thanks for making this PR. It would definitely be nice to support windows ARM support in this project, with the assumption that the code is properly written and doesn't increase complexity for other architectures. So the fixes in this PR would ideally be as small as possible, and extend our existing ARM code so that its arm support extends to Windows. I do think this PR is kind of chaotic. From what I can see, a big issue seems to be the assembly, at least with the dependencies. Its preferable if we used the upstream version.

I did see a number of tells that a decent amount of AI usage was involved with these changes. Although its fine to use AI while writing code, we have some hesitations about PRs like this that change existing code in multiple places, potentially affecting other platforms as well. Reviewing code takes a good chunk of time away from working on our own projects within Cemu, and with the increase of AI PRs, we generally will be quite conservative with what to merge.

So, ideally in the next PR, please declare whether you've used AI, and if yes, know that we expect you to understand the changes it has made and whether those changes match the quality of what you or any other maintainer would've written manually. This includes checking the coding guidelines, bundling up unrelated changes and making a precise, targeted PR.

src/Cafe/HW/Espresso/Recompiler/IML/IMLInstruction.h
Exzap: `operation` is actually a uint8, so assigning -999 causes a warning. It's just something we didn't bother to clean up yet. I will say that personally I prefer feature PRs to be focused on the feature and not randomly clean up or refactor other things unless it relates to the feature in some way. Probably the LLM saw the warning and decided to take care of it?

src/Cafe/HW/Espresso/PPCTimer.cpp
qurious-pixel: the MSVC ARM code does not have the headers as-is with the Windows ARM runner. I can "install" them in to the image and use as is, but it will fail to link using the reusable code.

src/Cafe/HW/Espresso/Recompiler/IML/IMLInstruction.h
qurious-pixel: I have seen this error for a long time, and was trying to clean up some of the noise when building. can revert as before. just noise.

qurious-pixel:
> NOTE: I see that this has now been marked as a draft PR. But since I started reviewing this before then. I hope my feedback will guide you through a compatible solution that can be merged, because I think as-is there's some concerns.
> 
> Thanks for making this PR. It would definitely be nice to support windows ARM support in this project, with the assumption that the code is properly written and doesn't increase complexity for other architectures. So the fixes in this PR would ideally be as small as possible, and extend our existing ARM code so that its arm support extends to Windows. I do think this PR is kind of chaotic. From what I can see, a big issue seems to be the assembly, at least with the dependencies. Its preferable if we used the upstream version.
> 
> I did see a number of tells that a decent amount of AI usage was involved with these changes. Although its fine to use AI while writing code, we have some hesitations about PRs like this that change existing code in multiple places, potentially affecting other platforms as well. Reviewing code takes a good chunk of time away from working on our own projects within Cemu, and with the increase of AI PRs, we generally will be quite conservative with what to merge.
> 
> So, ideally in the next PR, please declare whether you've used AI, and if yes, know that we expect you to understand the changes it has made and whether those changes match the quality of what you or any other maintainer would've written manually. This includes checking the coding guidelines, bundling up unrelated changes and making a precise, targeted PR.

Thanks for the feedback. Set this as draft since the Windows on Arm code for MSVC is giving some issues i did not experience with Clang Arm for Windows. Specifically a thread lock when closing Cemu. And interaction between WxWidgets windows and the backend.

src/Cafe/CMakeLists.txt
qurious-pixel: Unnecessary after adding fallback code in `Intrinsics.h`

src/Common/Intrinsics.h
```diff
@@ -0,0 +1,101 @@
+#pragma once
```
exverge-0: Intrinsics are already defined in `precompiled.h`, any new ones should probably stay there.

src/Common/Intrinsics.h
```diff
@@ -0,0 +1,101 @@
...
+        }
+        if (remainder) *remainder = r;
+        return q;
+    #endif
+#else
+    unsigned __int128 dividend = ((unsigned __int128)high << 64) | low;
+    if (remainder) *remainder = (uint64)(dividend % denominator);
+    return (uint64)(dividend / denominator);
+#endif
+}
```
exverge-0: `_udiv128` and `_umul128` is an MSVC intrinsic which should be avaliable for aarch64, and is already implemented for other compilers in precompiled.hpp otherwise

CMakeSettings.json
```diff
@@ -16,6 +16,14 @@
...
       "installRoot": "${projectDir}\\out\\install\\${name}"
     },
+    {
+      "name": "ARM64-Release",
+      "configurationType": "Release",
+      "generator": "Ninja",
+      "inheritEnvironments": [ "msvc_arm64_x64" ],
+      "buildRoot": "${projectDir}\\out\\build\\${name}",
+      "installRoot": "${projectDir}\\out\\install\\${name}"
+    },
```
exverge-0: What's different in this vs. `Release`? CMake configs should be compatible cross platforms

CMakeLists.txt
```diff
@@ -51,8 +51,16 @@ if (ENABLE_VCPKG)
 	# CONFIG option
 	set(CMAKE_FIND_PACKAGE_PREFER_CONFIG TRUE)
 	if (WIN32)
-		set(VCPKG_TARGET_TRIPLET "x64-windows-static" CACHE STRING "")
-	endif()
+		message(STATUS "CPU Arch is: $ENV{PROCESSOR_ARCHITECTURE}")  
+		if($ENV{PROCESSOR_ARCHITECTURE} MATCHES "x86_64|amd64|AMD64")
```
exverge-0: Should probably use `CEMU_ARCHITECTURE` for consistency

src/Common/precompiled.h
```diff
@@ -341,48 +341,70 @@ inline uint64 _udiv128(uint64 highDividend, uint64 lowDividend, uint64 divisor,
...
 inline void _mm_mfence()
 {
-	asm volatile("" ::: "memory");
-	std::atomic_thread_fence(std::memory_order_seq_cst);
+#if defined(_MSC_VER)
+    __dmb(_ARM64_BARRIER_ISH); // Inner Shareable Data Memory Barrier
+#else
+    asm volatile("" ::: "memory");
+    std::atomic_thread_fence(std::memory_order_seq_cst);
+#endif
```
exverge-0: `memory_order_seq_cst` should compile to `ISH` anyways so this is redundant here, just use C++'s atomic fence
```suggestion
    std::atomic_thread_fence(std::memory_order_seq_cst);
```

src/Cafe/HW/Espresso/PPCTimer.cpp
exverge-0: @qurious-pixel `__rdtsc` is [already implemented](https://github.com/cemu-project/Cemu/blob/326933b2485d87594c5e0d41638a1f0318510a3b/src/Common/precompiled.h#L355-L374) for arm. I'm not sure if MSVC uses the `__aarch64__` preprocessor, but if not you'd have to add the `_M_ARM64` there and wherever else it is used.

dependencies/ih264d/CMakeLists.txt
```diff
@@ -1,204 +1,148 @@
...
+    set(LIBAVCDEC_ARM_INCLUDES "common/armv8" "decoder/arm")
+    include_directories("common/" "decoder/" ${LIBAVCDEC_ARM_INCLUDES})
+
+    if(MSVC)
+        message(STATUS "MSVC ARM64 detected: Forcing portable C implementation.")
+        target_sources(ih264d PRIVATE
+            "decoder/arm/ih264d_function_selector.c"
+            "decoder/arm/ih264d_function_selector_av8.c"
+        )
+        target_compile_definitions(ih264d PRIVATE PORTABLE_C ARCH_GENERIC DISABLE_NEON)
```
exverge-0, line 100: What's the purpose of this? MSVC should be able to support NEON.

src/Common/Intrinsics.h
```diff
@@ -0,0 +1,101 @@
...
+        #define READ_TSC()      _ReadStatusReg(ARM64_CNTVCT_EL0)
+    #else 
+        #define BARRIER_FENCE() __asm__ __volatile__("dmb sy" : : : "memory")
+        inline uint64 READ_TSC() {
+            uint64 virtual_timer;
+            __asm__ __volatile__("mrs %0, cntvct_el0" : "=r" (virtual_timer));
+            return virtual_timer;
+        }
+    #endif
+#endif
```
exverge-0: Also, BARRIER_FENCE is already implemented for arm64 in `_mm_mfence` (which you used in your implementation?) Same with rdtsc and `__rdtsc()`

PR #1916, opened 2026-05-14
debugger: Add support for loading symbols from .map file

Crementif:
To use this feature, you'll want to put a .map file inside `C:\Users\YourUserName\AppData\Roaming\Cemu\debugger`'s folder, or a similar `debugger` folder if you're using the portable Cemu installation. The name of the file is easily found by checking the line in your log.txt file. It'll say something like `Loaded module 'u-king' with checksum 0x6267bfd0`. Its binary name `[module name]_[checksum hash].map`.

image

Also adds support for storing multiple symbols on one address by using a linked list.

### DIFF
--- a/src/Cafe/OS/RPL/rpl_symbol_storage.cpp
+++ b/src/Cafe/OS/RPL/rpl_symbol_storage.cpp
@@ -63,7 +63,7 @@ char* rplSymbolStorage_storeLibname(const char* libName)
 	return libEntry->libName;
 }
 
-RPLStoredSymbol* rplSymbolStorage_store(const char* libName, const char* symbolName, MPTR address)
+RPLStoredSymbol* rplSymbolStorage_store(const char* libName, const char* symbolName, MPTR address, uint32 type)
 {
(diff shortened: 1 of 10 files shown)


src/Cafe/OS/RPL/rpl_symbol_storage.cpp
```diff
@@ -84,10 +84,7 @@ RPLStoredSymbol* rplSymbolStorage_store(const char* libName, const char* symbolN
 RPLStoredSymbol* rplSymbolStorage_getByAddress(MPTR address)
 {
 	std::unique_lock<std::mutex> lck(rplSymbolStorage.m_symbolStorageMutex);
-	auto it = rplSymbolStorage.map_symbolByAddress.find(address);
-	if (it == rplSymbolStorage.map_symbolByAddress.end())
-		return nullptr;
-	return it->second;
+	return rplSymbolStorage.map_symbolByAddress[address];
```
Exzap: Your previous version was actually better. [] operator on an (unordered_)map has the side effect of creating a new default-initialized entry. Same issue in `rplSymbolStorage_getByClosestAddress` (it was like this before, you fixed it but then undid the fix)

PR #1917, opened 2026-05-16
Vulkan: gate transform feedback usage without VK_EXT_transform_feedback

oltolm:
Avoid the validation warning when creating the streamout ring buffer on devices without VK_EXT_transform_feedback.

Warning: vkCreateBuffer(): pCreateInfo->usage has VkBufferUsageFlagBits values (VK_BUFFER_USAGE_TRANSFER_SRC_BIT|VK_BUFFER_USAGE_STORAGE_BUFFER_BIT|VK_BUFFER_USAGE_TRANSFORM_FEEDBACK_BUFFER_BIT_EXT) that requires the extensions VK_EXT_transform_feedback.

### DIFF
--- a/src/Cafe/HW/Latte/Renderer/Vulkan/VulkanRenderer.cpp
+++ b/src/Cafe/HW/Latte/Renderer/Vulkan/VulkanRenderer.cpp
@@ -263,6 +263,14 @@ void VulkanRenderer::GetDeviceFeatures()
 	pwf.pNext = prevStruct;
 	prevStruct = &pwf;
 
+	VkPhysicalDeviceTransformFeedbackFeaturesEXT tff{};
+	if (m_featureControl.deviceExtensions.transform_feedback)
+	{
+		tff.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_TRANSFORM_FEEDBACK_FEATURES_EXT;
+		tff.pNext = prevStruct;
+		prevStruct = &tff;
+	}
+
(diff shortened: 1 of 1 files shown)


goeiecool9999 (approved):
Looks good. Although Cemu currently does not actually use the transform feedback code path it's always good to avoid validation errors. Also, perhaps we should add some logic somewhere so that even in cases where that code path is re-enabled it falls back to the current SSBO implementation by setting `useTFEmulationViaSSBO` to true when the transform feedback feature is not supported?

oltolm:
Sorry, I can not answer that question. I started learning Vulkan, but I can only fix warnings right now.

Exzap:
Vendors generally [don't like transform feedback](https://www.gfxstrand.net/faith/blog/2018/10/transform-feedback-is-terrible-so-why/) and it never became a Vulkan core extension. There is no universal support for it so at this point It's better to just remove the unused transform feedback path in Cemu than to put more energy into it. Feel free to submit a PR for that

Ammar-Sadaoui:
> Vendors generally [don't like transform feedback](https://www.gfxstrand.net/faith/blog/2018/10/transform-feedback-is-terrible-so-why/) and it never became a Vulkan core extension. There is no universal support for it so at this point It's better to just remove the unused transform feedback path in Cemu than to put more energy into it. Feel free to submit a PR for that

http://www.vulkan.gpuinfo.org/displayextensiondetail.php?extension=VK_EXT_transform_feedback

more than 90 %  GPU support this extension

PR #1928, opened 2026-05-23
gui: close dialogs/frames on Esc, add Ctrl+Q exit shortcut

schuay:
Several of Cemu's settings dialogs (GeneralSettings2, InputSettings2,
GraphicPacksWindow2, ...) and most of its tool wxFrames have no
Cancel/Close/OK button. On a Wayland compositor that doesn't draw
server-side decorations there is also no compositor-drawn close button,
so once opened these windows cannot be dismissed.

wxDialog's built-in Esc handling (wxDialogBase::OnCharHook in
src/common/dlgcmn.cpp) forwards Esc to a button with wxID_CANCEL or
the affirmative-id button via EmulateButtonClickIfPresent; with
neither button present, SendCloseButtonClickEvent returns false and
the handler just Skip()s. wxFrame has no Esc handling at all (the
event table in src/common/framecmn.cpp only carries menu-open/close
entries).

Add wxHelper::BindEscapeCloses(target), a one-liner that binds
WXK_ESCAPE on wxEVT_CHAR_HOOK to target->Close(). Close() goes
through the wxEVT_CLOSE_WINDOW path, where wxDialogBase::OnCloseWindow
does fall back to EndDialog(wxID_CANCEL) when no Cancel/OK button
is found - so the helper closes buttonless dialogs correctly. Apply
as the first statement of every wxDialog and wxFrame ctor in
src/gui/wxgui/, except:
  - MainWindow: Esc shouldn't quit Cemu.
  - PadViewFrame: already binds Esc to leave fullscreen.
  - HotkeySettings: inline variant guarded by !m_activeInputButton
    so an in-progress keyboard capture (whose cancel path is
    OnKeyUp + IsValidKeycodeUp rejecting WXK_ESCAPE) still wins.

Tag the File>Exit menu item label with "\tCtrl+Q"; wxMenuItemBase::
GetAccel() parses everything after the tab via wxAcceleratorEntry::
Create and wxMenuBar registers it as an accelerator.

### DIFF
--- a/src/gui/wxgui/AudioDebuggerWindow.cpp
+++ b/src/gui/wxgui/AudioDebuggerWindow.cpp
@@ -3,6 +3,7 @@
 
 #include "Cafe/OS/libs/snd_core/ax.h"
 #include "Cafe/OS/libs/snd_core/ax_internal.h"
+#include "wxgui/helpers/wxHelpers.h"
 
 enum
 {
@@ -24,6 +25,8 @@ wxEND_EVENT_TABLE()
 AudioDebuggerWindow::AudioDebuggerWindow(wxFrame& parent)
 	: wxFrame(&parent, wxID_ANY, _("AX voice viewer"), wxDefaultPosition, wxSize(1126, 580), wxCLOSE_BOX | wxCLIP_CHILDREN | wxCAPTION | wxRESIZE_BORDER)
 {
+	wxHelper::BindEscapeCloses(this);
+
 	wxPanel* mainPane = new wxPanel(this);
 	wxBoxSizer* sizer = new wxBoxSizer(wxVERTICAL);
 	
--- a/src/gui/wxgui/CemuUpdateWindow.cpp
+++ b/src/gui/wxgui/CemuUpdateWindow.cpp
@@ -21,6 +21,7 @@
 #include <curl/curl.h>
 #include <zip.h>
 #include <boost/tokenizer.hpp>
+#include "wxgui/helpers/wxHelpers.h"
 
 
 wxDECLARE_EVENT(wxEVT_RESULT, wxCommandEvent);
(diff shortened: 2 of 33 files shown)


Exzap:
Shouldn't this be handled upstream in wxWidgets/wxGTK?

schuay:
The problem I'm facing is: Cemu dialogs have no close/cancel/ok buttons, and without a compositor-drawn close button and no Esc binding, there is was no way to close the windows and persist settings.

After a second look, the original root cause claim was incorrect. The reason default Esc handling currently doesn't hit is that there is no button with wxID_CANCEL or the affirmative-id button (`EmulateButtonClickIfPresent`), so the handler skips.

https://github.com/wxWidgets/wxWidgets/blob/v3.2.10/src/common/dlgcmn.cpp#L442-L454
https://github.com/wxWidgets/wxWidgets/blob/v3.2.10/src/common/dlgcmn.cpp#L407-L431
https://github.com/wxWidgets/wxWidgets/blob/v3.2.10/src/common/dlgcmn.cpp#L526-L560

Another way to fix this would be to add the cancel/ok buttons, but that is a larger change and requires thinking about layout changes in all these dialogs.

I should also point out that this PR is AI-assisted.

schuay:
@Exzap friendly ping.

Exzap:
You didn't actually answer my earlier question, but either way this workaround does not belong in Cemu. Cemu is requesting a titlebar with close button, it's wxWidgets/wxGTKs job to make sure a close affordance is provided in such a case and not just silently ignored.

schuay:
> You didn't actually answer my earlier question

Maybe not explicitly but my earlier reply does elaborate on current behavior, ie: wxWidgets does provide default Esc handlers but only when cancel/ok buttons exist. In Cemu they don't, so decoration-less compositors have this issue. Imo the upstream behavior is reasonable, and either passing explicit Esc bindings or adding the buttons would be an option here.

PR #1929, opened 2026-05-23
Vulkan: skip zero-size readback buffer barriers

oltolm:
Do not create Vulkan async texture readbacks when the format has no supported linear readback size. The shared readback launcher now tolerates a null readback object, which prevents zero-byte buffer barriers from being submitted for unsupported readback formats.

Warning: vkCmdPipelineBarrier(): pBufferMemoryBarriers[0] VkBuffer 0x1080000000108 has a size of 0. The Vulkan spec states: If size is not equal to VK_WHOLE_SIZE, size must be greater than 0.


Exzap:
The call to `vkGetImageMemoryRequirements` can also be removed now since the result is no longer used

oltolm:
I removed it.

PR #1937, opened 2026-05-30
Clamp OpenGL vsync value in UpdateVSyncState

codeman4033:
Before, this assumed that the vsync config value is only ever 0 or 1. But there's a possibility it isn't because of Vulkan. This becomes a problem in a game profile where the graphics api is OpenGL, but in the global config, its Vulkan. This fixes that issue.

### DIFF
--- a/src/gui/wxgui/canvas/OpenGLCanvas.cpp
+++ b/src/gui/wxgui/canvas/OpenGLCanvas.cpp
@@ -119,7 +119,7 @@ public:
 	void UpdateVSyncState()
 	{
 		int configValue = GetConfig().vsync.GetValue();
-		(value < 0) ? 0 : (value > 1) ? 1 : value;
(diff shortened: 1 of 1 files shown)


src/gui/wxgui/canvas/OpenGLCanvas.cpp
```diff
@@ -119,6 +119,7 @@ class OpenGLCanvas : public IRenderCanvas, public wxGLCanvas
 	void UpdateVSyncState()
 	{
 		int configValue = GetConfig().vsync.GetValue();
+		(value < 0) ? 0 : (value > 1) ? 1 : value;
```
goeiecool9999: The issue raised is valid but your submission does not compile.
I think the best solution would be
```suggestion
		configValue = configValue > 0 ? 1 : 0;
```

PR #1943, opened 2026-06-02
Update to SDL 3.4.10 (and update vcpkg)

GiveMeTheButter:
Since I use a Steam Controller and SDL 3.4.10 has support for it, I decided to try updating it on Cemu myself and it seems to work... No idea what I'm doing but I decided to do a pr!

vcpkg has been updated to its latest commit and not its latest release, which I thought would be good to mention. That was done to be able to get SDL 3.4.10.

### DIFF
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -76,7 +76,7 @@
     },
     {
       "name": "sdl3",
-      "version": "3.4.2"
+      "version": "3.4.10"
     },
     {
       "name": "wxwidgets",


Exzap:
Looks good, but just a fyi: Updating vcpkg has the side effect of also upgrading dependencies that are not pinned to a specific version. In this case it's fine but just something that goes beyond the intended scope of the PR

goeiecool9999:
@Exzap 
> Updating vcpkg has the side effect of also upgrading dependencies that are not pinned to a specific version

Isn't that why we have https://github.com/cemu-project/Cemu/blob/251662aa94754c6c31ea22f151ac432f2e67d874/vcpkg.json#L4
Doesn't that pin the versions independently of the vcpkg commit that's checked out?

Exzap:
Yeah it does. I forgot about that option

PR #1948, opened 2026-06-07
GraphicPack: Allow overriding cos.xml permissions

techmuse8:
Allows overriding cos.xml permission masks in `rules.txt`, using the newly added `Permissions` section.

Example:
```
[Permissions]
FS = 0xFFFFFFFFFFFFFFFF
UHS = 0x100
```

### DIFF
--- a/src/Cafe/CafeSystem.cpp
+++ b/src/Cafe/CafeSystem.cpp
@@ -950,6 +950,23 @@ namespace CafeSystem
 	{
 		if (sLaunchModeIsStandalone)
 			return CosCapabilityBits::All;
+
+		uint64 resultMask = 0;
+		for (const auto& pack : GraphicPack2::GetGraphicPacks())
+		{
+			if (pack->IsEnabled()) 
+			{	
+				for (const auto& permissionOverrides : pack->GetPermissionOverrides()) 
+				{
+					if (permissionOverrides.first == group)
+						resultMask |= permissionOverrides.second;
+				}
+			}
+		}
+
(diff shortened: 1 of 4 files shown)


src/Cafe/CafeSystem.cpp
```diff
@@ -950,6 +950,23 @@ namespace CafeSystem
 	{
 		if (sLaunchModeIsStandalone)
 			return CosCapabilityBits::All;
+
+		uint64 resultMask = 0;
+		for (const auto& pack : GraphicPack2::GetGraphicPacks())
```
Exzap: Use `GraphicPack2::GetActiveGraphicPacks()` here. Then you also don't need to manually check if its enabled

src/Cafe/CafeSystem.cpp
```diff
@@ -950,6 +950,23 @@ namespace CafeSystem
...
+				for (const auto& permissionOverrides : pack->GetPermissionOverrides()) 
+				{
+					if (permissionOverrides.first == group)
+						resultMask |= permissionOverrides.second;
+				}
+			}
+		}
+
+		if (resultMask != 0)
+    		return static_cast<CosCapabilityBits>(resultMask);
```
Exzap: Returning early here means that the permissions bits from the base game or update are dropped. I think it makes more sense to not return here and instead OR the bits together

techmuse8: Did so by implementing `operator|` and `operator|=` overloads for `CosCapabilityBits`

PR #1949, opened 2026-06-07
Fix Windows system version detection

SirHrVedel:
GetVersionEx has been deprecated since Windows 8.1. Thus it reports Windows 8 RTM's info purely for compatibility reasons.

This new method gets the registry keys from "HKLM\SOFTWARE\Microsoft\Windows NT\CurrentVersion" that hold the correct version info. Additionally it now reports the Windows feature version (1809, 22H2, 25H2) present since Windows 10. 
Additionally includes a small fix for systems that upgraded to Windows 11 but still have the ProductName reporting Windows 10 (like in my case).
This method retain compatibility with Windows 7.

### DIFF
--- a/src/Cafe/CafeSystem.cpp
+++ b/src/Cafe/CafeSystem.cpp
@@ -526,7 +526,7 @@ namespace CafeSystem
 			RegCloseKey(hKey);
 		}
 		std::string result(productName);
-		// ProductName can still read "Windows 10" on Windows 11 after an in-place upgrade
+		// ProductName still reads as "Windows 10" on Windows 11. Find and replace with "Windows 11" based on build number.
 		if (buildNumber >= 22000)
 		{
 			size_t pos = result.find("Windows 10");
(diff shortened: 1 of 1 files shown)


Crementif:
We don't have time to manually test each PR with the increased amount of PRs we get these days, so small changes that don't affect actual usage might be declined. It seems like these changes aren't too unsafe, but can you please comment on whether you've tested this implementation (it looks AI written?) to be fully working (like checking your log.txt file, or wherever else this is actually used). Other then that, it seems like a nice change and something we should address probably.

SirHrVedel:
I did test it on my Windows 11 machine, as well on a Windows 7 virtual machine. Works fine on my end (logs included). Did some more research and the ProductName reg key reporting Windows 10. And its seems to be left like that for compatibility reasons, so my assumption was incorrect (Il'l update the comment asap). The fix already in place should report Windows 11 properly by checking the build number.
The only thing i need to confirm test, is the windows 10 pre 20H2 versions, which reports the feature update version using a different reg key as i tried to implement.
Logs:
[logWin7.txt](https://github.com/user-attachments/files/28699194/logWin7.txt)
[logWin11.txt](https://github.com/user-attachments/files/28699252/logWin11.txt)

Exzap:
I think this is good enough as-is, tested it on a Windows 10 22H machine as well and it reports the correct version.

PR #1950, opened 2026-06-07
Debugger: Tweak value writing and add value options

Swiftshine:
Memory dump:
- User can now right-click to open a context menu to select the type of data to write (byte, int16, int32, float, string).; double-clicking functionality is preserved.

Register view + memory dump:
- Integral values entered are no longer forced to be entered in hexadecimal format; user must now specify a `0x` or `0X` prefix to enter in hex. Otherwise, the values are entered in decimal format.
- Integral values that are displayed upon opening the dialog are presented in hexadecimal format.

### DIFF
--- a/src/gui/wxgui/debugger/DumpCtrl.cpp
+++ b/src/gui/wxgui/debugger/DumpCtrl.cpp
@@ -1,5 +1,6 @@
 #include "wxgui/wxgui.h"
 #include "wxgui/debugger/DumpCtrl.h"
+#include "wxgui/helpers/wxHelpers.h"
 #include "Cafe/OS/RPL/rpl.h"
 #include "Cafe/OS/RPL/rpl_structs.h"
 #include "Cafe/HW/Espresso/Debugger/Debugger.h"
@@ -28,6 +29,8 @@ DumpCtrl::DumpCtrl(wxWindow* parent, const wxWindowID& id, const wxPoint& pos, c
 		m_memoryRegion.size = 0x1000;
 		Init();
 	}
+
+	Bind(wxEVT_MENU, &DumpCtrl::OnMenuSelected, this);
 }
 
 void DumpCtrl::Init()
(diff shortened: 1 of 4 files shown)


Crementif:
I see that you included some unnecessary header includes. Just based on past coding reviews, this is a very good tell that a PR is mostly AI generated code, and that it didn't get edited afterwards by a human (which is nice I guess, better then being unaware of these things). Although we do appreciate PRs that upstream features that are usually personal wishes, we do expect all final output to be reviewed by the human. Or else we suffer potentially buggy code becoming a debugging burden on us. Also, please review our coding style guide if you can.

We don't have time to manually test each PR, so the submitter is expected to take care that its a fully working implementation, test their own code to make sure it doesn't break other parts and review their changes. Thanks for taking your time to do so.

Swiftshine:
I appreciate you checking out my PR. Respectfully, however:
- I didn't use AI to write my code (and I personally don't like AI submissions, either, so I understand fully why you'd be hesitant)
- I _did_ test this manually

I'm not really sure what else to say.

Also, which header includes exactly were unnecessary? I'm not really sure what you mean here

Exzap:
I will try to do a proper review tomorrow, but just on a quick glance:
`<concepts>` does not need to be included
You can use `ConvertString` from `util/helpers/helpers.h` for number parsing instead of providing your own helper.

Swiftshine:
> `<concepts>` does not need to be included

Gotcha. I'm used to working with C++98 and don't really have or use these compile-time type checks often because of that, so I wasn't aware that it was only `<type_traits>` that was needed (had to google it). I've just been using `<concepts>` in any non-C++98 code I've written since 1) I thought that's where it was, and 2) it's a C++20 feature, which Cemu uses, so I thought it was appropriate

> You can use `ConvertString` from `util/helpers/helpers.h` for number parsing instead of providing your own helper.

Oh nice, I wasn't aware of that since the original code didn't use it in these files. I'll be sure to use that instead though

src/gui/wxgui/debugger/DumpCtrl.h
```diff
@@ -2,6 +2,14 @@
 #include "wxgui/components/TextList.h"
 
 
+enum {
```
Exzap: You should move this enum into the .cpp file to avoid cluttering the global namespace

src/gui/wxgui/debugger/DumpCtrl.cpp
```diff
@@ -284,3 +290,155 @@ wxSize DumpCtrl::DoGetBestSize() const
 {
 	return TextList::DoGetBestSize();
 }
+
+void DumpCtrl::OnContextMenu(const wxPoint& position, uint32 line)
+{
+	wxPoint pos = position;
+
+	if (pos.x <= OFFSET_ADDRESS + OFFSET_ADDRESS_RELATIVE)
```
Exzap: I assume the plan was to use `PositionToAddress` here, but you missed it by accident?

src/gui/wxgui/debugger/DumpCtrl.cpp
```diff
@@ -284,3 +290,155 @@ wxSize DumpCtrl::DoGetBestSize() const
...
+
+	if (!memory_isAddressRangeAccessible(address, 1))
+		return;
+
+	m_writerContextAddress = address;
+	m_writerContextLine = line;
+
+	wxMenu menu;
+
+	menu.Append(ID_WRITE_U8, "Write Byte");
```
Exzap: These strings should be translatable, wrap them like this: `_("Write Byte")`

Exzap (changes requested):

Exzap:
Looks good. Thanks!

PR #1955, opened 2026-06-13
Clean up includes in MainWindow and remove duplicate helper

goeiecool9999:
Here is a good overview of the differences in MainWindow.cpp
image

CafeSystem is included from MainWindow.h, likewise for LoggingWindow.h
TitleInfo is included from TitleList, which is already included.
If relying on cascading includes like this is undesired I can add them back.

As far as I can tell the following were entirely unused in MainWindow.cpp:
- Cafe/Filesystem/FST/FST.h
- CemuApp.h
- DownloadGraphicPacksWindow.h
- GettingStartedDialog.h
- util/helpers/helpers.h

I've also removed the duplicate to_underlying in favor of the new one in precompiled.h

### DIFF
--- a/src/Cemu/nex/nex.h
+++ b/src/Cemu/nex/nex.h
@@ -1,6 +1,6 @@
 #pragma once
 #include "nexTypes.h"
-#include<mutex>
+#include <mutex>
 
 const int NEX_PROTOCOL_AUTHENTICATION = 0xA;
 const int NEX_PROTOCOL_SECURE = 0xB;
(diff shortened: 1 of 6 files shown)


Exzap:
> If relying on cascading includes like this is undesired I can add them back.

Relying on cascaded headers is acceptable within reason I would say. For example getting TitleInfo via including TitleList seems pretty obvious to me. CafeSystem from MainWindow.h is a little less obvious but still ok. But for example indirectly pulling in utility headers via some of the unrelated main headers would be questionable (although no doubt we are doing this by accident already in some places). Basically we want to avoid making the include structure too brittle, where optimizing an include away from some header breaks the build in a completely different and mostly unrelated place.

There is also the concern of build times, where having less cascaded includes in headers is better.

goeiecool9999:
> There is also the concern of build times, where having less cascaded includes in headers is better.

That is what led me down this track. Excluding precompiled.h MainWindow.cpp expands to ~16MiB of source code after preprocessing.

PR #1959, opened 2026-06-20
add in extra command line flags for testing, usage

### LINKED ISSUE #1958: Adding CLI flags for Better Terminal/Testing Experience
### Your suggestion

I'm writing Wii-U homebrew (and a Wii-U compilation toolchain), and my goal is to be able to have a nice CI/CD testing environment for that homebrew being developed. A large part of wanting new cli flags come from that desire (but I believe these flags will be usable for anyone wanting to be able to run CEMU from a terminal, and not click around the UI a whole bunch, so hopefully not *too* niche, I wanted to file a discussion issue before just sending a PR).

To describe the problems I'm running into.... Today you can kind of do testing in one of two ways:

* Setup your builds so that headers that are not part of WUT/Wii-U specific functions, etc. are included in the testing binary, compile those for your host, and then run it. 
  * Then you're not testing behavior of  that code under big endian (most likely), PPC instructions, or any of the functionalities of say your toolchain like WUT. This makes the testing worthless for a large number of things.
* On the other hand you could build your own test suite that runs as a normal Wii-U RPX, distribute it to some console/cemu, run it, output the test results as some sort of file & check that file locally/in CI.
  * This is an awful lot of work, and would require a completely custom test runner. You wouldn't be able to just drop in GoogleTest, Catch2, or any common testing framework you might be thinking of. Which isn't particularly ideal.

So what would it take to allow a user to "just run Catch2", or integrate with a wide variety of test runners? One of the ideal ways is to figure out how to make CEMU run as a [Crosscompiling Emulator in CMake](https://cmake.org/cmake/help/latest/prop_tgt/CROSSCOMPILING_EMULATOR.html). You can actually do this today with CMake (with the `-g` flag being able to point to an RPX, you can setup CEMU as your crosscompiling emulator and "natively" run tests through CEMU). However, there's some noticeable problems:

- If running the `ctest` command, you don't get logs in your terminal. You'd have to go check some other file, compared to `ctest` normally which outputs the logs of your test results directly in your terminal.
- Many test runners require running the binary once with `--list-tests` (or equivalent) to get a list of tests that can be sharded (then run the tests again in parallel, or sharding out to multiple machines), and since there is no way to pass command line arguments to an RPX through the CEMU CLI, you just can't do this?
  - At least not without forking the discovery process, to output a file on the RPX side, and read a file on the discovery side.

So I'm proposing we add 3 new flags (all of which will allow a great integration with testing & allowing many testing providers to 'just work out of the box', but also provide a much better experience for folks wanting to use the command line):

- `-l`, `--stream-logs`: Output the logs that normally appear in 'Debug' > 'Open Logging Window' directly to `stdout`/`stderr` (will allow things like list-tests, and logs from test runners to appear when you run ctest).
- `--arguments`: Pass arguments directly to launching the rpx that populates it's argc/argv (will make it easier to integrate with `--list-tests`, etc. without having to write out a whole cos.xml that specifies a custom `argstr`).
- `-b`, `--base-output-directory`: change the base output directory, so our test result files can be written to a temporary directory that is easy to remove, and we don't clog up their normal save file directory.

Mythra:
fixes #1958

this adds in a series of extra command line flags that should make it easier to run games directly from the command line. these flags include:

- a way to mount extra host directories (for one off runs where you want to write to a temporary directory)
- a way to forward OSReport/etc. to stderr.
- a way to specify the arguments to a game when launching it directly through the command line.

### DIFF
--- a/src/Cafe/CafeSystem.cpp
+++ b/src/Cafe/CafeSystem.cpp
@@ -685,6 +685,22 @@ namespace CafeSystem
         fsc_unmount("/cemuBossStorage/", FSC_PRIORITY_BASE);
     }
 
+	void MountExtras()
+	{
+		for (const auto& [host_path, emulatedPath] : LaunchSettings::ExtraMounts())
+		{
+			FSCDeviceHostFS_Mount(boost::nowide::narrow(emulatedPath), _pathToUtf8(host_path), FSC_PRIORITY_BASE);
+		}
+	}
+
+	void UnmountExtras()
+	{
+		for (const auto& [_, emulatedPath] : LaunchSettings::ExtraMounts())
+		{
+			fsc_unmount(boost::nowide::narrow(emulatedPath), FSC_PRIORITY_BASE);
+		}
+	}
+
 	PREPARE_STATUS_CODE LoadAndMountForegroundTitle(TitleId titleId)
 	{
         cemuLog_log(LogType::Force, "Mounting title {:016x}", (uint64)titleId);
@@ -838,6 +854,7 @@ namespace CafeSystem
 		if (r != PREPARE_STATUS_CODE::SUCCESS)
 			return r;
 		InitVirtualMlcStorage();
+		MountExtras();
 		return PREPARE_STATUS_CODE::SUCCESS;
 	}
 
@@ -881,6 +898,7 @@ namespace CafeSystem
         // load executable
         PrepareExecutable();
 		InitVirtualMlcStorage();
+		MountExtras();
 		return PREPARE_STATUS_CODE::SUCCESS;
 	}
 
@@ -961,7 +979,12 @@ namespace CafeSystem
 	std::string GetForegroundTitleArgStr()
 	{
 		if (sLaunchModeIsStandalone)
+		{
+			auto optional_arguments = LaunchSettings::StandaloneTitleArguments();
+			if (optional_arguments.has_value())
+				return *optional_arguments;
 			return "";
+		}
 		auto& update = sGameInfo_ForegroundTitle.GetUpdate();
 		if (update.IsValid())
 			return update.GetArgStr();
@@ -1059,6 +1082,7 @@ namespace CafeSystem
         PPCRecompiler_Shutdown();
         GraphicPack2::Reset();
         UnmountCurrentTitle();
+        UnmountExtras();
         MlcStorageUnmountAllTitles();
         UnmountBaseDirectories();
         DestroyMemorySpace();
--- a/src/Cafe/CafeSystem.cpp
+++ b/src/Cafe/CafeSystem.cpp
@@ -687,7 +687,7 @@ namespace CafeSystem
 
 	void MountExtras()
 	{
-		for (const auto& [host_path, emulatedPath] : LaunchSettings::ExtraMounts())
+		for (const auto& [host_path, emulatedPath] : LaunchSettings::CosMounts())
 		{
 			FSCDeviceHostFS_Mount(boost::nowide::narrow(emulatedPath), _pathToUtf8(host_path), FSC_PRIORITY_BASE);
 		}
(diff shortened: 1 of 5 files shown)


src/Cafe/OS/libs/coreinit/coreinit_Misc.cpp
```diff
@@ -544,15 +547,18 @@ namespace coreinit
...
 	{
 		std::unique_lock _l(sCafeConsoleMutex);
 		CafeLogBuffer& logBuffer = getLogBuffer(cafeLogType);
 		// once a line is full or \n is written it will be posted to log
-		auto flushLine = [](CafeLogBuffer& cafeLogBuffer, std::string_view cafeLogName) -> void
+		auto flushLine = [](CafeLogBuffer& cafeLogBuffer, std::string_view cafeLogName, bool forwardToConsole) -> void
 		{
 			cemuLog_log(LogType::CoreinitLogging, "[{0}] {1}", cafeLogName, std::basic_string_view(cafeLogBuffer.lineBuffer.data(), cafeLogBuffer.lineLength));
+			if (forwardToConsole)
+				std::cerr << fmt::format("{0}\n", std::basic_string_view(cafeLogBuffer.lineBuffer.data(), cafeLogBuffer.lineLength));
```
Exzap: It's preferable to directly call `fwrite(msg, 1, len, stdout);` and also do so outside of `flushLine`. The flushLine logic is only there because log.txt has the limitation that we can only write whole lines. 
Additionally `fwrite` is preferable over C++ iostreams because it allows for null characters in the middle of the string and it avoids any kind of extra processing that may cause the output to be modified (relevant when a tool wants to output raw binary). I would also write to stdout for `cafeLogType == CafeLogType::OSCONSOLE` (which for now is the only COS logging channel we support anyway), and maybe later once properly implemented we could direct actual COS error logging to stderr

src/Cafe/CafeSystem.cpp
```diff
@@ -961,7 +979,12 @@ namespace CafeSystem
 	std::string GetForegroundTitleArgStr()
 	{
 		if (sLaunchModeIsStandalone)
```
Exzap, line 984: Any reason to limit custom args to standalone mode?

Mythra: I wasn't exactly sure how it would play with override arguments, or just the argstr of the RPX, and didn't want to conflict with those. Also should it persist? The standalone seemed like a quick and easy way to fulfill my needs, but also not impact user behavior too much.

src/Cafe/OS/libs/coreinit/coreinit_Misc.cpp
Mythra: Sure I can make those changes. I originally had it here in `flushLine` because I wanted to ensure the two matched, but if we're not worried about that I can just write directly.

src/Cafe/CafeSystem.cpp
Exzap: I would just move the check out of the if to the beginning of the function and that is probably fine for all intents and purposes. Being able to replace arguments is useful for games too, some have debug parameters for example.
Override args only apply when launching another title from within an RPX so they don't apply for the first launch anyway.

And yes the custom args probably shouldn't persist beyond the first title being launched. Easiest way to achieve this is to clear `s_standalone_title_arguments` when `ShutdownTitle()` is called

Exzap:
One more nitpick, can you rename the option strings:
`standalone-title-arguments` -> `cos-argstr` (cos.xml calls it "argstr" so I think this is a bit more consistent if we reuse that term)
`extra-mounts` -> `cos-mounts`
A little shorter and I kinda prefer having the cos prefix because it indicates that this applies to the emulated OS and not Cemu itself.

Mythra:
That all makes sense to me, I've made all the associated changes in the latest commit.

PR #1961, opened 2026-06-21
Look for key file next to wud/wux file

JosJuice:
This makes it possible to use dumps from wudump or wudd without having to manually add the keys to keys.txt.

### DIFF
--- a/src/Cafe/Filesystem/FST/FST.cpp
+++ b/src/Cafe/Filesystem/FST/FST.cpp
@@ -200,7 +200,7 @@ bool FSTVolume::FindDiscKey(const fs::path& path, NCrypto::AesKey& discTitleKey)
 	if (dataSource->readData(0, 0, 0x18000 + 0x100, header, sizeof(header)) != sizeof(header))
 		return false;
 
-	// try all the keys
+	// try all the keys in the key cache
 	uint8 headerDecrypted[sizeof(header)-16];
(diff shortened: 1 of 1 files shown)


src/Cafe/Filesystem/FST/FST.cpp
```diff
@@ -215,7 +215,12 @@ bool FSTVolume::FindDiscKey(const fs::path& path, NCrypto::AesKey& discTitleKey)
 			return true;
 		}
 	}
-	return false;
+
+	// check for a key file next to the WUD file
+	fs::path keyPath = path;
+	keyPath.replace_extension(L"key");
```
Exzap:
```suggestion
	keyPath.replace_extension("key");
```
Since the extension doesn't contain any unicode characters I would just make it an ordinary string literal

src/Cafe/Filesystem/FST/FST.cpp
```diff
@@ -215,7 +215,12 @@ bool FSTVolume::FindDiscKey(const fs::path& path, NCrypto::AesKey& discTitleKey)
...
+	FileStream* keyFile = FileStream::openFile2(keyPath);
+	return keyFile && keyFile->readData(discTitleKey.b, sizeof(discTitleKey.b)) == sizeof(discTitleKey.b);
```
Exzap, line 223: Missing `delete keyFile;` (or you can make it unique_ptr) Currently you are leaking the object

PR #1963, opened 2026-06-23
allow `coreinit.exit` to actually "exit"

### LINKED ISSUE #1960: coreinit.exit doesn't actually exit properly
### Current Behavior

As mentioned in #1958 , I am working on trying to make the command line experience better for running one-off titles within CEMU, specifically for testing binaries. While doing so I noticed that `Cemu_debug` will _always_ exit with an error (because of debug assert false), and `Cemu_release` will always "loop forever" (meaning an external process has to kill Cemu). This means tools by default still think they failed (even if the underlying RPX exited with a successful status code). Meaning some of the workflows I'd like to do don't work.

See the implementation of: https://github.com/cemu-project/Cemu/blob/079a4af651f8a8e33c788c6e9b4a3545ebedb4d3/src/Cafe/OS/libs/coreinit/coreinit.cpp#L241-L248

### Expected Behavior

When an RPX exits, CEMU should not loop forever, or abort unexpectedly. Either it should exit the process with the return code of the RPX (or just stop emulating the game if not running standalone).

### Steps to Reproduce

Attempt to run an RPX that exits. I've compiled an RPX that I've attached to this issue which contains the following code (compiled with WUT -- the devkitPro/homebrew toolchain):

```c
#include <coreinit/thread.h>
#include <coreinit/time.h>

#include <whb/log.h>
#include <whb/log_console.h>
#include <whb/proc.h>

int
main(int argc, char **argv)
{
   int last_tm_sec = -1;
   OSCalendarTime tm;

   WHBProcInit();
   WHBLogConsoleInit();
   WHBLogPrintf("Hello World!");

   WHBLogPrintf("Exiting... good bye.");
   WHBLogConsoleDraw();
   OSSleepTicks(OSMillisecondsToTicks(1000));

   WHBLogConsoleFree();
   WHBProcShutdown();
   return 0;
}
```

That file is uploaded here for convenience:
[exits-rpx.zip](https://github.com/user-attachments/files/29167778/exits-rpx.zip)

On a release build of CEMU, CEMU will stop responding (because it is sleeping forever), on a debug build of CEMU you will get an assert failure leading to a signal abort:

```
RPXHash: 0xae93c5cc
Trace/breakpoint trap!
Error: signal 5:
/usr/lib/libc.so.6(+0x9a29c) [0x7fda9269a29c]
/usr/lib/libc.so.6(+0x3e8f0) [0x7fda9263e8f0]
/usr/lib/libc.so.6(+0x9a29c) [0x7fda9269a29c]
/usr/lib/libc.so.6(gsignal+0x20) [0x7fda9263e7d0]+0x20) [0x7fda9263e7d0]
./bin/Cemu_debug(cemu_assert_debug(bool)+0x27) [0x557da097d837]
./bin/Cemu_debug(coreinit::coreinit_exit(unsigned int)+0x5a) [0x557da0baee7a]
./bin/Cemu_debug(void std::__invoke_impl<void, void (*)(unsigned int), unsigned int&>(std::__invoke_other, void (*&&)(unsigned int), unsigned int&)+0x1f) [0x557da0bb2d7f]
./bin/Cemu_debug(std::__invoke_result<void (*)(unsigned int), unsigned int&>::type std::__invoke<void (*)(unsigned int), unsigned int&>(void (*&&)(unsigned int), unsigned int&)+0x1d) [0x557da0bb2d2d]
./bin/Cemu_debug(decltype(auto) std::__apply_impl<void (*)(unsigned int), std::tuple<unsigned int>&, 0ul>(void (*&&)(unsigned int), std::tuple<unsigned int>&, std::integer_sequence<unsigned long, 0ul>)+0x2d) [0x557da0bb2cfd]
./bin/Cemu_debug(decltype(auto) std::apply<void (*)(unsigned int), std::tuple<unsigned int>&>(void (*&&)(unsigned int), std::tuple<unsigned int>&)+0x1d) [0x557da0bb252d]
./bin/Cemu_debug(_Z21cafeExportCallWrapperITnDaXadL_ZN8coreinit13coreinit_exitEjEEZNS0_14InitializeCoreEvE13StringWrapper_7L7LogType16EEvP16PPCInterpreter_t+0x13a) [0x557da0bb088a]
./bin/Cemu_debug(PPCInterpreter_virtualHLE(PPCInterpreter_t*, unsigned int)+0x6f) [0x557da0ef2b0f]
./bin/Cemu_debug(PPCInterpreterContainer<PPCItpCafeOSUsermode>::executeInstruction(PPCInterpreter_t*)+0xa2) [0x557da0a71c22]
./bin/Cemu_debug(PPCInterpreterSlim_executeInstruction(PPCInterpreter_t*)+0x15) [0x557da0a713f5]
./bin/Cemu_debug(coreinit::__OSFiberThreadEntry(void*)+0x9e) [0x557da0d1e0de]
/usr/lib/libc.so.6(+0x5ab10) [0x7fda9265ab10]
```

### System Info (Optional)

OS: Arch Linux (kernel 7.0.12-arch1-1)
GPU: AMD Radeon RX 7900 XTX

_No response_

### Logs (Optional)

from a copy of Cemu_debug: [log.txt](https://github.com/user-attachments/files/29167847/log.txt)

Mythra:
fixes #1960

this commit implements one particular approach for exiting emulation when the RPX being emulated calls `coreinit.exit`. I don't believe this might be the best way long term, but it seemed to be the best way given the state (and my knowledge of the codebase). This threads through 'End Emulation' (normally a debug only feature in the Cemu menu) into one more spot in the codebase, when `coreinit.exit` is called. Why did we go this way?

Stopping PPC Emulation _while_ the emulator is actively executing an instruction (the call to `coreinit.exit`) seemed like a path of el derado everytime i tried it. calling `Shutdown`, or `ShutdownTitle` is not ideal because it doesn't do things like update status in discord or any of the other window actions, and they lock up because the emulator is already running the coreinit instructions (even if you launch it in a delayed thread, the PPC emulator is still doing stuff).

so the only way to get this to somewhat work is _to reach out to the window system_. So I threaded through an `EndEmulation` request, which started off just calling `EndEmulation` that already existed. There is a note on that function about how it memory leaks if called repeatedly, and it's "not finished". However, this seemed better than looping forever, or just crashing with a debug assert. At least you still have a semi-functional window (even if memory is leaking), and a log window you can investigate?

However, just calling endemulation didn't exit when Cemu was launched from the command line (my goal), it just returns to the main Cemu window. To make Cemu usable as a command line app that you can fire and forget, I threaded a new variable through, that when end emulation is called and the title was launched with `-g`, or `-t`, it just closes cemu altogether. this allows my CLI commands to "just work"

### DIFF
--- a/src/Cafe/CafeSystem.cpp
+++ b/src/Cafe/CafeSystem.cpp
@@ -650,22 +650,23 @@ namespace CafeSystem
 		s_implementation = impl;
 	}
 
-    void Shutdown()
-    {
-        cemu_assert_debug(s_initialized);
-        // if a title is running, shut it down
-        if (sSystemRunning)
-            ShutdownTitle();
-        // shutdown persistent subsystems (deprecated manual shutdown)
+	void Shutdown()
+	{
+		if (!s_initialized)
+			return;
+		// if a title is running, shut it down
+		if (sSystemRunning)
+			ShutdownTitle();
+		// shutdown persistent subsystems (deprecated manual shutdown)
 		iosu::odm::Shutdown();
 		iosu::act::Stop();
-        iosu::mcp::Shutdown();
-        iosu::fsa::Shutdown();
+		iosu::mcp::Shutdown();
+		iosu::fsa::Shutdown();
 		// shutdown IOSU modules
 		for(auto it = s_iosuModules.rbegin(); it != s_iosuModules.rend(); ++it)
 			(*it)->SystemExit();
-        s_initialized = false;
-    }
+		s_initialized = false;
+	}
 
 	std::string GetInternalVirtualCodeFolder()
 	{
@@ -1086,6 +1087,12 @@ namespace CafeSystem
 		DestroyMemorySpace();
 		LaunchSettings::ClearCosArgstr();
 		sSystemRunning = false;
+	}
+
+	void RequestEndTitle(int exitCode)
+	{
+		WindowSystem::SetExitCode(exitCode);
+		WindowSystem::RequestEndCurrentEmulation();
 	}
 
 	/* Virtual mlc storage */
--- a/src/Cafe/CafeSystem.h
+++ b/src/Cafe/CafeSystem.h
@@ -23,7 +23,7 @@ namespace CafeSystem
 
 	void Initialize();
 	void SetImplementation(SystemImplementation* impl);
-    void Shutdown();
+	void Shutdown();
 
 	PREPARE_STATUS_CODE PrepareForegroundTitle(TitleId titleId);
 	PREPARE_STATUS_CODE PrepareForegroundTitleFromStandaloneRPX(const fs::path& path);
@@ -45,6 +45,7 @@ namespace CafeSystem
 	CosCapabilityBits GetForegroundTitleCosCapabilities(CosCapabilityGroup group);
 
 	void ShutdownTitle();
+	void RequestEndTitle(int exitCode);
 
 	std::string GetMlcStoragePath(TitleId titleId);
 	void MlcStorageMountAllTitles();
--- a/src/Cafe/HW/Latte/Renderer/Vulkan/VulkanRenderer.cpp
+++ b/src/Cafe/HW/Latte/Renderer/Vulkan/VulkanRenderer.cpp
@@ -364,7 +364,7 @@ int BreathOfTheWildChildProcessMain()
 	vkEnumeratePhysicalDevices(instance, &count, nullptr);
 
 	vkDestroyInstance(instance, nullptr);
-	return 0;
+	return WindowSystem::GetExitCode();
 }
 
 static void LinuxBreathOfTheWildWorkaround(VkInstance& instance, const VkInstanceCreateInfo* create_info)
--- a/src/Cafe/OS/libs/coreinit/coreinit.cpp
+++ b/src/Cafe/OS/libs/coreinit/coreinit.cpp
@@ -2,6 +2,7 @@
 #include "Common/SysAllocator.h"
 #include "Cafe/OS/RPL/rpl_symbol_storage.h"
 
+#include "Cafe/CafeSystem.h"
 #include "Cafe/OS/libs/coreinit/coreinit_Misc.h"
 
 // includes for Initialize coreinit submodules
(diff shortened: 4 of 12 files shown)


Crementif:
The reason why coreinit.exit() isn't accurately emulated is because some games will call this right after switching to another title, like the manual in XCX if I remember correctly. And its really easy to trigger that in that game, or something. But maybe @Exzap can correct me.

Just for the record. But quite a few subsystems in Cemu aren't properly able to free/reset their resources. Hence why the Exit Title option in Cemu is only enabled in debug builds of Cemu, and should not be used; it'll lead to people getting bugs that wouldn't be there if they just exited Cemu entirely. Definitely something that you could improve upon, and we do want to make this better, but limited time on our end has only allowed for slow progress on that front.

Instead of using exit(1) to close Cemu entirely, I think it'd be more accurate to fix/add [OSShutdown](https://github.com/devkitPro/wut/blob/6f9a38068d8feec5a315c99845053165eb5907e5/include/coreinit/launch.h#L48) instead since its more accurate to the console. But, if you just wanted to close Cemu from PowerPC code, you can also write to a nullptr to crash the process inside the app if you're fine with some hacky code just for the sake of automating something.

Mythra:
I assume in the case of XCX (and similar games), the work picks up in other threads (and hopefully gets cleaned up eventually)? If that’s the case maybe we can guard it behind the main thread calling `coreinit.exit`?

Yeah I agree long term, it would be good to fix exit, and clean to actually clean resources. I’d even be happy to slowly take that on, but my hope was we wouldn’t have to block this use case with “proper cleanup long term”. Maybe that is just wishful thinking on my part. Maybe we can add a hidden CLI flag that allows this behavior?

I’d like to try and avoid the writing address to a nullptr if possible, that doesn’t feel like it’d be a good fit, and I should be clear exit is only called when a user specifically requested to launch only a single game. (And in reality it just shuts down the main window, it doesn’t actually call Exit(1))

Exzap:
This PR is appreciated, however I ended up implementing this myself because it's tricky to weave this into the correct systems. The commit is 53a74e44df53865e9471f8139e5dd93518f7da14. Let me know if there are any issues with it, but I tested it with a custom homebrew application and it worked. On a side note: Cemu will only close and return the status code if you use -g or -t launch parameters. 

> The reason why coreinit.exit() isn't accurately emulated is because some games will call this right after switching to another title, like the manual in XCX if I remember correctly. And its really easy to trigger that in that game, or something. But maybe @Exzap can correct me.

I am pretty sure this used to be a bug where title switching didn't work and game's then responded to the error by calling exit. Nowadays the game just ceases execution at the point of the switch, so even if there is an exit afterwards it will not reach it (and we don't support switching back).

Mythra:
Hey you won’t hear me complaining about someone else shepherding it through 😄 , I’ll confirm with my homebrew application that everything works, and if so close out the issue (and move any bug noted there).

appreciate the quick fix!

PR #1971, opened 2026-06-29
Add mouse to wiimote pointer feature

EddyFrc:
This adds a "Mouse" controller, which only serves to provide _position_ information. The result is that the Wiimote pointer can now be emulated using the mouse pointer's position.

Quick test :
- Associate the mouse with an emulated Wiimote :
  - Add a "Mouse" controller under the "Mouse" API
  - Click on "Settings" while the Mouse controller is selected
  - Check "Use position"
  - Click on "OK"
- Launch a game that uses the Wiimote pointer (like Pikmin 3)

The mouse should be now controlling the Wiimote pointer inside the render canvas.

Feedback would be greatly appreciated. Don't hesitate to ask modifications if my solution doesn't meet the standards of the project. 

edit: tested lightly on Linux and Windows, I think it is ready.

### DIFF
--- a/.gdbinit
+++ b/.gdbinit
@@ -1,0 +1,47 @@
+python
+
+import gdb
+import re
+
+
+class BetypePrinter:
+    PATTERN = re.compile("^betype<.*>$")
+
+    def __init__(self, obj):
+        self._obj = obj
+
+    def to_string(self):
+        underlying = self._obj["m_value"]
+        reversed_bytes = bytes(reversed(underlying.bytes))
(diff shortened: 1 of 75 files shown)


Exzap:
Thanks for the contribution, just a quick heads up that I am pretty busy this week so reviewing will take a bit of time.

EddyFrc:
> Thanks for the contribution, just a quick heads up that I am pretty busy this week so reviewing will take a bit of time.

Yeah no problem 👍

PR #1974, opened 2026-07-01
CloudSync: pull save from Dropbox on init, push after; add Linux/AppI…

gdiazbanuelos:
…mage rclone support

Pushes the local save directory to Dropbox via rclone after SAVEInit, and pulls it down first if no local save exists yet (e.g. fresh install or after wiping local saves). Shows in-game overlay notifications for both.

Adds a Linux implementation alongside the existing Windows one, using posix_spawn with a hardened rclone lookup (probes common install paths plus a /proc/1/root fallback to escape the AppImage's mount namespace) and a stripped child environment (drops LD_LIBRARY_PATH/APPDIR/etc.) so the AppImage's bundled libs don't leak into rclone's process.

### DIFF
--- a/src/Cafe/OS/libs/nn_save/nn_save.cpp
+++ b/src/Cafe/OS/libs/nn_save/nn_save.cpp
@@ -6,11 +6,24 @@
 
 #include <filesystem>
 #include <sstream>
+#include <thread>
 #include "config/ActiveSettings.h"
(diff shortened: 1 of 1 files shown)


gdiazbanuelos:
Sorry meant to push to my own fork. This can be ignored.

PR #1982, opened 2026-07-07
link SDL3 earlier

Samueru-sama:
This is needed to fix builds on archlinux with wxWidgets.

wxWidgets still depends on SDL2, which causes conflicts with SDL3.

See also: 

* https://github.com/pkgforge-dev/Cemu-AppImage-Enhanced/issues/17
* https://github.com/pkgforge-dev/Cemu-AppImage-Enhanced/blob/main/link-sdl3-before-wxwidgets.patch

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -211,6 +211,10 @@ set_target_properties(CemuBin PROPERTIES
 	OUTPUT_NAME "${OUTPUT_NAME}"
 )
 
+if(ENABLE_SDL)
+	target_link_libraries(CemuBin PRIVATE SDL3::SDL3)
+endif()
+
 target_link_libraries(CemuBin PRIVATE
 	CemuAudio
 	CemuCafe
@@ -221,10 +225,6 @@ target_link_libraries(CemuBin PRIVATE
 	CemuInput
 	CemuUtil
 )
-
-if(ENABLE_SDL)
-	target_link_libraries(CemuBin PRIVATE SDL3::SDL3)
-endif()
 
 if(UNIX AND NOT APPLE)
 	# due to nasm output some linkers will make stack executable


Exzap:
This workaround hides the immediate crash but it does not fix the underlying problem. Cemu uses SDL3 and wxWidgets has a dependency on SDL2. Using both in the same binary will cause symbol collision. The crash is caused because Cemu ends up using mixed SDL2/SDL3 functions due to that. If you change the link order then Cemu seems to use SDL3 correctly, but wxWidgets will then use mixed SDL2/SDL3 API and have the same crashing problem if it ever ends up initializing SDL.

The proper solution is that you compile wxWidgets without SDL support. As far as I know SDL is only used by wxWidgets for it's wxSound implementation, which Cemu does not use directly but it could still be initialized indirectly. wxWidget's SDL dependency is optional and can be disabled using `--without-sdl` when building with configure or `-DwxUSE_LIBSDL=OFF` when building with CMake (this is what our vcpkg build does). Since your build already compiles wxgtk-git from what I can see, it should be possible to change its build options?

Samueru-sama:
> The proper solution is that you compile wxWidgets without SDL support.

That fix would work for us but not for other people packaging cemu.

The [AUR package](https://aur.archlinux.org/packages/cemu-git) has people comenting that you need to preload `libSDL3.so.0` to fix crashes and it is likely the same issue that this PR would fix.

Building `wxwidgets` without SDL is likely to cause issues for other applications.

---

Anyways, I'm personally gonna keep using that patch unless things begin to break again. And tbh the proper proper fix would be to move away from `wxwidgets` in general, [this is not the first time this library has caused us issues.](https://gitlab.archlinux.org/archlinux/packaging/packages/wxwidgets/-/work_items/7#note_371398)

Exzap:
I'll merge it because its not really an intrusive change anyway, and most likely wxWidgets won't actually call any SDL functions since we are not utilizing wxSound. So in practical terms this is a tolerable workaround.

PR #1984, opened 2026-07-09
Fix Mii Maker image export feature

ariankordi:
This PR gets rid of the "please insert an SD card" error that [previously showed up](https://discord.com/channels/286429969104764928/1017917335874572308/1259142305642840154) if you tried to use this feature, so that it now works and is fully preserved (like it is in Citra).
￼
Saved image of

Mii QR codes can also export now, albehit the data is all zeroes because Mii encryption is unimplemented:
Pasted Graphic 2

## Changes made

* 345ba088d1972f70374dcb98b4253cfadc632590: Invoke SD card callback/notification when FSAddClientEx is called.
  - If Mii Maker never receives this, it doesn't even try to mount the SD card despite it being present.
  - The workaround I applied is to invoke the callback immediately with the string "sdcard".
    * It’s possible this affects other games that use the SD card, if… there are any.
  - That `FSAttachCallback` is slightly [documented in Decaf](https://github.com/decaf-emu/decaf-emu/blob/e6c528a20a41c34e0f9eb91dd3da40f119db2dee/src/libdecaf/src/cafe/libraries/coreinit/coreinit_fs_client.cpp#L86-L102).
* 1e1b2c14c51808d363652e2315afb366a2e38a4d: Fix image texture readback in Mii Maker
  - As you can see, I had to remove an assertion that got hit while doing this. Context is needed on why it was originally added.
  - Also had to enable `TM_LINEAR_GENERAL` to be mirrored to CPU RAM as well as `TM_LINEAR_ALIGNED`.
  - Even with these changes, JPEGs for QR codes are still completely black under the Metal renderer while Vulkan is fine.
* 244aeb00dd42e9276dd499eb7318e8d28ede85f4: Extra change I snuck in to export `FSRewindDir`.
    * Already implemented, called by "FFLUtility". (Search in RGD Discord for more info)
        * According to the paths, it's built from the [same sources as Mii Maker](https://discord.com/channels/370011926823960576/374765704877965312/1296536499860410449). So yes, Mii
    * This patch and a mount at /fflutility will make the tool work (`--cos-mounts /path/to/hostio:/fflutility`)
        * Weirdly not required on a real console.
    * DM me if you'd like to test this.

## Fixing QR codes

Proper QR code exports involve implementing ACPMiiWrap/Unwrap, which I could do, but is out of scope and well documented anyway.
* Used for Mii QR code encryption yes, but also for some obscure Mii network features.
    * sonic2013.rpx: `WrapStoreData__31__N_15_FFLiNetwork_cpp_5a76681bFP20FFLiWrappedStoreDataPC16FFLiStoreDataCFLb`
* The concern here is with [the AES key it needs](https://github.com/azahar-emu/azahar/blob/0c19c33bd339827d35b15420f74e37c260e249ea/src/core/hle/service/apt/apt.cpp#L1351), which Citra uses from the txt.
    * Not my call if it should be included in the code, but does Cemu even require keys anywhere else?
    * Citra actually uses 0 as the key if unspecified. Those null-key QR codes, while mostly useless, CAN actually be read/created by [a tool I made](https://ariankordi.github.io/my-jsfiddles/mii-tomo-dachi-topia-qr-tool/index.html).
* Or we can defer this to until camera support works. Then the entire feature can be tested end-to-end.

### DIFF
--- a/src/Cafe/HW/Latte/Core/LatteTexture.cpp
+++ b/src/Cafe/HW/Latte/Core/LatteTexture.cpp
@@ -1307,8 +1307,9 @@ LatteTexture::LatteTexture(Latte::E_DIM dim, MPTR physAddress, MPTR physMipAddre
 			}
 		}
 	}
-	// determine if this texture should ever be mirrored to CPU RAM
-	if (this->tileMode == Latte::E_HWTILEMODE::TM_LINEAR_ALIGNED)
+	// determine if this texture should ever be mirrored to CPU RAM.
+	if (this->tileMode == Latte::E_HWTILEMODE::TM_LINEAR_ALIGNED ||
+		this->tileMode == Latte::E_HWTILEMODE::TM_LINEAR_GENERAL)
 	{
 		this->enableReadback = true;
 	}
--- a/src/Cafe/OS/libs/coreinit/coreinit_FS.cpp
+++ b/src/Cafe/OS/libs/coreinit/coreinit_FS.cpp
@@ -275,7 +275,37 @@ namespace coreinit
 		cemuLog_log(LogType::Force, "FSShutdown called");
 	}
 
-	FS_RESULT FSAddClientEx(FSClient_t* fsClient, uint32 uknR4, uint32 errHandling)
+	/// Buffer for the notification passed to the FSAddClientEx callback.
+	/// Only the name at 0x18 ("sdcard") is read out of this from Mii Maker.
+	SysAllocator<uint8, 0x40> s_fsClientNotification;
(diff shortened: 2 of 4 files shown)


Exzap:
Appreciate the contribution, but our LLM policy requires all code to be written by a human. You can find the full policy and reasons in [README.md](https://github.com/cemu-project/Cemu/blob/main/README.md)

ariankordi:
Sorry about that, the SD change was from Claude with rewritten comments. I reverted and redid that one by hand.

PR #1986, opened 2026-07-10
nn_fp: Use proper Comment struct and fix invalid Mii name on Linux

primaviera:
This fixes empty comments from appearing as corrupted characters because GetMyComment incorrectly treated the whole comment as a uint16 array (also applied this to friend comments).

| Before | After |
| - | - |
comment_without_pr | comment_with_pr

This also fixes sending invalid Mii names to online servers on Linux since the name is a uint16 array but is casted as a pointer to wchar_t, which is 4 bytes long on Linux when it should be 2.
| Before | After |
| - | - |
nickname_without_pr | nickname_with_pr

The last change is that gamemode descriptions start at 0xE6, not 0xE4, so prepending \0 to it is unnecessary.

### DIFF
--- a/src/Cafe/IOSU/legacy/iosu_fpd.cpp
+++ b/src/Cafe/IOSU/legacy/iosu_fpd.cpp
@@ -68,7 +68,7 @@ namespace iosu
 			std::atomic_uint64_t m_notificationQueueIndex{1};
 		}g_NotificationQueue;
 
-		struct  
+		struct
 		{
 			bool isThreadStarted;
 			bool isInitialized2;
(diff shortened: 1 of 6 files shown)


Exzap:
Appreciate the contribution. Just a note about string handling, `CafeWideString` is meant to replace the manual C-style management of strings but your PR is refactoring the use of it back to C `uint16be[]` arrays. Can you change it to use `CafeWideString` where it makes sense. If needed you can also add helper methods to `CafeWideString` since it is a bit bare bones right now.

primaviera:
I reverted the `uint16be[]` array change and the comment string uses `CafeWideString` now.

Issue #1890, opened 2026-04-25
When ARM64 Windows 11 ARM Native version of CEMU?


Ncard00123:
### Your suggestion

Since Snapdragon laptop adoption is growing at a rapid pace, with the launch of the X2 chips, having a Native ARM64 version of the CEMU emulator is crucial. 

So please consider this at some point in the future.

Exzap:
We have ARM support for pretty much every platform except for Windows but it depends on external contributors (I dont have an ARM Windows device), so feel free to submit a PR for this. But also this is a duplicate feature request

rhemfur:
I took a shot at this and opened a PR adding native Windows ARM64 (AArch64)
support: https://github.com/cemu-project/Cemu/pull/1992

It builds with clang-cl and reuses the existing AArch64 recompiler backend, so it
runs natively, no x64 emulation. I've verified it builds and runs on my Surface
Laptop 7 (ARM); the "Getting started" wizard and main UI come up fine.

The changes are guarded per-architecture, so x86/Linux/macOS builds are unaffected.
Build steps are in BUILD.md (Ninja + clang-cl + -DCMAKE_LINKER_TYPE=MSVC).

Would love for other Windows-on-ARM folks to test it and report back. Feedback and
review very welcome! Enjoy!

PR #1992, opened 2026-07-16
Add native Windows ARM64 (AArch64) support

rhemfur:
Addresses #1890.

After seeing #1890 I decided to give this a shot, since I have a Surface Laptop 7
(ARM). Hope you like it!

This enables building and running Cemu natively on Windows on ARM (e.g. Snapdragon
devices) with clang-cl. The existing AArch64 recompiler backend is used, so no x64
emulation is involved. Verified building and running on my device.

Build system (CMakeLists.txt, dependencies/ih264d, src/Cafe):
- Auto-select the arm64-windows-static vcpkg triplet and detect the AArch64
  architecture even when the emulated host shell misreports it.
- Disable IPO for clang-cl (LTO would force lld-link, which cannot read vcpkg's MSVC
  /GL objects) and link clang_rt.builtins for the compiler-rt calls (e.g. __udivti3)
  that lld-link does not add.
- Set the static CRT on the xbyak_aarch64 target.
- Locate pkgconf under the actual vcpkg host triplet instead of a hardcoded
  x64-windows path.
- Build the bundled ih264d decoder as portable C on ARM64/Windows; its NEON .s files
  use ELF-only :got: relocations that do not assemble for COFF targets.

Source (precompiled.h, ExceptionHandler_win32, Debugger, LatteTextureCache):
- Guard x86-only intrinsics/paths behind ARCH_X86_64. clang-cl defines _MSC_VER but
  not __GNUC__, so several `#if defined(_MSC_VER)` branches wrongly selected x86
  intrinsics on ARM64. Adds AArch64 register dumping to the crash handler.

Manifest: use processorArchitecture="*" for Common-Controls so it resolves on ARM64.

x86 / Linux / macOS builds are unaffected. Build steps are documented in BUILD.md
(Ninja + clang-cl + -DCMAKE_LINKER_TYPE=MSVC).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -204,6 +204,84 @@ jobs:
         name: cemu-installer-windows-x64
         path: ./src/resource/cemu-${{ inputs.next_version_major }}.${{ inputs.next_version_minor }}-windows-x64-installer.exe
 
+  build-windows-arm64:
+    continue-on-error: true
+    runs-on: windows-11-arm
+    steps:
+    - name: "Checkout repo"
+      uses: actions/checkout@v6
+      with:
+        submodules: "recursive"
+
+    - name: Setup release mode parameters
+      run: |
+        echo "BUILD_MODE=release" | Out-File -FilePath $Env:GITHUB_ENV -Encoding utf8 -Append
+        echo "BUILD_FLAGS=" | Out-File -FilePath $Env:GITHUB_ENV -Encoding utf8 -Append
+        echo "Build mode is release"
+
+    - name: Setup build flags for version
+      if: ${{ inputs.next_version_major != '' }}
+      run: |
+        echo "[INFO] Version ${{ inputs.next_version_major }}.${{ inputs.next_version_minor }}"
+        echo "BUILD_FLAGS=${{ env.BUILD_FLAGS }} -DEMULATOR_VERSION_MAJOR=${{ inputs.next_version_major }} -DEMULATOR_VERSION_MINOR=${{ inputs.next_version_minor }}" | Out-File -FilePath $Env:GITHUB_ENV -Encoding utf8 -Append
+
+    - name: "Ensure clang-cl (LLVM) toolset is installed in Visual Studio"
+      run: |
+        $vswhere = "${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer\vswhere.exe"
+        $vsPath = & $vswhere -latest -products * -property installationPath
+        $setup = "${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer\setup.exe"
+        $modifyArgs = "modify --installPath `"$vsPath`" --add Microsoft.VisualStudio.Component.VC.Llvm.Clang --add Microsoft.VisualStudio.Component.VC.Llvm.ClangToolset --quiet --norestart"
+        Start-Process -FilePath $setup -Wait -ArgumentList $modifyArgs
+
+    - name: "Setup MSVC environment (arm64)"
+      uses: ilammy/msvc-dev-cmd@v1
+      with:
+        arch: arm64
+
+    - name: "Setup Ninja"
+      run: choco install ninja -y
+
+    - name: "Setup cmake"
+      uses: jwlawson/actions-setup-cmake@v2
+      with:
+        cmake-version: '3.29.0'
+
+    - name: "Bootstrap vcpkg"
+      run: |
+        ./dependencies/vcpkg/bootstrap-vcpkg.bat
+
+    - name: 'Setup NuGet Credentials for vcpkg'
+      shell: 'bash'
+      run: |
+        `./dependencies/vcpkg/vcpkg.exe fetch nuget | tail -n 1` \
+        sources add \
+        -source "https://nuget.pkg.github.com/${{ github.repository_owner }}/index.json" \
+        -storepasswordincleartext \
+        -name "GitHub" \
+        -username "${{ github.repository_owner }}" \
+        -password "${{ secrets.GITHUB_TOKEN }}"
+        `./dependencies/vcpkg/vcpkg.exe fetch nuget | tail -n 1` \
+        setapikey "${{ secrets.GITHUB_TOKEN }}" \
+        -source "https://nuget.pkg.github.com/${{ github.repository_owner }}/index.json"
+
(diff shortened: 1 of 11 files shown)


npark2860-cyber:
Very interesting to see this PR. I've been working independently on a Windows ARM64 Cemu port on a Snapdragon X Elite / Adreno X1-85 system, and the initial bring-up ended up following a very similar direction: clang-cl, the existing `xbyak_aarch64` backend, and removing Windows == x86-64 assumptions.

After the native build was working, however, most of my time went into runtime correctness on the Qualcomm Windows Vulkan driver. One of the more interesting problems turned out not to be the AArch64 JIT at all, but Cemu's Vulkan occlusion-query readback path.

Cemu normally does roughly:

`vkCmdCopyQueryPoolResults -> persistent mapped result buffer -> CPU consumes mapped value`

On this Adreno configuration I found reproducible cases where that copied/mapped value was wrong even though the query had completed.

For example, while debugging Star Fox Zero I repeatedly captured cases with:

```text
cmdFinished = 1
vkGetQueryPoolResults = VK_SUCCESS
direct result > 0
mapped result = 0
```

So this was not simply the CPU reading an unfinished query.

I then compared the mapped result against a direct `vkGetQueryPoolResults()` call for the exact same query index. In the failing Star Fox scene, hundreds of thousands of queries reproduced the mismatch; in one full run there were 404,428 direct-vs-mapped mismatches, while the direct result kept the previously broken/flickering scene correct.

Bayonetta 2 reproduced the same behavior.

I went fairly far trying to determine whether this was simply a Vulkan synchronization/coherency mistake on our side. I tested:

* an explicit `TRANSFER_WRITE -> HOST_READ` dependency after `vkCmdCopyQueryPoolResults`
* explicit `vkInvalidateMappedMemoryRanges`
* an intermediate DEVICE_LOCAL buffer:
  `QueryPool -> DEVICE_LOCAL -> mapped host buffer`
* explicit transfer dependencies around that intermediate copy

None of these repaired the copied query result.

The selected query-result memory in the Bayonetta 2 capture was actually:

```text
DEVICE_LOCAL | HOST_VISIBLE | HOST_COHERENT | HOST_CACHED
```

so missing non-coherent invalidation was ruled out as well.

Even with the DEVICE_LOCAL intermediate path, Bayonetta 2 still produced more than 23,000 captured cases where the command buffer was finished, the direct query result was non-zero, and the copied/mapped result was zero.

That narrowed the failure boundary down quite strongly: on this Qualcomm Windows Vulkan configuration, `vkGetQueryPoolResults()` returned the correct completed result while the `vkCmdCopyQueryPoolResults()` path systematically failed to deliver the same value for these workloads.

The workaround I ended up validating was therefore deliberately small:

```text
HasCommandBufferFinished(...)
    -> vkGetQueryPoolResults(..., VK_QUERY_RESULT_64_BIT)
       -> VK_SUCCESS: consume direct result
       -> VK_NOT_READY: retain the query/fragment and retry later
```

There is no `VK_QUERY_RESULT_WAIT_BIT`, so this does not introduce a blocking wait.

This produced full runtime passes for both games:

* **Star Fox Zero**: the previously broken/flickering scene remained fixed. In one validation run, 404,653 direct reads were logged and every observed call returned `VK_SUCCESS`.
* **Bayonetta 2**: remained in the fixed state with 65,236 logged direct reads, also with no observed `VK_NOT_READY`, query-index exhaustion, or retention stall.

Bayonetta 2 was especially misleading during diagnosis because the visible failure initially looked like a rendering/resource problem. I traced it backwards through the affected draw, resource identity, producer resources/uniforms, depth identity, index contents, sampled texture state and depth-compare texture history before the query path finally became the common failure boundary.

**Xenoblade Chronicles X** exposed the query problem through another path. I initially kept XCX separate and also tested a crude `0 -> 1` "force visible" occlusion-result experiment. I rejected that approach rather than shipping a game-logic-changing hack. The final release instead extends the same direct `vkGetQueryPoolResults` readback behavior to XCX as well, while retaining the normal non-blocking semantics. The release currently covers the JPN/USA/EUR title IDs for Bayonetta 2, Star Fox Zero and XCX.

There was another unrelated Adreno issue earlier in the port: missing/default vertex inputs were not represented robustly in the vertex-shader key on this path. Synthesizing `DEFAULT_VAL` into the VS input/linkage state stabilized the shader key and fixed rendering in titles including Breath of the Wild and Tekken Tag Tournament 2, and also eliminated the `VK_ERROR_UNKNOWN (-13)` path I had been seeing.

Performance has been encouraging as well. On the same Snapdragon X Elite machine and equivalent Breath of the Wild settings, I measured roughly:

* x64 Cemu under Windows ARM emulation: **30–32 FPS**
* native Windows ARM64 Cemu: **47–52 FPS**

So at least on Snapdragon X Elite, native AArch64 is a very meaningful performance target, not just a portability exercise.

The existing Cemu AArch64 backend is impressively mature. Once the Windows/x86 assumptions were removed, most of the difficult work I encountered was around platform/driver behavior rather than having to redesign the CPU backend.

My current work is separated into Release / Diagnostics / Test branches so that all the tracing and rejected experiments stay out of the release build. I'll compare my release branch against this PR more carefully — there are probably several runtime/Adreno findings that could be useful upstream.
If you see any areas in the current AArch64 backend or Windows ARM64 path that look especially promising for further optimization, I'd really appreciate any pointers or ideas. I've started profiling the JIT and Vulkan paths, but suggestions from people familiar with Cemu's internals would be extremely useful.

In particular, I'm interested in optimization opportunities that may be specific to native AArch64 rather than general Cemu optimizations.

rhemfur:
Closing this PR.

After opening it I read the project's AI policy (in the README, now in CONTRIBUTING.md) and this PR does not meet it. The code here was written by an AI assistant (Claude, as the Co-Authored-By trailers on the commits show); my part was directing the work, building and testing it on my Surface Laptop 7 (Snapdragon X, Adreno X1-85), and running the CI. It was not written by me, and I should have read the policy before submitting. Sorry for the review time this cost you.

I'll keep the work as an independent, clearly unofficial fork under MPL-2.0: https://github.com/rhemfur/Cemu/tree/windows-arm64 (not affiliated with or endorsed by the Cemu project).

Thanks for approving the CI run, and for Cemu. If the build notes are ever useful as a reference for a human-written ARM64 port, they are all in the commit messages and BUILD.md there.

PR #2010, opened 2026-07-25
gx2: add GX2SetTVEnable, GX2SetDRCEnable and GX2SetTVScale

NKR0ne:
These three exports were unimplemented and fell through to the generic `Unsupported lib call` handler.

Cemu presents whichever scan buffer a title copies to regardless of the enable state, so **behaviour is unchanged**. Implementing them removes noise from the log and makes a title's intended output configuration visible under `LogType::GX2`, which was useful while working out how Disney Infinity drives its TV and GamePad output.


capitalistspz:
https://github.com/cemu-project/Cemu/blob/main/CODING_STYLE.md#hle-interfaces

PR #2022, opened 2026-08-06
coreinit: Return correct size from `MEMResizeForMBlockExpHeap` 

capitalistspz:
When block size equalled the requested size, 0 was returned

### DIFF
--- a/src/Cafe/OS/libs/coreinit/coreinit_MEM_ExpHeap.cpp
+++ b/src/Cafe/OS/libs/coreinit/coreinit_MEM_ExpHeap.cpp
@@ -643,6 +643,10 @@ uint32 MEMResizeForMBlockExpHeap(MEMHeapHandle heap, void* memBlock, uint32 size
 				newSize = 0;
 		}
 	}
+	else
+	{
+		newSize = dataSize;
+	}
 
 	heap->ReleaseLock();
 


Crementif:
Just in case we're not patching out an intentional replication of the real console, did you test this same thing on the Wii U?

But other then that, this seems very nice!

capitalistspz:
> Just in case we're not patching out an intentional replication of the real console, did you test this same thing on the Wii U?
> 
> But other then that, this seems very nice!

I tested on the console and on Cemu with this program
```cpp
#include <coreinit/memdefaultheap.h>
#include <coreinit/memexpheap.h>
#include <coreinit/memheap.h>
#include <whb/proc.h>
#include <coreinit/debug.h>
#include <cstddef>

static constinit std::byte ExpHeap[65536] = {};



int main()
{
    WHBProcInit();

    auto expHeap = MEMCreateExpHeapEx(ExpHeap, sizeof(ExpHeap), MEM_HEAP_FLAG_USE_LOCK);
    auto allocation = MEMAllocFromExpHeapEx(expHeap, 128, 128);

    auto const size = MEMGetSizeForMBlockExpHeap(allocation);
    OSReport("Allocation size is %d\n", size);
    auto resize = MEMResizeForMBlockExpHeap(expHeap, allocation, 128);
    OSReport("Resized to %d\n", resize);

    resize = MEMResizeForMBlockExpHeap(expHeap, allocation, 256);
    OSReport("Resized to %d\n", resize);
    while (WHBProcIsRunning()) {}

    MEMDestroyExpHeap(expHeap);

    WHBProcShutdown();

    return 0;
}
```

Exzap:
good catch, thanks

PR #2034, opened 2026-08-21
Vulkan/GLSL: type pixel shader color outputs to match render-target datatype

jmasides:
Addresses the integer render-target pipeline failure reported in #396.

### Problem

Cemu's GLSL pixel shader generator (shared by the Vulkan and OpenGL backends)
unconditionally declares and exports color outputs as `float`/`vec4`,
regardless of the actual datatype of the bound render target. The native
Metal backend already handles this correctly via `GetColorBufferDataType()`.

On macOS with MoltenVK, this mismatch fails pipeline creation outright
whenever a game binds an integer-format render target (e.g. `RGBA16Uint`):

```
output of type float4 is not compatible with a MTLPixelFormatRGBA16Uint color attachment
```

This blocks affected titles - Breath of the Wild included - from rendering
at all on Vulkan/MoltenVK past the point where such a render target is
first bound.

### Fix

- Add a renderer-neutral `Latte::ColorBufferDataType` classifier
  (`LatteReg.h`), derived from the existing GX2 format `FMT_BIT_INT`/
  `FMT_BIT_SIGNED` flags, rather than duplicating Metal's format table.
- `LatteDecompilerEmitGLSLHeader.hpp`: declare `uvec4`/`ivec4`/`vec4` per
  color output based on the bound target's datatype, matching what the
  Metal backend already does. Skip the declaration entirely when no
  target is bound, also matching Metal.
- `LatteDecompilerEmitGLSL.cpp`: wrap the color export in
  `floatBitsToUint`/`floatBitsToInt` for integer targets. The GPR read
  stays float-typed throughout - only the bitcast at the export site
  changes, mirroring Metal's `as_type<>` there. Only the
  `passPixelColorN` assignment is touched; the alpha-test float read a
  few lines above is untouched.
- `LatteShader.cpp`: since the generated GLSL is now render-target-type
  dependent, fold that datatype into the pixel shader's cache identity
  (`LatteSHRC_CalcPSAuxHash`) for both GLSL-based backends (Vulkan and
  OpenGL), so a shader compiled for one render-target datatype can't be
  reused against a different one. Kept outside any Metal-specific gate
  so this applies correctly on non-Apple Vulkan/OpenGL builds too, and
  scoped narrowly to just the datatype (not full `VkFormat`, depth
  state, or Metal's texture-as-render-target hashing).

Vulkan and OpenGL are the only backends touched; the native Metal
backend's own (already-correct) typing logic is unchanged.

### Testing

- Reproduced the failure on an unmodified `main` build first, to confirm
  it wasn't specific to any local changes.
- Built with the fix and ran Breath of the Wild for an extended real
  session (13+ minutes) past the point that previously failed
  immediately - no pipeline-creation errors, stable throughout.
- Audited the Vulkan pixel-shader cache/variant key specifically for
  this change (a first version of the fix had a gap here: the cache key
  needs the render-target datatype folded in, or a shader compiled
  against one datatype can get incorrectly reused against another).

### Scope note

This is a rendering-correctness fix, not a performance change. It
doesn't touch the deeper Vulkan/MoltenVK performance characteristics on
this title - that's tracked separately.

### DIFF
--- a/src/Cafe/HW/Latte/Core/LatteShader.cpp
+++ b/src/Cafe/HW/Latte/Core/LatteShader.cpp
@@ -690,6 +690,22 @@ uint64 LatteSHRC_CalcPSAuxHash(LatteDecompilerShader* pixelShader, uint32* conte
 	}
 #endif
 
+	// GLSL: generated pixel-color output types depend on the bound color buffer's datatype
+	// (see LatteDecompilerEmitGLSLHeader.hpp / LatteDecompilerEmitGLSL.cpp). That GLSL is shared
+	// by both GLSL-based backends (Vulkan and OpenGL), so it must be part of shader identity for
+	// both. Renderer-neutral (no ENABLE_METAL dependency) so this is correct on Windows/Linux
+	// builds as well, not just macOS.
+	if (g_renderer->GetType() == RendererAPI::Vulkan || g_renderer->GetType() == RendererAPI::OpenGL)
+	{
+		for (uint8 i = 0; i < LATTE_NUM_COLOR_TARGET; i++)
+		{
+			auto format = LatteMRT::GetColorBufferFormat(i, LatteGPUState.contextNew);
+			uint8 dataType = (uint8)Latte::GetColorBufferDataType(format);
+			auxHash = std::rotl<uint64>(auxHash, 7);
+			auxHash += (uint64)dataType;
+		}
+	}
+
 	return auxHash;
 }
 
--- a/src/Cafe/HW/Latte/ISA/LatteReg.h
+++ b/src/Cafe/HW/Latte/ISA/LatteReg.h
@@ -309,6 +309,26 @@ namespace Latte
 	};
 	DEFINE_ENUM_FLAG_OPERATORS(E_GX2SURFFMT);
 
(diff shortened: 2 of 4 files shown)


Crementif:
This would invalidate the entirety of the community's graphic pack shaders, no? I'm obviously biased as its maintainer, but the Metal port dropping support for the graphic pack shaders wasn't really a positive. So replacing that for other platforms doesn't sound like a huge win. It should at least be made specific for MoltenVK (which tbf, likely will be deprecated anyway, so I'm not sure what the point of this PR would even be then).

Also, this reads like AI generated code, which we don't tolerate in our AI policy. Its fine for research, but we expect the PR person to understand every line they've written. Otherwise it just burdens us with having to understand your changes and verify that it doesn't exhibit unwanted behavior. Just letting you know in case this gets closed for that reason.

jmasides:
Thanks for taking the time to review this. After looking deeper into the shader identity / graphic pack interaction, I agree this approach has a wider compatibility impact than I initially accounted for. The intent was to fix a MoltenVK-specific issue, but the implementation touches shader identity and you're right that it could affect existing graphic packs. Rather than ask you to review a patch that's going to be deprecated anyway, I'll close this PR. The investigation was still useful for me (and now it works on my mac!). Thanks!

PR #2035, opened 2026-08-21
Metal: independent texture/sampler binding mapping

jmasides:
### Problem

The native Metal backend used one shared, incrementing counter to assign
binding indices for a shader's active texture units, then reused that
identical index for both Metal's texture argument table (31 slots) and
its sampler argument table (16 slots) - two genuinely separate binding
tables. Any shader with more than 16 active texture units overflowed the
smaller table with no bounds check catching it:

```cpp
// LatteDecompilerAnalyzer.cpp, _initTextureBindingPointsMTL (before)
decompilerContext->output->resourceMappingMTL.textureUnitToBindingPoint[i]
    = decompilerContext->currentTextureBindingPointMTL;
...
decompilerContext->currentTextureBindingPointMTL++;
```

```cpp
// LatteDecompilerEmitMSLHeader.hpp, _emitTextureDefinitions (before)
uint32 binding = shaderContext->output->resourceMappingMTL.textureUnitToBindingPoint[i];
src->addFmt(" tex{} [[texture({})]]", i, binding);
src->addFmt(", sampler samplr{} [[sampler({})]]", i, binding);  // same index for both
```

That unguarded index (up to 30 for a fully-populated shader) then reached
`SetSamplerState`, which had no bounds check of its own and wrote directly
into a fixed 16-slot array (`m_samplers[shaderType][index]` in
`MetalRenderer.h`) - an out-of-bounds write, and an invalid index handed
straight to Metal's `setFragmentSamplerState`/`setVertexSamplerState`.
Breath of the Wild is the specific case that surfaces this: several of its
pixel shaders use 17-18 simultaneous texture units.

### Why a simple fix is viable here (not argument buffers)

The natural-seeming alternative - Metal argument buffers, which remove the
16-slot ceiling entirely - is a substantially larger change touching
pipeline creation and encoder setup throughout the Metal renderer, and
isn't used anywhere in the backend today. Before reaching for that, this
PR checks a narrower question: do the shaders that exceed 16 *texture*
units also exceed 16 *unique sampler configurations*? Texture and sampler
are modeled as genuinely distinct resources at the Latte/GX2 ISA level - a
texture-fetch instruction carries separate `bufferId` and `samplerId`
fields - so a shader can easily use many texture units while reusing far
fewer distinct samplers.

Measured directly, across two full Breath of the Wild sessions (1440p and
4K, real gameplay - open world, village, water, weather, Guardian combat,
the Ganon fight, ~5,000 shaders analyzed combined): the answer is yes.
Every shader that exceeded 16 active textures (max observed: 18) still
resolved to at most 16 unique guest sampler ids - never more. That's the
precondition this fix depends on, and it held with zero exceptions.

### Fix

- `LatteDecompiler.h`: add `textureUnitToSamplerBindingPoint[]` to
  `LatteDecompilerShaderResourceMapping` - an independent Metal sampler
  mapping, parallel to but separate from the existing texture mapping,
  keyed by texture unit and densely assigned by *unique guest sampler id*.
  Texture units that share a guest sampler id share a Metal sampler slot.
- `LatteDecompilerAnalyzer.cpp` (`_initTextureBindingPointsMTL`): assign
  the new mapping during analysis via a dense per-shader guest-sampler-id
  dedup. A shader needing more than `MAX_MTL_SAMPLERS` unique guest
  samplers (not observed in any measured BOTW shader) sets
  `shader->hasError` and logs clearly, rather than aliasing, clamping, or
  writing out of bounds.
- `LatteDecompilerEmitMSLHeader.hpp` (`_emitTextureDefinitions`): emit
  `[[texture()]]` and `[[sampler()]]` from the two independent bindings
  instead of reusing one value for both - the actual bug. MSL forbids two
  live parameters at the same `[[sampler(N)]]` index (MSL spec 5.2.1), so
  a texture unit that shares a Metal sampler slot with an earlier unit
  skips declaring its own sampler parameter and instead gets a
  `#define samplrN samplrM` aliasing its body references to the owning
  unit's parameter. Removed the disabled `% 31` / `% 16` dead code this
  replaces, present-but-inactive since the Metal backend's introduction.
- `MetalRenderer.cpp` (`BindStageResources`): compute the independent
  sampler binding per texture unit and use it for both `SetSamplerState`
  call sites, instead of reusing the texture binding. `SetSamplerState`
  itself gained a hard bounds guard (`index >= MAX_MTL_SAMPLERS`) as
  defense in depth - refuses rather than writing out of bounds. The
  primary guarantee is upstream (`shader->hasError`); this is a second,
  independent line of defense.
- `MetalSamplerCache` is unchanged, still used only for its existing
  `MTL::SamplerState` object-level dedup - this fix doesn't touch how
  sampler *objects* are created or cached, only which binding slot a
  given texture unit's sampler is placed in.

Vulkan and OpenGL resource mapping are untouched - this is Metal-exclusive
(`_initTextureBindingPointsMTL`, the MSL emitter, and `MetalRenderer.cpp`
are all reached only from the Metal backend).

### Testing

- **Structural, before any gameplay**: rebuilt with Cemu's shader-dumping
  enabled, swept every shader compiled during a real title load (~3,000
  shaders). Zero contain `[[sampler(16)]]` or higher, verified against
  the literal generated MSL text. The known 18-texture shader's dumped
  source shows exactly the intended pattern: units 0-15 get real
  `[[sampler(0..15)]]` declarations, units 16/17 get `#define samplr16
  samplr0` / `#define samplr17 samplr1`. All ~3,000 shaders compiled
  successfully through Metal's actual shader compiler (dump happens after
  renderer object creation, not just text generation) - no crashes, no
  compile rejections.
- **Real gameplay, twice**: two full sessions on the built fix (1440p
  ~48 FPS, 4K ~30 FPS - matching the pre-fix baseline, no regression from
  adding the independent mapping), covering open world, a village, water,
  weather, Guardian combat, and the Ganon fight. Zero shaders exceeded 16
  unique samplers in either session; four distinct shaders exceeded 16
  textures across the two sessions, all correctly capped.
- **Camera Rune and a shrine interior**: checked separately, no visual
  issues.
- **Targeted follow-up on a reported visual issue**: after a
  (separately-caused) report of incorrect water rendering, swept every
  shader compiled during that specific session and confirmed none of them
  use the new aliasing mechanism at all - the water shader's behavior is
  unchanged by this fix either way. That white/transparent water turned
  out to be a pre-existing, previously-documented Cemu/BOTW rendering
  quirk that also occurs on Vulkan, unrelated to this change.

### Note

This PR is independent of the Vulkan/GLSL typed-output PR (#2034)
- same base commit, different renderer backend, no code overlap - not
meant to be combined with it.

### DIFF
--- a/src/Cafe/HW/Latte/LegacyShaderDecompiler/LatteDecompiler.h
+++ b/src/Cafe/HW/Latte/LegacyShaderDecompiler/LatteDecompiler.h
@@ -53,12 +53,21 @@ struct LatteDecompilerShaderResourceMapping
 		std::fill(relBindingPointToRelTextureUnit, relBindingPointToRelTextureUnit + LATTE_NUM_MAX_TEX_UNITS, UNUSED_BINDING);
 		std::fill(uniformBuffersBindingPoint, uniformBuffersBindingPoint + LATTE_NUM_MAX_UNIFORM_BUFFERS, UNUSED_BINDING);
 		std::fill(attributeMapping, attributeMapping + LATTE_NUM_MAX_ATTRIBUTE_LOCATIONS, UNUSED_BINDING);
+		std::fill(textureUnitToSamplerBindingPoint, textureUnitToSamplerBindingPoint + LATTE_NUM_MAX_TEX_UNITS, UNUSED_BINDING);
 	}
 	// texture
 	sint8 textureUnitToBindingPoint[LATTE_NUM_MAX_TEX_UNITS]; // mostly for OpenGL backwards compatibility where texture units are not remapped and binding points are sparse
 	sint8 relBindingPointToRelTextureUnit[LATTE_NUM_MAX_TEX_UNITS]; // (only used on VK and Metal) index is relative binding point (absoluteBindingPoint - textureUnitBaseBindingPoint)
 	sint8 textureUnitBaseBindingPoint{UNUSED_BINDING};
 	sint8 textureUnitCount{0}; // number of entries set in textureUnitToBindingPoint
+	// Metal exclusive: independent sampler mapping, keyed by texture unit. Deliberately separate
+	// from textureUnitToBindingPoint - Metal's texture (31-slot) and sampler (16-slot) argument
+	// tables are genuinely distinct, and reusing one index for both overflows the smaller table
+	// once a shader has >16 active texture units. Densely assigned by unique guest sampler ID
+	// (see _initTextureBindingPointsMTL), so texture units sharing a guest sampler share a Metal
+	// sampler slot too. UNUSED_BINDING means this shader exceeded MAX_MTL_SAMPLERS unique guest
+	// sampler ids - shader->hasError is set in that case, this value should never reach emission.
+	sint8 textureUnitToSamplerBindingPoint[LATTE_NUM_MAX_TEX_UNITS];
 	// uniform buffer
 	sint8 uniformVarsBufferBindingPoint{UNUSED_BINDING}; // special block for uniform registers/remapped array/custom variables
 	sint8 uniformBuffersBindingPoint[LATTE_NUM_MAX_UNIFORM_BUFFERS];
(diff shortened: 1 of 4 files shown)


jmasides:
I've moved the useful part of this work (the reproduction, measurements and root-cause analysis) to #2038.

I'm going to close this PR rather than ask the maintainers to review an implementation that doesn't fit the project's contribution policy. Leaving the PR and branch available as an experimental prototype/reference if it's useful to whoever eventually works on the issue.

Thanks!

PR #2047, opened 2026-09-02
GX2: Sync with GPU when polling a pending CPU occlusion query (fixes ZombiU lights visible through walls)

### LINKED ISSUE #635: ZombiU issues with lights through the walls
Zombi U lights can cross the walls, this problem happens since the first cemu version able to run this game and also happens in both backends (Vulkan and opengl)

Tested with latest experimental version: 2.0-26

Cemu 2023-01-28 20-45-14
Cemu 2023-01-28 20-45-16
Cemu 2023-01-28 20-44-04
Cemu 2023-01-28 20-44-48
Cemu 2023-01-28 20-45-01

Specs: Windows 11
GTX 970
16 GB DDR3
i7 4790k

Forgottenshadow89:
Fixes #635 (ZombiU: light glows / lens flares visible through walls). I opened that issue back in 2023 and finally got around to tracking it down.

**What the game does**

ZombiU (Ubisoft's LyN engine) decides whether each light's glow is occluded by drawing a small invisible quad at the light position inside a CPU-type occlusion query (`GX2QueryBegin(GX2_QUERY_TYPE_OCCLUSION_CPU)`), and then reads the result with `GX2QueryGetOcclusionResult` only a few milliseconds later, **within the same frame**. Zero samples → glow hidden. Samples > 0, *or result not ready yet* → glow drawn at full intensity.

This works on the console because the game does a full GPU wait at the end of every frame (`GX2Flush` → `GX2GetLastSubmittedTimeStamp` → `GX2WaitTimeStamp` right before `GX2SwapScanBuffers`), so the GPU is never far behind and the query result has already been written by the time the game polls it.

**What goes wrong in Cemu**

The GPU thread can be well behind the PPC thread at that point, so the result isn't written yet and `GX2QueryGetOcclusionResult` returns `GX2_FALSE`. With logging added around the query functions, about 94% of ZombiU's polls hit the "not ready" path, which the game treats as "fully visible" — hence glows shining through walls, in both OpenGL and Vulkan, since forever. The query results themselves are correct when they do arrive (0 samples behind walls, > 0 when visible); it's purely a timing issue.

**The fix**

When `GX2QueryGetOcclusionResult` finds a CPU query that is still pending, emit the `IT_HLE_SYNC_ASYNC_OPERATIONS` packet, flush and wait for the GPU to retire it (the same thing `GX2DrawDone` does with full sync), then re-check. Only polls of queries that are actually pending pay for this, so games whose results are already available are unaffected, and the sync is skipped while a display list is being recorded. GPU-type queries (and the XCX workaround) are untouched.

**Testing**

ZombiU (EUR), Vulkan, on an i5-13600KF / NVIDIA: glows are now correctly hidden behind walls both standing still and while moving, fade in normally when the light comes into view, and the game stays at a steady 30 FPS. In this game the sync triggers about 1.6 times per frame (the game polls in two batches per frame). I also tried the simpler alternative of not resetting the query data at `GX2QueryBegin` so that the previous frame's result stays readable, but ZombiU also feeds these queries into another system that expects the current frame's result, and that produced heavy motion blur artifacts — so waiting for the real result is the right thing to do here.

### DIFF
--- a/src/Cafe/OS/libs/gx2/GX2_Query.cpp
+++ b/src/Cafe/OS/libs/gx2/GX2_Query.cpp
@@ -4,6 +4,7 @@
 #include "Cafe/HW/Latte/Core/LattePM4.h"
 #include "Cafe/CafeSystem.h"
 #include "GX2_Query.h"
+#include "GX2_Command.h"
 
 #define LATTE_GC_NUM_RB							2
 #define _QUERY_REG_COUNT						8 // each reg/result is 64bits, little endian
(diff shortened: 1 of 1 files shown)


Forgottenshadow89:
PS: I did the code with AI help. In case the devs doesn't want to merge this commit because they don't like AI things, I totally understand. But this fixes a long-standing issue with just 14 lines of code. So, in case any other person wants to develop a better fix, at least he/she can take this as an example or reference.

Issue #1823, opened 2026-03-03
ZombiU issue - Blackscreen when update is present


LuismaSP89:
### Current Behavior

This was a very old issue, but I didn't reported thil now, sorry for being so lazy. 

The thing is that the game ZombiU currently shows a blackscreen with no sound after the ubisoft's logo. This only happens when a update is installed. Removing the update solves the issue, but since the update fixes many issues ingame, I think it's important to have this one fixed.

### Expected Behavior

Game working using the update.

### Steps to Reproduce

Install the update for the zombiU (PAL EU) and try to run the game. After the ubisoft logo it only shows blackscreen with fixed FPS. Uninstalling the update works and you can get ingame.

### System Info (Optional)

OS: W11 pro x64
GPU:  RTX 4070
i5 13600KF
32 GB DDR5

### Logs (Optional)

[log.txt](https://github.com/user-attachments/files/25724516/log.txt)

cabrerahector:
Known issue, see https://wiki.cemu.info/wiki/ZombiU

Forgottenshadow89:
> Known issue, see https://wiki.cemu.info/wiki/ZombiU

Uh, you're right, but it's strange nobody opened an issue on this tracker. I'll keep it opened unless the devs are not fine with it.

cabrerahector:
The reason why nobody had opened a ticket about that issue here is because it was already documented in the wiki @LuismaSP89.

Generally speaking, there's no need to track the same problem in two different places. If the problem is already documented somewhere then the devs most likely know already about it.

Forgottenshadow89:
Im my humble opinion this should be documented here, since this is the issue tracker for the emulator itself. Of course the wiki helps in case you have an specific problem with a game and someone knows what options you need to change in the emulator instead using the default ones to be able to run that game, but I don't think the devs are going to read each wiki for all the games on the console to search for issues, having it here helps them to have focus on the emulator's current issues. 

My approach here is something like, the issue can be documented previously, for example in discord by a user, but it can be forgotten with the time. Having the issue in the main issue tracker of the emulator makes more sense in order to get it solved in the future. At least that's my opinion, of course I'll accept any decission the devs can take about this.

Exzap:
[...]

PR #2048, opened 2026-09-02
FSC: Sort merged directory listings by name (fixes ZombiU black screen with update v32)

Forgottenshadow89:
Fixes the long-standing "ZombiU doesn't boot with update v32 installed" problem documented on the wiki (and reported in #1823): the game stays on a black screen after the Ubisoft logo, at 30 FPS, forever.

**What the game does**

At boot ZombiU enumerates `/vol/content/lynfsatest/BinCafe` with `FSReadDir` and registers its `.lin.bfz` archives in the order they are returned. Its boot sequence then loads the intro video by resource id. That resource exists in two archives (`Video_Common_1` and a duplicate in `Gen_Common`), and which copy gets used depends on the archive order.

**What goes wrong in Cemu**

`FSCVirtualFileDirectoryIterator::PopulateIterationList` builds the listing of a directory that exists in several mounted layers by iterating the layers in priority order and appending each layer's entries. With an update installed, every file from the update is therefore listed before any file from the base title. The v32 update adds `BinCafe/Patch.lin.bfz`, which becomes the first archive the game registers and shifts all the others. The boot sequence resolves the intro video to the wrong archive (and starts streaming it from the wrong block), waits out its timeouts and the game never leaves the boot screen. On hardware the directory is returned in FST order, i.e. sorted by name, so `Patch.lin.bfz` lands between `Mip0_Common_1` and `SndStream_Common_0` and everything works.

Two experiments showed it is purely the order: removing `Patch.lin.bfz` from the update makes v32 boot, and so does moving the very same file into the base title's `BinCafe` folder, where the host filesystem lists it alphabetically.

**The fix**

Sort the merged listing by name (case-insensitively, like the FSA name comparison) whenever more than one folder contributes to a directory. Listings that come from a single layer are not touched, so games without an update or DLC see exactly what they saw before.

**Testing**

Confirmed with a build of this branch: ZombiU (EUR) with update v32 installed in its normal location now plays the intro video and reaches the main menu, and the game runs normally from there. Base v0 behaves as before.

### DIFF
--- a/src/Cafe/Filesystem/fsc.cpp
+++ b/src/Cafe/Filesystem/fsc.cpp
@@ -366,6 +366,31 @@ private:
 				}
 			}
 		}
+		// When several mounted folders contribute to the same directory (e.g. base title + update),
+		// concatenating the per-layer listings puts the update's files first. On hardware a directory
+		// is listed in FST order, which is sorted by name, and some games rely on that order.
+		// ZombiU indexes its .bfz archives in FSReadDir order and never gets past the boot screen when
+		// the update's Patch.lin.bfz is enumerated before the base archives.
+		if (m_folders.size() > 1)
+		{
+			std::sort(dirIterator->dirEntries.begin(), dirIterator->dirEntries.end(), [](const FSCDirEntry& a, const FSCDirEntry& b) {
+				std::string_view sa(a.path, strnlen(a.path, FSC_MAX_DIR_NAME_LENGTH));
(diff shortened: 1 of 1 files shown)


Forgottenshadow89:
PS: Same as the previous PR, I did this using AI help. And again... In case the devs doesn't want to merge this commit because they don't like AI things, I totally understand. But this fixes a long-standing issue with just 16 lines of code (Not including the comments). So, in case any other person wants to develop a better fix, at least he/she can take this as an example or reference.

PR #2050, opened 2026-09-03
sndcore:  Add Some `Mix` functions and update `AXSetVoiceDeviceMix`

capitalistspz:
Tested on both the Wii U and Cemu with https://github.com/capitalistspz/wiiu_audio_examples/blob/master/2-play-stereo/main.c

### DIFF
--- a/src/Cafe/OS/libs/snd_core/ax_multivoice.cpp
+++ b/src/Cafe/OS/libs/snd_core/ax_multivoice.cpp
@@ -164,4 +164,18 @@ namespace snd_core
 		return result;
 	}
 
+	sint32 AXSetMultiVoiceDeviceMix(AXVPBMULTI* mv, int32_t deviceType, int32_t deviceIndex, uint32_t busIndex, uint16_t vol, int16_t delta)
+	{
+		for (auto voiceChannel = 0; voiceChannel < mv->channelCount; ++voiceChannel)
+		{
+			std::array<std::array<AXCHMIX2, AX_BUS_COUNT>, AX_TV_CHANNEL_COUNT> mix {};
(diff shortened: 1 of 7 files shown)


src/Cafe/OS/libs/snd_core/ax_multivoice.cpp
```diff
@@ -164,4 +164,18 @@ namespace snd_core
 		return result;
 	}
 
+	sint32 AXSetMultiVoiceDeviceMix(AXVPBMULTI* mv, int32_t deviceType, int32_t deviceIndex, uint32_t busIndex, uint16_t vol, int16_t delta)
```
Exzap: Basic types without `_t`

src/Cafe/OS/libs/snd_core/ax_multivoice.cpp
```diff
@@ -164,4 +164,18 @@ namespace snd_core
 		return result;
 	}
 
+	sint32 AXSetMultiVoiceDeviceMix(AXVPBMULTI* mv, sint32 deviceType, sint32 deviceIndex, uint32 busIndex, uint16 vol, sint16 delta)
+	{
+		for (auto voiceChannel = 0; voiceChannel < mv->channelCount; ++voiceChannel)
+		{
+			std::array<std::array<AXCHMIX2, AX_BUS_COUNT>, AX_TV_CHANNEL_COUNT> mix{};
+			mix[voiceChannel][busIndex].vol = vol;
```
Exzap, line 182: voiceChannel (mv->channelCount) and busIndex should be bounds checked at the beginning of this function to prevent out of bounds access here. Maybe print a log line with LogType::APIErrors so that homebrew devs that use this incorrectly can see it in the log

Issue #1147, opened 2024-04-01
Skylanders Swap Force: Cutscenes/PNG/Icon assets corrupted?


KaoKraft:
### Current Behavior

There are a lot of visual bugs that I've noticed when playing Swap Force, mainly to do with some of the icons and buttons being either wrong or missing.

As well as the video cutscenes suffering from unsynced RGB shift. 
https://github.com/cemu-project/Cemu/assets/76142804/ce32b1fa-cbc4-4997-9b48-ef32f363905a

Icon bugs I've noticed: The game profile selection buttons being missing/invisible
![image](https://github.com/cemu-project/Cemu/assets/76142804/c24f74de-ffd5-4fd5-8f23-fd1d862a6f2b)

The collection tab icons being wrong and only showing the first selected list of characters across all tabs, overwriting the actual images of the correct characters and their bios. Other similarly formatted pictures and icons, such as upgrades, suffer from the same issue through the game as well. It only resets after loading either a different Skylander, or a different level, while still having the issue of whatever icons are shown first overwrite everything else.

https://github.com/cemu-project/Cemu/assets/76142804/5526cde3-80da-4ec8-aab5-42cf4b47aec0

### Expected Behavior

I expected the videos to be played properly without the rgb shifting. All the icons and pictures should be matched to their proper character and not overwrite any other pictures that come up.

### Steps to Reproduce

I loaded up the game and started a new save file, skipped the cutscenes after a moment, placed down my Skylander, and then went into the pause menu, entered the "COLLECTION" menu.

### System Info (Optional)

OS: Win 11 Home
GPU: RTX 3060

### Emulation Settings (Optional)

N/A

ZorkIsEmo:
this is also happening on Imaginators for most icons for collections, and upgrade path icons

mxcrow:
> Is there a fix for it?

From what I’ve noticed there are a few things you can do to have it running smoothly again. 

- Set your graphics API to Vulkan
- Go into your graphics pack for Swapforce and change your resolution to either 1280x720 or 2560x1440 (either have worked for me)
- Set FPS limit to 60

Drowkex220:
I have encountered the same issue and I haven't found any solutions

OddLittlePumpkin:
[...]

PR #2051, opened 2026-09-05
Add 16-byte compressed texture handling to _quickStochasticHash

Cereal-Lain:
BC3 and BC2 use 8 bytes for alpha followed by 8 bytes for rgb, repeating. If memstep happens to be even for a texture, the 37 bytes checked in _quickStochasticHash always end up reading alpha values, making it miss changes to opaque textures. This changes it to read 16 bytes instead of 8 for the 37 spots, so rgb changes are detected properly.

This fixes issues in skylanders swap force for the skylanders collection images. Should resolve most if not all other streamed texture issues in skylanders games, and have positive effects on other games as well. Since we still check the same 37 spots just 1 byte more each, it shouldn't cause any regressions.

Resolves issue #1147, except the save files texture in swap force, seems to be an unrelated issue. Needs testing to see how many of the other streamed in texture issues mentioned [here](https://github.com/cemu-project/Cemu/issues/1147#issuecomment-2818914293) are also resolved by this.

### DIFF
--- a/src/Cafe/HW/Latte/Core/LatteTextureCache.cpp
+++ b/src/Cafe/HW/Latte/Core/LatteTextureCache.cpp
@@ -22,7 +22,7 @@ void LatteTC_UnregisterTexture(LatteTexture* tex)
 }
 
 // sample few uint64s uniformly over memory range
-uint32 _quickStochasticHash(void* texData, uint32 memRange)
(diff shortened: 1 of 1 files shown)


happyraphael66-design:
Im new to all that but can you tell me how i put your fix into my game

PR #2058, opened 2026-09-06
Vulkan: Reuse the running instance in GetDevices to fix crash when opening settings in-game

lmncaj:
Opening the general settings while a game is running crashed every time for me on Linux/Wayland with the NVIDIA driver (595). `GeneralSettings2` fills the graphics device list through `VulkanRenderer::GetDevices()`, which creates a temporary `VkInstance` plus a temporary surface on the main window and destroys both again. With NVIDIA's Wayland WSI that teardown also takes down state shared with the live swapchain: the next `vkAcquireNextImageKHR` on the Latte thread ends up calling a null function pointer inside the driver (gdb shows `?? () at 0x0` below `libnvidia-glcore` in `AcquireNextSwapchainImage`). Cemu's own crash handler only prints `[(nil)]` for this, which is probably why it's hard to spot in user logs.

While the Vulkan renderer is active, enumerate the devices through its existing instance and main swapchain surface instead of creating a second instance. The temporary instance path is kept for the case where no renderer is running yet.

### DIFF
--- a/src/Cafe/HW/Latte/Renderer/Vulkan/VulkanRenderer.cpp
+++ b/src/Cafe/HW/Latte/Renderer/Vulkan/VulkanRenderer.cpp
@@ -108,6 +108,32 @@ std::vector<VulkanRenderer::DeviceInfo> VulkanRenderer::GetDevices()
 
 	std::vector<DeviceInfo> result;
 
+	// use the instance of the running renderer if there is one. Creating and destroying a second instance while a swapchain
+	// is active breaks the NVIDIA Wayland driver (crash in vkAcquireNextImageKHR when opening the settings in-game)
+	if (g_renderer && g_renderer->GetType() == RendererAPI::Vulkan)
+	{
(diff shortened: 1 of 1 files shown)


goeiecool9999:
There's no point in closing PRs where you were open about your use of AI after discovering our LLM policy. Reopening the same PR while hiding signs of AI usage only _reduces_ trust.
I'd also like to see the LLM's claim about global WSI state being corrupted by destroying the temporary instance and surface validated with some outside references. If it's in the shared Claude session I cannot see it because I do not have an account. It seems like something an LLM would likely hallucinate about.
If it is validated, that's technically not our bug. AFAIK nothing in the Vulkan spec disallows our use of a temporary instance and surface. If it was a violation I think the validation layers probably would've caught it.
I'm not opposed to implementing a workaround, but I'm also not pleased with the code duplication the LLM came up with. Ideally this bug would be squashed upstream. I'm curious if we can figure out which compositors are at fault. For example does Weston also exhibit the bug? What about KDE, GNOME, or wlroots based compositors?

Issue #2042, opened 2026-08-29
Add Skylanders Portal IPC server


Cereal-Lain:
### Your suggestion

This would allow for external programs to control which skylanders are loaded, instead of having to go through the ui. There is also currently a pull request open in rpcs3 for exactly this feature, it would be nice to align this implementation to theirs so any programs that might get written to take advantage of this could be cross compatible with little effort. [RPCS3](https://github.com/RPCS3/rpcs3/pull/18960)

Exzap:
What kind of external programs? I don't fully understand how an external UI is better than, for example, just improving the UI/usability in Cemu?

Cereal-Lain:
Customizability would be one, if I dont like the default cemu loader I could use a different one without having to build cemu from source and I could have the same loader on rpcs3 (if cross compatibility ends up working out). 

The main reason for me would be managers and loaders that end up out of scope for an emulator to ship with. One example would be virtport, which hosts a local api to control the loaded skylanders from an android companion app. I personally thought of a similar companion app that's difficult to implement to work with unmodified cemu releases. Virtport had to ship with a modified version of cemu [VirtPort](https://github.com/TonyHacker03/SkylandersVirtualPortal)

RagingLeviathan:
I'd be for this too, I'm after spending too long exporting dumps from my own figures to use with Cemu, only to realise that even though a dump came up as unknown in Cemu, it worked fine in the game. (once I disconnected a physical portal to stop any conflict, and restarted Cemu.)

Is there a list or something that shows what skylanders have been mapped in Cemu? I tried a few different dumps of figures from the 5th game, Superchargers, and noticed they came up as unknown. But if I tried with a figure dump from the first game, Spyro's Adventure, Cemu identified the dump correctly.

PR #2059, opened 2026-09-06
Add Skylanders Portal IPC server

Cereal-Lain:
This adds a skylanders portal IPC server largely inspired/modified from this pull request for [the same feature in rpcs3](https://github.com/RPCS3/rpcs3/pull/18960)

1. adds requestedSlot and GetFigureInfo to Skylander.cpp
2. adds the SkylandersPortalIPC.cpp file, where commands are processed essentially the same as in rpcs3. The server is instead implemented with boost/asio (hopefully to the same behavior)
3. changes the emulated portal UI to add a checkbox to enable the IPC server and changes the UI onto a 500ms timer to update the entries shown

Solves feature request #2042 by me. Shoutout to Der-Floh for their rpcs3 pull request being the blueprint for this

### DIFF
--- a/src/Cafe/CMakeLists.txt
+++ b/src/Cafe/CMakeLists.txt
@@ -430,6 +430,7 @@ add_library(CemuCafe
   OS/libs/nsyshid/Infinity.cpp
   OS/libs/nsyshid/Infinity.h
   OS/libs/nsyshid/Skylander.cpp
+  OS/libs/nsyshid/SkylanderPortalIPC.cpp
   OS/libs/nsyshid/Skylander.h
   OS/libs/nsyshid/SkylanderXbox360.cpp
   OS/libs/nsyshid/SkylanderXbox360.h
--- a/src/Cafe/OS/libs/nsyshid/Skylander.cpp
+++ b/src/Cafe/OS/libs/nsyshid/Skylander.cpp
@@ -815,7 +815,7 @@ namespace nsyshid
 		}
 	}
 
-	uint8 SkylanderUSB::LoadSkylander(uint8* buf, std::unique_ptr<FileStream> file)
+	uint8 SkylanderUSB::LoadSkylander(uint8* buf, std::unique_ptr<FileStream> file, int requestedSlot)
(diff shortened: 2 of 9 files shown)


Exzap:
I don’t think we should add a Skylanders-specific IPC server to Cemu. There are many problems with it:
- Are we going to also add a Lego Dimensions Portal, Disney Infinity, Kamen Rider IPC services? The implementation burden is quite high
- The current IPC protocol is expecting the emulator to already provide a very specific high level Skylanders API within the emulator (e.g. understanding of .sky files). If the goal is to make this a shared protocol across emulators then this is unfriendly towards emulators which have no such portal emulation yet
- Reinventing the wheel with a text based protocol when Skylander portals already have a protocol

If we expose this functionality externally, it should be through a generic “remote USB device” interface that can be reused for other USB peripherals as well. Game or device-specific IPC belongs outside the emulator.

So I’d be in favor of a generic solution here that could be reused for other portal types, but not the Skylanders-specific API in its current form.

Cereal-Lain:
Youre correct that this is pretty specific to skylanders right now, but the command parser itself is pretty universal and could be extended later. This pr is mainly just a clone of the rpcs3 skylanders ipc server, I think supporting the same protocol there would be beneficial. I dont agree that it should be a generic usb protocol, the portal emulation for skylanders, disney infinity and lego dimensions should be the emulators concern so specific fixes can be applied for that specific emulator. If we want to add portal support for the other games later, we can just extend the command parser to add a specifier to say which portal the external app want to control, if that specifier is omitted we can assume that it is skylanders to keep any developed external tools working as intended. Reinventing the wheel with a text based protocol makes it kinda generic as to support other game portals later.

edit: Im also in talks with the person who did the pr for the ipc server in rpcs3. If we want to change stuff about the command parser, like adding a specifier for which game portal, that could be coordinated ahead of time as well. I think the protocol having a function to return the id and variant of the loaded skylanders isnt unreasonable though, based on the specifier this function could return different data for the other game portals down the line.

PR #2063, opened 2026-09-16
made ink submerge pipeline thingy synchronous, fixes ink submerge bug

Synthle:
[SPLATOON VULKAN]

was bugging me alot when playing on pretendo <3


Exzap:
What is the exact problem this fixes? Splatoon uses occlusion queries to check whether you can submerge in ink, but this alters the rules of which pipelines can be compiled async? I don't fully understand the connection. Also this change may break other games and would need much broader testing

